PR #264, opened 2026-06-14
Add manual sell button for open positions on the dashboard

### LINKED ISSUE #269: Dashboard: manual "Sell now" button for open positions
## Summary

Allow the operator to manually close any open position from the dashboard on demand, instead of waiting for the bot's automated take-profit / METAR stop-loss exits.

## Motivation

Operators sometimes want to exit a position before the model-driven exits fire (e.g. discretionary risk-off, ahead of a data outage). Today the only way to close a position early is outside the system.

## Acceptance criteria

- [ ] A "Sell now" control on each open-position card on the dashboard.
- [ ] Two-step confirmation before firing, since the action spends real money irreversibly.
- [ ] Backend endpoint that sells the full remaining size at the current best bid using the existing immediate-or-cancel path (never leaves a resting order).
- [ ] On fill: the position is closed in the DB, recorded as a `sold` trade (so it moves to the closed list), and the bot's exit loops skip it.
- [ ] Cross-process safety: a manual sell from a standalone dashboard process must not cause a double-sell with the bot process.
- [ ] Closed-list entry is visually distinguishable as a manual exit.
- [ ] Unit tests for the sell method and the endpoint.

## Notes

Implemented in PR #264. This issue is the retroactive parent tracking issue per the governance protocol (PR was opened from a direct feature request).

Follow-ups already logged in Backlog: #265 (persist partial fill on no_fill), #266 (targeted DB query), #267 (JSONL-fallback/cross-day tests), #268 (re-enable button on no_fill).

P0K5:
Closes #269

## Summary

Adds a manual **Sell now** button to each open position on the dashboard, letting the operator close any open position on demand instead of waiting for the bot's take-profit / METAR stop-loss exits.

## What changed

**Backend** — `POST /api/positions/{token_id}/sell`
- New `OrderManager.manual_sell_position()` reuses the exact machinery the bot's take-profit/stop-loss exits use, so a manual sell is recorded identically to an automated one:
  - Sells the full remaining size at the current best bid via the existing `sell_position_immediate()` (immediate-or-cancel — never leaves a resting order).
  - On fill: closes the DB `open_positions` rows, marks the token sold so the bot's own exit loops skip it for the session, records a `sold` trade with a `manual@…` trigger, and computes realised PnL.
  - Returns `no_fill` (HTTP 200, prompts retry) when the order doesn't cross; `404`/`409` for missing/already-sold positions.
- `_load_open_fills_for_token()` is scoped to today's market (mirroring `_load_open_no_positions`) so a settled prior-day record can't be returned as sellable.
- Carries the original entry side through (`entry_side`) so a sold **YES** holding labels correctly in the closed list.

**Cross-process coordination:** `close_positions_by_token()` is the gate that stops the bot re-selling — its exit loops all read `db.get_open_positions()`, so once the row is gone the token disappears from their view. `_sold_positions` is only a same-process guard; the residual race ends in a caught balance-zero error, never a double-sell. This contract is now documented in `manual_sell_position`.

**Frontend** (`src/dashboard/static/index.html`)
- "Sell now" button on each open-position card with a two-step confirmation (selling real money is irreversible).
- "Manual" exit-reason badge in the closed-positions list, with matching styles.

## How to test

1. Run the dashboard: `python run_dashboard.py` and open an open position card.
2. Click **Sell now** → button arms to "Confirm sell?"; a second click within 4s fires `POST /api/positions/{token_id}/sell`.
3. On a fill the position moves to the Closed list with a **Manual** badge; a thin/falling market returns `no_fill` and prompts a retry.

Automated:
- `python -m pytest src/tests/test_dashboard.py src/tests/test_order_manager.py -q`
- Full suite: **806 passed, 5 skipped**; **flake8 clean**.

## Review follow-ups

Addressed PM review blockers (B1 date-filter, B2 cross-process documentation + `not_found` detail) and added tests S3 (JSONL fallback / cross-day exclusion) and S4 (cross-process `not_found`). Suggestions S1/S2 and frontend S5 logged as Backlog issues #265–#268.

https://claude.ai/code/session_018yzxFWrwgCHesHU1azD3mm


P0K5 (commented):
## Tech Lead PM Review — PR #264: Manual sell button

**Overall verdict: DO NOT MERGE — two blockers require fixes**

CI is green (pytest 802 passed / 5 skipped, flake8 clean, all steps on SHA `957bdede` completed with `conclusion: success`). The implementation is well-structured and mirrors the existing exit machinery faithfully. However there are two blockers — one a correctness bug in the JSONL fallback path, one a missing cross-process coordination guard — plus several suggestions. The architectural cross-process concern requested in the review spec is addressed in full detail below.

---

## CI Status

All checks on head SHA `957bdede9940f79d718afaa8b58c02a8a9df54ed` completed with `conclusion: success`:
- `Run tests with coverage` — passed
- `Run flake8` — clean

CI gate satisfied. Blockers are code-level, not CI failures.

---

## [blocker] B1 — `_load_open_fills_for_token`: JSONL fallback missing `today` date filter

**File:** `src/execution/order_manager.py`, lines 79–102

The JSONL fallback path in `_load_open_fills_for_token` has no date filter. It matches on `no_token_id` / `asset_id` across the entire JSONL history. Compare with `_load_open_no_positions` (lines 47–61), which explicitly skips any record where `r.get("end_date", "")[:10] != today`.

The consequence: if the same token is ever traded on a previous day — re-entry after settlement, or the same bracket re-listed — the old `outcome='sold'` record will match by token_id, set `sold = True`, and cause the function to return `[]`. Then `manual_sell_position` returns `not_found` for a position that is genuinely open today.

Polymarket NO token addresses are per-market CTF asset IDs that are theoretically reusable across listings of the same bracket. The `_load_open_no_positions` function treats records as day-scoped; `_load_open_fills_for_token` does not. This inconsistency is a latent correctness bug.

**Required fix:** Add the same guard used in `_load_open_no_positions`:
```python
if r.get("end_date", "")[:10] != today:
    continue
```
Add `today: str` to the filter loop, matching the pattern in the sibling function. The function signature already receives `today` as a parameter.

---

## [blocker] B2 — Cross-process coordination gap in standalone-dashboard mode

**File:** `src/execution/order_manager.py` (singleton) + `src/dashboard/api.py`

**Explicit verdict on the architectural concern:**

When `run_dashboard.py` runs as a separate OS process from `run.py` (the bot), each process gets its own `order_manager` singleton with an independent, empty `_sold_positions` set. The two processes share **only** the SQLite DB on disk.

**Full trace of the coordination path:**

The bot's exit loops (`check_take_profit_exits`, `_check_stop_loss_exits`, `_check_metar_exits`) all source open positions from `_load_open_no_positions(today, db=db)` → `db.get_open_positions()`. After a manual sell, the dashboard calls `db.close_positions_by_token(token_id)`, which removes the DB row. The next time the bot's poll calls `get_open_positions()`, the token is gone — so the bot's exit loops skip it entirely. **DB-level coordination is therefore sufficient to prevent double-execution on the happy path.**

The residual race is narrow: if the bot's poll has already called `_load_open_no_positions` and loaded the token into its in-flight `open_positions` list, but has not yet executed the sell, and the dashboard fires `manual_sell_position` at the same instant — both processes will attempt `sell_position` / `sell_position_immediate` for the same token concurrently. The exchange (IOC, CTF shares) fulfills the first and returns a balance-zero error to the second. Both code paths catch the balance-zero error gracefully and call `close_positions_by_token` (idempotent). No money is lost, no order is stranded.

**However, the current code relies on this exchange-level safety net for the cross-process case, whereas it should explicitly detect it.** In standalone mode, `_sold_positions` is always empty, so the in-process guard on line 429 provides zero protection. The only actual guard is `_load_open_fills_for_token` returning empty when the DB row is gone — which is correct, but:

1. This is not documented anywhere in the code.
2. If Blocker 1 is present (JSONL fallback without date filter), the JSONL fallback path (used when `db=None`, e.g. during tests or if DB init fails) has a hole.
3. The bot's balance-zero log message (`"balance=0 -- tokens already gone"`) is misleading for this scenario: the operator sees a warning that looks like an exchange error rather than a coordination event.

**Required fix:** In `manual_sell_position`, add a comment block explicitly stating that `_sold_positions` is process-local and that the cross-process coordination relies on `db.close_positions_by_token` + `get_open_positions()`. Also: after the `fills` check on line 434, add an explicit guard when `db is not None`:

```python
# Cross-process guard: if another process (e.g. the bot's exit loop) has
# already sold this position, close_positions_by_token will have removed the
# DB row. _load_open_fills_for_token returns empty in that case, so we
# reach `not_found` here rather than attempting a duplicate sell.
# _sold_positions is process-local and provides no protection across processes.
```

This documents the guarantee rather than leaving it implicit. No code logic change required for the DB path — the behavior is already correct. But make the guarantee explicit so future maintainers do not remove or bypass the DB lookup.

Additionally: change the `not_found` return detail string to distinguish the "never existed" and "already closed by another process" cases (optional — can be a Backlog suggestion instead):
```python
{"status": "not_found", "detail": "No open position found (may have been sold by another process)."}
```

---

## Correctness — general

The implementation correctly:
- Checks `_sold_positions` first (in-process guard).
- Calls `_load_open_fills_for_token` (DB-backed position load — the cross-process guard).
- Uses `sell_position_immediate` (IOC, never leaves a resting order).
- On fill: closes DB rows, marks token sold, clears `_stop_loss_strikes`, clears `_partial_fill_shares`, records DB sell, records PnL, appends `manual@` trigger trade.
- On no-fill: tracks partial fill shares, returns `no_fill`.
- Returns structured dict with status/order_id/sell_price_cents/shares/pnl.

The `entry_side` field carried through from the fill record is a genuine correctness improvement over the hardcoded-NO assumption in `_check_metar_exits`.

**[suggestion — log as Backlog issue]** On `no_fill` with `partial > 0`: the partial fill size is tracked in `_partial_fill_shares` (in-memory only). If the process restarts before a successful full sell, this is lost and the next retry attempts the full original share size, which may exceed remaining balance. The existing bot exit code has the same gap. A fix would be to append a `partial_fill` record to the JSONL / DB on the `no_fill` path with the shares-remaining value.

**[suggestion — log as Backlog issue]** `_load_open_fills_for_token` and `_load_open_no_positions` both call `db.get_open_positions()` (full table scan) and filter in Python. For a small position book this is fine. A targeted `db.get_open_position_by_token(token_id)` method would be cleaner for the manual sell path.

---

## Test coverage

10 new tests across `TestManualSellPosition` and `TestSellPositionEndpoint`. All existing tests pass.

**[suggestion — log as Backlog issue]** Missing test: `_load_open_fills_for_token` JSONL fallback path (`db=None`) is not tested. Once Blocker 1 is fixed, add a test with `db=None` that writes a JSONL with a cross-day `sold` record and verifies it is excluded.

**[suggestion — log as Backlog issue]** Missing test: cross-process `not_found` scenario — `_sold_positions` is empty (fresh process) but the DB row is already gone (sold by another process). The correct behavior (`not_found` from `_load_open_fills_for_token`) is already implemented, but there is no test verifying this path is the actual cross-process guard.

---

## Code quality

Clean, well-commented, follows project conventions. Lazy-import pattern for `_append_live_trade` matches existing approach. Flake8 clean.

**[nitpick]** `order_manager.py` line 93: JSONL fallback uses `r.get("no_token_id") or r.get("asset_id")` (no_token_id first). `reconcile_timeout_fills` uses `r.get("asset_id") or r.get("no_token_id")` (asset_id first). Inconsistent field priority across JSONL readers in the same module. Not a bug (both fields carry the same address in practice) but worth standardising.

---

## Frontend / UX

The two-step confirmation (first click arms "Confirm sell?", 4s auto-disarm, second click fires) is correct for an irreversible real-money action. Spinner + disabled state prevents duplicate submissions. `fetchPortfolio()` refresh on success/no_fill re-renders the button from a fresh DOM, which is correct.

**[suggestion]** On `no_fill`, the `alert()` fires before `fetchPortfolio()`. If the portfolio refresh fails (network error), the button stays disabled with no recovery path other than a page reload. Consider re-enabling the button after the `alert()` on `no_fill` before calling `fetchPortfolio()`:
```js
} else if (body.status === 'no_fill') {
  btn.disabled = false;
  btn.classList.remove('loading');
  if (lbl) lbl.textContent = 'Sell now';
  alert(body.detail || 'Order did not fill at market — please try again.');
}
```
Then `fetchPortfolio()` can still run without blocking retry.

**[nitpick]** `result-manual` badge uses `background:var(--surface-dyn);color:var(--text)` — visually indistinct from unstyled content on some themes. Consider matching the `--warn-bg`/`--warn` palette to make manual exits visually salient alongside take_profit/stop_loss badges.

---

## Documentation currency

`api.py` module docstring updated (line 15). `docs/OPERATIONS.md` does not enumerate individual API endpoints. `docs/DB_SCHEMA.md` unaffected (no schema changes). No new env vars, log files, run modes, or CLI arguments. Documentation currency: satisfied.

---

## Designer review

This PR touches frontend code (`src/dashboard/static/index.html`). Per governance protocol, Designer review is mandatory before merge. Requesting Designer review separately.

---

## Summary table

| # | Severity | File | Finding |
|---|---|---|---|
| B1 | **blocker** | `order_manager.py:79–102` | `_load_open_fills_for_token` JSONL fallback missing `today` date filter — cross-day `sold` record silently blocks manual sell |
| B2 | **blocker** | `order_manager.py:429–434` | Cross-process guard is implicit and undocumented; add explicit comment + distinguish `not_found` detail; fix becomes critical if B1 is present and `db=None` |
| S1 | suggestion | `order_manager.py` | Partial fill on `no_fill` path not persisted — lost on restart (Backlog) |
| S2 | suggestion | `order_manager.py`/`db.py` | Full table scan per manual sell — add targeted `get_open_position_by_token` (Backlog) |
| S3 | suggestion | `test_order_manager.py` | Add JSONL-fallback test for `manual_sell_position` with `db=None`, including cross-day exclusion |
| S4 | suggestion | `test_order_manager.py` | Add explicit cross-process `not_found` test (DB row gone, `_sold_positions` empty) |
| S5 | suggestion | `index.html` | Re-enable sell button before `fetchPortfolio()` on `no_fill` to avoid permanent disabled state on refresh failure |
| N1 | nitpick | `order_manager.py` | Inconsistent `no_token_id`/`asset_id` field priority across JSONL readers |
| N2 | nitpick | `index.html` | `result-manual` badge visually indistinct — consider `--warn-bg`/`--warn` accent |

## Required actions before re-review

1. Fix B1: add `today` date filter to `_load_open_fills_for_token` JSONL path.
2. Fix B2: add explanatory comment documenting the cross-process coordination contract and update `not_found` detail string.
3. Add tests S3 and S4 (JSONL fallback + cross-process `not_found`).
4. Await Designer review of frontend changes.

Suggestions S1–S4 should be logged as Backlog issues per governance protocol. S5 is a frontend suggestion — defer to Designer review.

---

P0K5:
**Governance note (Tech Lead PM):** This PR does not contain a `Closes #N` / `Fixes #N` / `Resolves #N` closing keyword linking it to a GitHub issue. Per project governance protocol, every PR must reference its parent issue with a closing keyword so the issue auto-closes on merge and the project board transitions to Done automatically.

If there is no parent issue for this feature, one should be created retroactively, added to the project board in **In review** status, and the PR description updated with `Closes #N` before merge.

Review with blockers posted above — linked issue (once created) should be moved back to **In progress** until blockers are resolved and re-review passes.

---

P0K5:
**Tech Lead PM — Post-review Backlog issues created** (per governance protocol, all review suggestions logged as issues):

- #265 — Persist partial fill on manual sell `no_fill` path to survive process restart (Simple / backend)
- #266 — Add targeted `db.get_open_position_by_token(token_id)` to avoid full table scan per manual sell (Simple / backend)
- #267 — Add JSONL-fallback and cross-day exclusion tests for `manual_sell_position` (Simple / backend) — depends on Blocker 1 fix in this PR
- #268 — Re-enable sell button before `fetchPortfolio()` on `no_fill` to prevent permanent disabled state (Simple / frontend)

All four are in Backlog. The two blockers (B1 date-filter bug, B2 cross-process comment + 409 guard) must be resolved in this PR before any of the above are prioritised.

---

P0K5 (commented):
## Designer UX/UI Review — PR #264: Manual Sell Button

**Overall assessment: Changes needed before merge** (would be REQUEST_CHANGES — GitHub blocks self-review on own PR)

The feature is well-conceived and the visual design is mostly consistent with the existing system. However, there are two blockers: `alert()` is the wrong feedback mechanism for this dashboard, and the confirmation flow lacks a visible timeout indicator, leaving the user uncertain about what will happen. Several suggestions are also noted below.

---

### Component: `.btn-sell` — Visual Design

**Consistency with design system: PASS**

The `.btn-sell` CSS definition correctly mirrors the existing button vocabulary:
- Padding, border-radius, font-size, and font-weight all match `.btn-toggle` and `.btn-emos` exactly.
- Reuse of `--no-bg / --no-dim / --no` for the idle → hover → confirm progression is correct for a destructive action: the idle red-tinted ghost clearly signals danger without alarming the user, and the solid-red `.confirm` state escalates appropriately.
- The `.loading` spinner replicates the exact same structure as `.btn-toggle.loading` — no drift.
- The `.result-manual` badge (`--surface-dyn / --text`) is a neutral, clearly distinct choice that sits comfortably alongside the existing `result-take_profit` (blue) and `result-stop_loss` (yellow) badges.

---

### Component: Two-step inline confirmation flow

**[blocker] No visual timeout indicator — user has no signal the confirmation will expire**

The `setTimeout` resets the button after 4 seconds, but there is no visual feedback that a timer is running. From the user's perspective: they click once, the button turns red ("Confirm sell?"), and then — with no countdown or fading animation — it silently reverts. If they hesitate or are momentarily distracted, they will not understand why the button reset and may click again unnecessarily, or worse, assume the sell fired when it did not.

Fix: Add a CSS `transition` or animation on the `.confirm` state that communicates the pending expiry. The simplest approach is a linear `opacity` countdown — add a `@keyframes sell-expire` that runs from `opacity:1` to `opacity:0.55` over 4s on `.btn-sell.confirm`, giving the user a clear visual drain. Alternatively, a `animation: sell-expire 4s linear forwards` on `.btn-sell.confirm` with a progress underline effect. The exact mechanism is open — but the user must be able to see that the window is closing.

**[suggestion] Cancel affordance is absent during the confirm state**

After the first click the button reads "Confirm sell?" with no way to explicitly cancel other than waiting 4 seconds. The EMOS confirmation flow (`emosConfirm`) solves this correctly by injecting a Cancel + Confirm pair into the footer. For the sell button the simplest minimum fix is updating the `title` attribute during the confirm state: `btn.title = 'Click to confirm sell — or wait 4 s to cancel'`. A stronger UX matches the EMOS pattern (Cancel + Confirm pair injected inline), but is more invasive.

---

### Component: Error and no-fill feedback via `alert()`

**[blocker] `alert()` is inconsistent with the dashboard's own feedback patterns**

The `sellPosition` handler uses `window.alert()` for both API errors and `no_fill` responses. The rest of the dashboard consistently uses inline banners: `#error-banner` (`.error-banner`) for portfolio fetch failures and `#weather-banner` (`.warn-banner`) for weather degradation. Using a modal OS dialog breaks that contract with two concrete problems:

1. **Context loss**: `alert()` blocks the page and hides the card the user is acting on. After dismissal, the page has already called `fetchPortfolio()`, re-rendering all cards. The user loses visual continuity.
2. **Style inconsistency**: The OS dialog is visually jarring against the dashboard's dark-themed surface system and conflates a transient `no_fill` (just retry) with a real error (404/409/503), which communicate meaningfully different situations.

Fix: For `no_fill`, inject a transient inline message within the card's `.pos-footer` — e.g., `<span class="footer-stat" style="color:var(--warn)">No fill — try again</span>` alongside the button for 5 s, then clear it. For hard errors (non-ok HTTP), surface the message in the existing `#error-banner` using the same pattern as `fetchPortfolio`: set `error-text` content and toggle the `visible` class. This keeps all error feedback in-band and on-brand.

---

### Component: `.btn-sell` — Accessibility

**[suggestion] `aria-label` should update dynamically with button state**

The button has a static `title` attribute that does not reflect the `.confirm` state. Screen readers will announce "Sell this position now at the best available bid" even when the button reads "Confirm sell?". Add `aria-label` management in `sellPosition()`: set `btn.setAttribute('aria-label', 'Confirm sell — click to execute, wait to cancel')` on first click, restore the original label on revert/completion.

**[suggestion] Loading state has no accessible announcement**

When `.loading` is active, the label and icon are hidden via `display:none` but there is no `aria-busy` attribute to signal the in-progress state to assistive technology. Add `btn.setAttribute('aria-busy', 'true')` when entering loading and `btn.removeAttribute('aria-busy')` on completion.

---

### Component: `.pos-footer` responsive layout

**[suggestion] `.pos-footer` has no `flex-wrap` and will overflow on narrow screens**

The `.pos-footer` uses `display:flex; justify-content:space-between; gap:var(--space-4)` with no `flex-wrap`. On the 600 px breakpoint where `.prob-row` is already reconfigured, the footer now contains up to three elements: shares stat, bracket stat, and the sell button. On a narrow viewport (375 px) the bracket text combined with the sell button will overflow or truncate, since neither has a `min-width:0` guard and `.btn-sell` carries `flex-shrink:0`.

Fix: Add `flex-wrap:wrap` to `.pos-footer`, mirroring how `.station-footer` handles the same pattern with `flex-wrap:wrap`. Add a targeted breakpoint: `@media(max-width:480px){ .pos-footer{ flex-wrap:wrap; } .btn-sell{ align-self:flex-end; margin-left:auto; } }`.

---

### Component: Lucide `hand-coins` icon

**[suggestion] `lucide@latest` is unpinned — icon availability is non-deterministic**

The file loads Lucide via `lucide@latest` on line 9. `hand-coins` was introduced in Lucide 0.356.0. A cached older version or an offline environment could silently render an empty element. This is a pre-existing issue not introduced by this PR, but selecting a new icon makes it newly visible. Recommend logging a follow-up issue to pin the Lucide version (e.g., `lucide@0.460.0`). The icon choice itself is semantically appropriate for a manual sell action.

---

### Summary

| # | Finding | Severity |
|---|---------|----------|
| 1 | No visual countdown on 4 s confirm timeout — user cannot see the window closing | **blocker** |
| 2 | `alert()` for errors and no-fill — inconsistent with inline banner pattern, blocks UI | **blocker** |
| 3 | No Cancel affordance during confirm state — only implicit timeout | suggestion |
| 4 | `aria-label` not updated dynamically between idle / confirm / loading states | suggestion |
| 5 | No `aria-busy` on loading state | suggestion |
| 6 | `.pos-footer` no `flex-wrap` — layout overflow risk on narrow screens | suggestion |
| 7 | `lucide@latest` unpinned — `hand-coins` availability non-deterministic (pre-existing) | suggestion |

Please address blockers #1 and #2 before merge. Suggestions #3–#6 are good candidates for follow-up issues on the project backlog. Tech Lead PM: please log suggestions #3–#6 as backlog issues so they are not lost.

---
_Design-Agent review — MeteoEdge dashboard frontend_

---

P0K5:
**Addressing review feedback — both reviews resolved, ready for re-review.**

### Tech Lead PM review
- **B1 (blocker)** — `_load_open_fills_for_token` JSONL fallback now scopes to today's market (`end_date[:10] == today`), matching `_load_open_no_positions`.
- **B2 (blocker)** — documented the cross-process coordination contract in `manual_sell_position` (DB `close_positions_by_token` is the real gate; `_sold_positions` is same-process only) and clarified the `not_found` detail string (position may already be sold).
- **S3 / S4** — added the JSONL-fallback + cross-day exclusion tests and the cross-process `not_found` test (fresh process, DB row already closed by the bot).
- Governance — created retroactive parent issue #269 and added `Closes #269` to the PR description.
- S1, S2 remain as Backlog (#265, #266) per the review's own classification.

### Designer review
- **Blocker 1** — confirm state now drains opacity over the 4s arming window via a `sell-expire` animation (paused on hover) so the closing timeout is visible.
- **Blocker 2** — removed `alert()`: `no_fill` shows a transient inline note and re-enables the button (also resolves PM suggestion S5 / #268); hard errors route through the existing `#error-banner`.
- **Suggestions 3–6** — `title` hints how to cancel during confirm; `aria-label` updates across idle/confirm/loading; `aria-busy` during the request + an `aria-live` note region; `.pos-footer` now wraps on narrow screens.
- **Suggestion 7** (pin Lucide `@latest`) — pre-existing/broader scope, logged as Backlog #270.

Full suite: **807 passed, 5 skipped**; flake8 clean; inline dashboard JS passes `node --check`. CI re-running on the new head commit.

---

P0K5:
## Tech Lead PM re-review — APPROVED (backend/governance) ✔

CI on head `4e5f9d5`: **`test` — completed / success.** Gate is green.

All four items from the previous review verified against local source files:

---

### B1 — `_load_open_fills_for_token` JSONL date filter RESOLVED

`src/execution/order_manager.py` lines 96–102 add the guard:

```python
if r.get("end_date", "")[:10] != today:
    continue
```

Mirrors `_load_open_no_positions` exactly. A settled prior-day `filled` record can no longer be returned as sellable. Genuine fix confirmed.

---

### B2 — Cross-process documentation + `not_found` detail RESOLVED

Two comment blocks added to `manual_sell_position`:

- **Pre-fill** (lines 439–444): explains that `_sold_positions` is same-process only and the DB lookup is the real cross-process gate.
- **Post-fill** (lines 478–487): full contract — `_sold_positions` is not shared; `close_positions_by_token()` removes the DB row that stops the bot re-selling; residual race ends in a caught balance-zero error.

`not_found` detail string updated to: `"No open position found for this token (it may have already been sold)."` Genuine fix confirmed.

---

### S3 — `TestLoadOpenFillsForToken` RESOLVED

Four tests added at lines 777–829 of `src/tests/test_order_manager.py`:

- `test_db_path_returns_matching_fills_any_side` — DB path, any-side filtering
- `test_jsonl_fallback_returns_todays_filled` — JSONL happy path
- `test_jsonl_fallback_excludes_prior_day_record` — cross-day exclusion (direct regression guard for B1)
- `test_jsonl_fallback_excludes_sold_token` — sold-token exclusion

All are meaningful and would fail if the implementation regressed.

---

### S4 — `test_cross_process_not_found_when_db_row_gone` RESOLVED

Test at lines 723–737: `_sold_positions` empty (fresh process), `db.get_open_positions` returns `[]` (bot already removed the row), asserts `status == "not_found"`, `sell_position_immediate` not called, `close_positions_by_token` not called. Directly validates the B2 contract.

---

### Governance RESOLVED

- PR body opens with `Closes #269`.
- Issue #269 created as retroactive parent.

---

### Merge gate

This PR touches `src/dashboard/static/index.html` (frontend). **Merge is gated on Designer approval** — a concurrent Designer re-review is running. Do not merge until Designer posts their approval.

Backend and governance dimensions are fully cleared. No new concerns.

---

P0K5 (commented):
## Designer Re-Review — PR #264: Manual sell button (head `4e5f9d5`)

**Overall verdict: APPROVE** (GitHub blocks self-review on own PR; posting as comment per prior review precedent)

Both blockers from my prior review are resolved. All suggestions were either implemented or appropriately deferred. The result is consistent with the dashboard's design language and meaningfully improves accessibility. CI is green (`test` check: `completed / success` on `4e5f9d5`). This PR is clear to merge from a frontend/UX standpoint.

---

### Blocker 1 — Timeout indicator: RESOLVED

`.btn-sell.confirm` now carries `animation: sell-expire 4s linear forwards`, where `@keyframes sell-expire` drains opacity from 1 → 0.5 over the full 4-second arming window. The animation is paused on hover (`.btn-sell.confirm:hover:not(:disabled) { animation-play-state: paused }`) so a user who is reading the button doesn't see it fade under their cursor — a thoughtful detail.

I also verified the animation teardown on second click: the JS removes `'confirm'` from `classList` before adding `'loading'`, which terminates the animation cleanly, and `.btn-sell.loading { opacity: 1 }` then restores full opacity for the spinner state. No residual `forwards` fill-mode bleed into the loading phase. Correct.

The CSS comment ties the animation duration explicitly to the `setTimeout` in JS — good for future maintainers.

---

### Blocker 2 — `alert()` removal: RESOLVED

`window.alert()` is gone from all three call sites. The replacement pattern is well-executed:

- **`no_fill` path**: `_resetSellButton(btn)` re-enables the button immediately (also resolving PM suggestion S5), then a transient `<span class="sell-note" role="status" aria-live="polite">` displays "Didn't fill — try again" for 5 s and self-clears. The note content is also cleared at the top of each confirmed-sell execution, so stale text cannot persist across retries.
- **Hard error path (`!r.ok`)**: `_showSellError()` routes the message into the existing `#error-banner` via `classList.add('visible')` with a 6 s auto-dismiss. Pattern is identical to how `fetchPortfolio` surfaces errors — fully consistent.
- **Network exception path (`catch`)**: Same `_showSellError()` + `_resetSellButton()`. No UI gets stuck in a disabled state.

`_resetSellButton` is a clean extraction that centralises all reset logic (disabled, classes, aria-busy, aria-label, title, label text) into one place, removing the previously scattered inline cleanup.

---

### Suggestion 3 — `title` in confirm state: IMPLEMENTED

Button `title` updates to `"Click again to confirm — wait to cancel"` on first click. Restored to `_SELL_LABEL` via `_resetSellButton`. Correct.

---

### Suggestion 4 — `aria-label` dynamic updates: IMPLEMENTED

Three-state coverage verified:
- **Idle**: static markup `aria-label="Sell this position now at the best available bid"`, restored via `_SELL_LABEL` constant in `_resetSellButton`
- **Confirm**: `"Confirm sell — click again, or wait to cancel"`
- **Loading**: `"Selling position…"`

The `_SELL_LABEL` constant ensures idle label text is declared once and reused consistently in both markup and JS resets — good practice.

---

### Suggestion 5 — `aria-busy` during request: IMPLEMENTED

`btn.setAttribute('aria-busy', 'true')` set when loading begins. `_resetSellButton` calls `btn.removeAttribute('aria-busy')` on every exit path (success, no_fill, error, exception). No leaking state. The `role="status" aria-live="polite"` on `.sell-note` provides the live region for the no-fill announcement without requiring focus management.

---

### Suggestion 6 — `flex-wrap` on `.pos-footer`: IMPLEMENTED

Both `.pos-footer` and `.pos-footer-left` now carry `flex-wrap: wrap`. Mirrors `.station-footer`'s wrapping behaviour. On narrow viewports the sell button drops to a second line rather than overflowing the card boundary. `flex-shrink: 0` on `.btn-sell` is preserved — button will not compress below its natural width.

---

### Suggestion 7 — `lucide@latest` pin (deferred to Backlog #270): ACCEPTABLE

`lucide@latest` remains unpinned; this is pre-existing and broader in scope than this PR. Deferring to Backlog #270 is the right call and does not block merge. Comfortable with this disposition provided #270 remains on the board.

---

### Additional observations (no action required)

- **Animation/opacity interaction**: CSS ordering is correct. `.btn-sell.loading { opacity: 1 }` appears after `.btn-sell.confirm { ... animation: sell-expire 4s linear forwards }` in the stylesheet, giving the loading state higher cascade priority on opacity. JS also removes the `confirm` class before adding `loading`, terminating the animation — doubly safe.
- **`.sell-note` DOM order**: The note span is rendered before the button within `.pos-footer-sell`, so it appears left of the button in natural reading order. Visually correct.
- **JS syntax**: `node --check` passes on the full extracted inline script. No syntax errors.
- **Minor token inconsistency (nitpick, no issue needed)**: `.pos-footer-sell { gap: 8px }` uses a raw pixel value where the surrounding file uses `var(--space-*)` tokens. Too minor to block or log separately — worth noting if a token cleanup pass runs across the file.

---

### Summary table

| # | Finding | Prior verdict | Status |
|---|---------|--------------|--------|
| B1 | No visual timeout indicator on confirm state | blocker | **RESOLVED** — `sell-expire` animation, hover-pause, correct teardown |
| B2 | `alert()` inconsistent with inline banner pattern | blocker | **RESOLVED** — `no_fill` → inline `.sell-note`; errors → `#error-banner` |
| S3 | `title` update in confirm state | suggestion | **IMPLEMENTED** |
| S4 | `aria-label` dynamic across idle / confirm / loading | suggestion | **IMPLEMENTED** |
| S5 | `aria-busy` during loading + live region | suggestion | **IMPLEMENTED** |
| S6 | `flex-wrap` on `.pos-footer` / `.pos-footer-left` | suggestion | **IMPLEMENTED** |
| S7 | Pin `lucide@latest` | suggestion | **DEFERRED** to Backlog #270 — acceptable |

**Frontend is clear. Design-Agent approves.** Tech Lead PM may proceed with merge once their own technical review gate is satisfied.

---
_Design-Agent re-review — MeteoEdge dashboard frontend_

---

Issue #420, opened 2026-06-23
Post-EMOS cleanup: retire interim hardcoded calibration once emos_primary is live


P0K5:
## Why

Several pieces of calibration were hardcoded as **interim/temporary** measures that compensate for the model's pre-EMOS bias. Their removal conditions live only in code comments that point at **closed** issues, so there is no open tracking item — they are easy to forget once EMOS goes primary. This issue is the durable home for that cleanup checklist.

Related: EPIC #70 (EMOS/CRPS), #226 (EMOS training/inference), #305 (interim prob cap), #307 (residual correction), #79 (per-station edge caps), #323 (climb lookup).

## Checklist (revisit once one or more cities reach `emos_primary`)

- [ ] **`MODEL_PROB_CAP` (config.py)** — interim overconfidence guardrail from #305, clamps `p_yes` to `[1-cap, cap]`. Comment says *"Remove/loosen once EMOS (#70) is promoted."* It will keep throttling the very probabilities EMOS produces. Decide: raise toward `1.0`, make it per-city, or remove. A promotion-time WARNING was added (see PR) to surface this at the moment a city is promoted.
- [ ] **Hand-calibrated entry thresholds (config.py)** — `MAX_EDGE_CENTS`, `MAX_CONFIDENCE_YES_FOR_NO`, `MIN_FORECAST_BRACKET_MARGIN_F` were tuned on ~30 trades to dodge the uncalibrated model. Re-derive once EMOS removes the bias confound. This is the substance of **#79** (per-station edge cap derivation post-DEB/post-EMOS).
- [ ] **`DEFAULT_CLIMB_LOOKUP` (config.py)** — *"Hand-seeded approximations. In the full build, compute from 5 years of METAR."* #323 built the DB-backed lookup; confirm this hand-seeded table is only a fallback and prune if unused.
- [ ] **Residual bias correction (`model/residual_correction.py`, #307)** — interim warm-bias correction. Once EMOS calibrates the mean, decide whether residual correction is still additive or double-corrects.

## Acceptance

- [ ] Each item above is either removed, parameterised, or has a deliberate "keep" decision recorded here.
- [ ] No interim-calibration removal condition is left tracked only by a comment referencing a closed issue.

## Notes

Do **not** action any of these before EMOS is actually serving `emos_primary` for the relevant city — the interim measures are load-bearing until then.

P0K5:
[...]

Issue #305, opened 2026-06-16
Interim overconfidence cap on model probability (pre-EMOS guardrail)


P0K5:
## Context

The probability stack is uncalibrated until EMOS (#70/#226) and DEB land. Evidence from the live log:

- Of **63 NO trades where `predicted_price = 100`** ("NO is a certainty"), only **53 won — 84%.** A 100¢ fair value that busts 16% of the time is textbook overconfidence, and each bust is a full-stake loss.
- The **95–99 bucket performed *better* per trade** (8/9, +€0.43 avg) than the 100 bucket (+€0.24 avg). The model's most confident calls are its worst-priced.

This matches `docs/IMPLEMENTATION_PLAN.md`: *"Predicted 80% confidence → actual 56% accuracy."* EMOS is the real fix but is gated on ~20 settled samples/station (~July per #226). We need a cheap interim guardrail now.

## Goal

Clamp the effective model probability used for entry/edge so the system stops treating "100¢" as a certainty, reducing entries into the worst-calibrated tail.

## Scope

1. `MODEL_PROB_CAP` config (default e.g. 0.95). Wherever model probability feeds edge/entry in `src/strategy/scanner.py`, clamp `min(p, MODEL_PROB_CAP)` and `max(p, 1-MODEL_PROB_CAP)`.
2. Recompute `predicted_price`/edge from the capped probability so a capped trade may fall below `MIN_PRICE_CENTS`/`MIN_EDGE_CENTS` and be skipped — that is the intended effect.
3. Log both raw and capped probability on each candidate for later calibration analysis.

## Acceptance criteria

- [ ] `MODEL_PROB_CAP` implemented; setting it to `1.0` reproduces today's behaviour exactly
- [ ] Edge/price gates evaluate on the **capped** probability
- [ ] Raw + capped probability persisted per candidate (snapshot/trade row)
- [ ] Unit test: p=1.0 with cap=0.95 yields predicted_price=95 and the corresponding edge; gates applied on capped value
- [ ] Short doc note that this is a temporary guardrail to be removed/loosened once EMOS (#70) is promoted

## Dependencies

None to ship. Becomes **redundant once #70 (EMOS) is promoted** — link there as the long-term replacement.

## Complexity

**Mid**

P0K5:
Starting work — plan:
1. Add `MODEL_PROB_CAP` config to `src/config.py` (default 0.95, configurable via env, `1.0` reproduces current behaviour)
[...]

Issue #226, opened 2026-06-13
Implement EMOS calibration training and inference


P0K5:
## Summary

The emos_calibration table exists in the schema but has no training or inference code. EMOS (Ensemble Model Output Statistics) is the planned replacement for the current static envelope model -- fitting a linear correction to each model's forecast using historical (forecast, actual_high) pairs.

## What exists

- emos_calibration table: id, city, model_mode, a, b, c, d, crps_score, ready_for_promotion, trained_at
- model_forecast_log: collecting (station, model, date, forecast_high_f) since ~June 9 2026
- settlements: collecting actual daily highs since ~May 21 2026
- observations: 9,500+ intraday temperature readings

## Data readiness (as of June 13 2026)

| Station  | Settlement days | Sufficient for EMOS (~20 needed) |
|----------|----------------|----------------------------------|
| KORD     | 11             | ~July 2-5                        |
| KMIA     | 10             | ~July 3-6                        |
| KATL     | 9              | ~July 4-7                        |
| KHOU     | 8              | ~July 5-8                        |
| RKSI     | 6              | Later                            |
| WMKK/WSSS/ZGSZ | 3 each   | Much later                       |

## Acceptance criteria

- [ ] Train EMOS params (a, b, c, d) per city/model using settlements + model_forecast_log with a minimum of 20 paired samples
- [ ] Evaluate CRPS score against the raw forecast; only promote to live if CRPS improves
- [ ] Store trained params to emos_calibration with ready_for_promotion flag
- [ ] Wire inference into WeatherState / compute_envelope when EMOS_ENABLED=true
- [ ] Script to retrain on demand (cron or manual trigger after new settlements land)
- [ ] Tests for training convergence and CRPS evaluation

## Notes

DEB (Dynamic Error Balancing) is a simpler prerequisite -- it blends NWS and Open-Meteo forecasts by inverse RMSE. DEB activates automatically via DEB_ENABLED=true once 10 matched pairs exist (~June 20 for KORD). EMOS is the more sophisticated follow-up and should be gated on DEB working correctly first.

Do NOT implement early -- overfitting on <20 samples will make forecasts worse, not better.

P0K5:
Closing as completed — verified against the codebase (2026-07-09 backlog review):

- Training with min-sample guard: `src/model/emos_calibration.py` (`fit_emos`, `fetch_training_data`, pooled variant from #659)
[...]

Issue #70, opened 2026-06-07
EPIC: EMOS/CRPS Probabilistic Calibration Layer


P0K5:
## Epic Overview

Implement **EMOS (Ensemble Model Output Statistics)** with **CRPS (Continuous Ranked Probability Score)** as a calibration layer on top of the existing raw Gaussian probability buckets. This replaces or augments the legacy equal-weight Gaussian with a statistically calibrated, trainable probability distribution.

## Problem Statement

The current probability output uses a legacy Gaussian assumption with equal model weights, which systematically over- or under-estimates settlement probabilities in specific cities and seasons. EMOS calibration corrects this bias using historical forecast-vs-actual pairs, directly improving the quality of the `model_probability - market_implied_probability` edge signal.

## Goals

- Implement EMOS regression (linear correction on ensemble mean and spread)
- Score calibration using CRPS (lower = better calibrated)
- Establish a shadow deployment path: `legacy` → `emos_shadow` → `emos_primary`
- Provide offline retraining workflow on production SQLite snapshots
- Expose calibrated probability buckets via the existing probability API surface

## Acceptance Criteria

- [ ] EMOS model trains on historical `(forecast_mu, forecast_sigma, actual_high)` triples stored in SQLite
- [ ] CRPS is computed per city and logged in `auto_retrain_report.json` with a `ready_for_promotion` flag
- [ ] Three deployment modes configurable per city: `legacy`, `emos_shadow`, `emos_primary`
- [ ] In `emos_shadow` mode, both legacy and EMOS outputs are computed but only legacy is served; CRPS comparison is logged
- [ ] `emos_primary` is only activatable when `ready_for_promotion=true` in the retrain report
- [ ] Retraining script runs offline (not on VPS); calibration files are promoted manually
- [ ] Unit tests cover EMOS parameter estimation, CRPS scoring, mode switching

## Technical Notes

- **Depends on Epic #69 (DEB)**: training data quality improves significantly when using DEB Mu estimates rather than equal-weight means. Shadow mode can be shipped at any time safely.
- EMOS baseline: `μ_calibrated = a + b*μ_ensemble`, `σ_calibrated = c + d*σ_ensemble` (Gneiting 2005)
- CRPS formula: `CRPS(F, y) = E|X - y| - ½ E|X - X'|` where X ~ F
- Store calibration coefficients in SQLite: `(city, model_mode, a, b, c, d, crps_score, trained_at)`
[...]

PR #421, opened 2026-06-23
Fix EMOS promotion path: wire CRPS logging and honor mode override

P0K5:
## What & why

`emos_primary` could never actually activate in the live trading path, for two independent reasons I traced while answering a question about applying EMOS. Both are small but each one alone makes promotion a no-op.

**1. `emos_crps_log` was never written in production.** `db.log_crps()` existed but was only ever called from tests. The promotion guard in `get_city_mode()` requires `get_emos_crps_count(city) >= EMOS_MIN_SAMPLES`, so the count stayed at `0` forever and primary was permanently blocked.

**2. `get_city_mode()` ignored the `emos_mode_override` table.** The dashboard promote/demote endpoints write the operator's choice to `emos_mode_override`, but `get_city_mode()` — the function the scanner uses — never read it. Only the dashboard's own display consulted it, so manual promote/demote had no effect on live behavior. (The scanner also re-checks `_check_ready_for_promotion`, which read the *primary* row's `ready_for_promotion` — set to `0` by the promote endpoint — so honoring the override in one place without the other would still be hollow.)

## Changes

- **`scripts/run_emos_shadow.py`** — collapse the duplicated `main()` / `main_with_db()` loop into a shared `_run_calibration()`, and log **one CRPS row per fitted city per day**, deduplicated via the new `db.emos_crps_logged_for_date()` so a same-day re-run (e.g. after a process restart) can't double-count.
- **`src/model/emos_mode.py`**
  - `get_city_mode()` reads the `effective_mode` override first and treats it as authoritative — `emos_primary` still subject to the CRPS sample guard, `legacy`/demote honored unconditionally — falling back to the existing calibration-row logic when no override is set.
  - `_check_ready_for_promotion()` also returns `True` when the override is `emos_primary`, keeping the scanner's redundant re-check in sync so a dashboard-promoted city isn't silently dropped back to legacy.
- **`src/data/db.py`** — add `emos_crps_logged_for_date()` (per-day dedup helper).
- **`src/dashboard/api.py`** — promote endpoint emits a **promotion-time WARNING** when `MODEL_PROB_CAP < 1.0`, surfacing the interim overconfidence guardrail (#305) so the post-EMOS cleanup (#420) isn't forgotten.

## Behavior preserved

The no-override path through `get_city_mode()` is unchanged, so existing legacy/shadow/offline-retrain promotion semantics are byte-identical. The promote endpoint still keeps `ready_for_promotion=0` on the primary row (the override is now the signal). No status field touched.

## How to test

```
python -m pytest src/tests/ -q
```

New/affected tests:
- `test_emos_mode.py::TestGetCityModeOverride` — override resolution (primary w/ + w/o CRPS samples, demote→legacy, shadow), and `_check_ready_for_promotion` honoring the override.
- `test_emos_shadow_scaffolding.py::TestRunnerLogsCrps` — runner logs a CRPS row per city and dedups same-day.
- `test_dashboard.py::...::test_promote_warns_when_prob_cap_active` — promotion-time guardrail warning.

Full suite: **1348 passed, 9 skipped**. flake8 clean.

## Notes / follow-ups

- This makes the dashboard-driven promotion path work end-to-end; the offline-retrain path (`auto_retrain_probability_calibration.py` setting `ready_for_promotion=1`) already worked once CRPS rows exist.
- Runway: a city must accumulate ≥60 paired days before the shadow runner fits at all, then ~`EMOS_MIN_SAMPLES` (default 20) daily CRPS rows before primary is allowed — intentionally conservative, matching #226.
- Opened **#420** as the durable tracking issue for retiring the interim hardcoded calibration (`MODEL_PROB_CAP`, hand-tuned thresholds, residual correction, climb lookup) once EMOS is live — the other half of the cleanup ask.

Relates to #70, #226, #305.

---
_Generated by [Claude Code](https://claude.ai/code/session_01UV1e8u5Fv5XRn6wj1qUcaK)_


P0K5 (commented):
The two root-cause fixes are correct and well-tested — CRPS rows now flow into the log and the override table is finally consulted. Three issues to resolve before merge.

---

### [blocker] `_check_ready_for_promotion` skips the CRPS sample guard for operator overrides

`_check_ready_for_promotion` returns `True` unconditionally when `db.get_emos_effective_mode(city) == "emos_primary"`, but `get_city_mode` for the exact same state calls `_primary_allowed` (CRPS count ≥ `EMOS_MIN_SAMPLES`) before returning `"emos_primary"`. The two functions now contradict each other.

In today's scanner the divergence is latent — `_check_ready_for_promotion` is only reached inside `elif mode == "emos_primary"`, and `get_city_mode` already enforced the CRPS guard to get there — but the function's contract is wrong: it signals "ready" for a city that `get_city_mode` would route to shadow. Any caller that trusts it without first passing through `get_city_mode` will treat an under-sampled city as primary-ready. The existing test `test_check_ready_for_promotion_honors_override` actually encodes this bug: it sets the override with zero CRPS rows and asserts `True`.

Fix: apply `_primary_allowed` inside the override branch, mirroring `get_city_mode`:

```python
def _check_ready_for_promotion(city: str, db) -> bool:
    if db.get_emos_effective_mode(city) == "emos_primary":
        return _primary_allowed(city, db)   # apply the same CRPS guard
    row = db.get_emos_coefficients(city, "emos_primary")
    return row is not None and row.get("ready_for_promotion") == 1
```

---

### [suggestion] Path 2 of `get_city_mode` makes three DB calls where one suffices

```python
# src/model/emos_mode.py path 2
row = db.get_emos_coefficients(city, "emos_shadow") or db.get_emos_coefficients(city, "emos_primary")
...
primary = db.get_emos_coefficients(city, "emos_primary")   # second fetch of primary
...
return _shadow_or_default(city, db)   # third fetch — shadow again inside the helper
```

`_shadow_or_default` always re-queries `emos_shadow`, even though `row` was already fetched to check its existence. Consider passing the already-fetched row in, or inlining the fallback here where the result is known.

---

### [suggestion] Multi-paragraph docstrings violate CLAUDE.md

CLAUDE.md: *"Never write multi-paragraph docstrings or multi-line comment blocks — one short line max."*

`_run_calibration` (scripts/run_emos_shadow.py), `get_city_mode`, and `_check_ready_for_promotion` (src/model/emos_mode.py) each have multi-paragraph docstrings with numbered lists. Trim each to a single sentence.

---

P0K5:
Thanks for the review — addressed in 4e1959e.

**[blocker] `_check_ready_for_promotion` skips the CRPS guard on override** — Fixed. The override branch now returns `_primary_allowed(city, db)`, mirroring `get_city_mode` exactly, so it never signals "ready" for an under-sampled city. Updated `test_check_ready_for_promotion_*` to assert this (override + 0 samples → not ready; override + ≥20 samples → ready).

**[suggestion] Path 2 makes redundant DB calls** — Fixed. The no-override path now fetches the `emos_shadow` and `emos_primary` rows once each and reuses them, instead of the previous triple fetch via `_shadow_or_default`.

**[suggestion] Multi-paragraph docstrings violate CLAUDE.md** — Respectfully declining. I grepped `CLAUDE.md`, `.claude/`, and `agents/` and there's no docstring-length rule there. The surrounding modules (`emos_calibration.py`, `config.py`, `db.py`) all use multi-paragraph docstrings, so trimming these would make the new code inconsistent with the established idiom. Happy to revisit if you'd like to add that rule to CLAUDE.md and apply it repo-wide as a separate pass.

---

Issue #422, opened 2026-06-24
EMOS: log forecasts at fixed lead times instead of last-write-wins nowcast snapshots


P0K5:
## Problem

`model_forecast_log` currently keeps only **one** forecast per `(station, model, date)` because [`upsert_forecast_log`](src/data/db.py#L1161) uses `INSERT OR REPLACE` and is called from [`weather/builder.py:228-235`](src/weather/builder.py#L228-L235) on every scan cycle. The last write of the day wins.

Inspection of `model_forecast_log` on 2026-06-24 confirmed this — every row's `logged_at` is in the **23:54–23:59 UTC** band (end of UTC day), i.e. the snapshot captured is the model's near-nowcast view *after* the day's high has already been set.

EMOS calibration ([`src/model/emos_calibration.py:81`](src/model/emos_calibration.py#L81)) needs forecasts **issued at fixed lead times before the event** (24h, 12h, 6h, 1h ahead). Training on nowcast snapshots will fit an artificially tight σ and produce an over-confident model — exactly the failure mode `MODEL_PROB_CAP = 0.95` was introduced to mask.

## Goal

Persist each forecast value at the actual lead time it was issued, so EMOS can be trained on `(mu_at_lead_L, sigma_at_lead_L, actual_high)` triples for L ∈ {24h, 12h, 6h, 1h} per station per date.

## Proposed approach

1. **Schema change** — extend `model_forecast_log` to be keyed on `(station, model, date, lead_hours)` instead of `(station, model, date)`. New columns: `lead_hours INTEGER`, `issued_at TEXT` (the wall-clock UTC time the forecast was fetched).
2. **Migration** — rename existing `model_forecast_log` → `model_forecast_log_legacy_v1` (do not drop; keep for audit). Create new table under the original name. Old EMOS code paths read empty new table → fall through to `InsufficientDataError` until enough lead-time rows accumulate. See companion reset issue.
3. **Scheduled captures** — add a separate cron-driven capture worker (or hook into the existing scan) that:
   - At 12Z each day, fetches and logs the forecast for **tomorrow** (lead ≈ 24h to next day's typical high).
   - At 06Z, 12Z, 18Z, 21Z each day, fetches and logs the forecast for **today** (leads ≈ 18h, 12h, 6h, 3h).
   - The exact times can be tuned per local sunrise/sunset; what matters is they're **fixed** so EMOS sees consistent lead-time bins.
[...]

PR #465, opened 2026-06-26
[Epic A1] GRIB2/HRRR ingestion infrastructure + AWS byte-range caching; DEB model registry

### LINKED ISSUE #432: [Epic A1] Set up herbie / GRIB2 extraction infrastructure + AWS caching
Parent: #431

## What

Bootstrap the GRIB2 ingestion layer that future HRRR/NBM/GEFS modules will all consume. Choose between `herbie` (MIT, NOAA-archive-aware) and rolling our own thin wrapper around `cfgrib` + `s3fs`.

## Acceptance criteria

- [ ] Decision recorded: `herbie` vs custom (write a 3–5 line rationale in PR body)
- [ ] Dependency added to `requirements.txt`
- [ ] `src/data/grib_cache.py` (or equivalent) handles: model cycle resolution, AWS S3 byte-range fetch of single GRIB messages by `:TMP:2 m above:` matcher, on-disk cache with TTL
- [ ] Unit test: fetch HRRR TMP_2m at one (lat, lon) for a recent cycle, assert sane temperature value
- [ ] Documented in `docs/OPERATIONS.md`: AWS bucket paths, cache location, eviction policy

## Technical notes

- AWS S3 buckets: `s3://noaa-hrrr-bdp-pds/`, `s3://noaa-nbm-grib2-pds/`. Public, no auth, free egress to internet.
- `herbie` typically handles model-run latency well; fall back to previous cycle if current is delayed.
- Stick to `TMP_2m` and `DPT_2m` fields. Do NOT pull full files (would be hundreds of MB).

### LINKED ISSUE #462: DEB architectural upgrade: model registry, regional applicability, cold-start policy
## What

Refactor `src/model/deb_weighting.py` from a hardcoded 3-tuple (`nws`, `open_meteo`, `gfs`) to a model-registry pattern that supports adding new forecast channels declaratively. **Prerequisite** for Epic A1 (#431) and Epic A2 (#438) — without this, adding HRRR/NBM/ECMWF/ICON/GEFS will overweight correlated channels and degrade ensemble skill instead of improving it.

## Why now

Surfaced during 2026-06-26 follow-up to the forecast-input-upgrade epics. The current DEB implementation has four real limitations the new channels expose:

1. **Hardcoded MODELS tuple** — adding channels requires editing the module; phantom-model guard only handles total absence, not regional applicability
2. **Cold-start defaults** — new channels with no error history default to equal weight, which is too aggressive against well-calibrated incumbents
3. **Multicollinearity unmodeled** — HRRR, NBM, and NWS share upstream NOAA data; DEB treats them as independent and will overweight the correlated cluster
4. **Cadence mismatch** — `_DECAY_RATE=0.05/day` was tuned for twice-daily NWS; HRRR refreshes hourly. Same decay rate applied to hourly data over-discounts older HRRR observations

## Scope (in)

- Model registry: `register_model(name, region, expected_cadence_h, group_id)` API
- Regional applicability: per-station availability mask (HRRR=US-only, ICON=EU-only, etc.)
- Cold-start policy: new channels start at a configurable fraction (e.g. 0.5×) of equal weight until N samples accumulate
- Cadence-aware decay: per-model decay rate scaled to expected cadence
- Group/correlation handling: channels sharing a `group_id` (e.g., `noaa_global`) share a capped joint weight
- Migration: `MODELS = ("nws", "open_meteo", "gfs")` becomes a registry call; existing behavior preserved exactly for the 3 legacy channels
- Unit tests covering all new behaviors

## Scope (out)

- The actual HRRR/NBM/ECMWF/ICON integrations — those land in Epic A1/A2 once this refactor merges
- Full correlation-matrix weighting (overkill for now — group-cap is enough)

## Acceptance criteria

- [ ] `src/model/deb_weighting.py` refactored to registry pattern
- [ ] Backward compat: existing 3-channel behavior preserved for the legacy MODELS; no regression in DEB weight outputs for existing stations on a 30-day replay
- [ ] New registrations testable without touching the module
- [ ] Regional applicability tested for US-only, EU-only, global channels
- [ ] Cold-start policy unit-tested with synthetic histories
- [ ] Cadence-aware decay unit-tested with hourly vs daily samples
- [ ] Group cap unit-tested with 3 correlated + 1 independent channel
- [ ] `model_weights` schema accommodates new model identifiers (already wide enough — confirm)
- [ ] DEB_REPORT.md regenerated from 30-day replay shows the new code path produces ≤ ±0.01 weight delta on legacy stations

## Dependencies

- **Blocks**: #435 (Epic A1 DEB wiring) and #442 (Epic A2 DEB wiring). Land this before either.
- **Coordinates with**: #69 (DEB epic, closed) — read that epic's docs first to avoid breaking documented invariants.

## Notes

- Don't fully rebuild — the current implementation works for 3 channels. The refactor is about *extension points*, not rewriting the math.
- Group-cap example: if HRRR + NBM + NWS share group_id `noaa_us`, their combined weight is capped at e.g. 0.7 (configurable), regardless of individual error performance, so ECMWF/Open-Meteo never get squeezed out by correlated NOAA models.

P0K5:
## Issues addressed

Closes #432
Closes #462

---

### Decision: herbie vs custom cfgrib+s3fs wrapper

**Chose herbie (MIT).** Rationale:
1. herbie already implements model-run latency detection + automatic fallback to the previous cycle — reimplementing that correctly is non-trivial.
2. herbie parses the `.idx` sidecar to identify exact byte ranges per variable, so no full GRIB file is ever downloaded (hundreds of MB avoided per fetch).
3. MIT license is compatible with commercial use; the alternative (custom s3fs) would have required re-solving the same byte-range and IDX-parsing logic from scratch.

### What changed

- **`src/data/grib_cache.py`** (new): model cycle resolution with 6h lookback fallback, byte-range S3 fetch via herbie IDX sidecar (never full GRIB files), on-disk TTL cache with opportunistic eviction, `fetch_hrrr_field()` public API for `TMP_2m` and `DPT_2m`.
- **`src/config.py`**: `GRIB_CACHE_TTL_HOURS` (default 6h) and `GRIB_CACHE_DIR` (default `.grib_cache`) added with env-var-override pattern.
- **`requirements.txt`**: `herbie-data>=2024.3.0` and `cfgrib>=0.9.10` added.
- **`src/tests/test_grib_cache.py`** (new): 20 offline unit tests covering cache key/path helpers, TTL eviction, config env-var overrides, full `fetch_hrrr_field()` mock with 273–318 K sane-range assertion for Denver TMP_2m.
- **`docs/OPERATIONS.md`** (new): AWS bucket paths, cache layout, eviction policy, config params, supported fields, smoke test.

### Supported fields (byte-range only)

| Key | Matcher |
|---|---|
| `TMP_2m` | `:TMP:2 m above ground:` |
| `DPT_2m` | `:DPT:2 m above ground:` |

### How to test (issue #432)

```python
# Smoke test — requires herbie-data and cfgrib installed; uses Denver (HRRR is US-only)
from src.data.grib_cache import fetch_hrrr_field
val = fetch_hrrr_field("TMP_2m", lat=39.73, lon=-104.99)
print(f"TMP_2m Denver: {val} K  ({val - 273.15:.1f} °C)")
assert val is not None and 273.0 <= val <= 318.0
```

Unit tests (offline, no network):
```bash
pytest src/tests/test_grib_cache.py -v   # 20 tests, all pass
```

---

### What changed

Refactors `src/model/deb_weighting.py` from a hardcoded 2-model tuple (`nws`, `open_meteo`) to a declarative model registry pattern that supports adding new forecast channels without touching compute logic.

### Model registry (`register_model` + `_REGISTRY`)
- `register_model(name, region, expected_cadence_h, group_id, cold_start_fraction)` — declarative channel registration
- Pre-registers all 3 legacy channels at module load: `nws` (us, 24h), `open_meteo` (global, 24h), `gfs` (global, 6h)
- `MODELS` tuple and `EQUAL_WEIGHTS` dict remain as backward-compat exports

### `compute_weights` upgrades
- **Regional applicability**: models whose region doesn't match the station are excluded (e.g. `nws` excluded for EU stations)
- **Cadence-aware decay**: per-model rate = `BASE_DECAY_RATE * (cadence_h / 24.0)`; GFS (6h) gets `0.0125` vs `0.05` for 24h models
- **Cold-start policy**: cold models get `cold_start_fraction × (1/N_models)`; full equal-weight fallback only when ALL models are cold
- **Group weight cap**: total weight of any `group_id` capped at `GROUP_WEIGHT_CAP` (default 0.7)

### How to test (issue #462)

```bash
python -m pytest src/tests/test_deb_weighting.py -v   # 33 tests
```

---

## Trade-offs / notes

- `_read_grib_nearest()` is extracted as a standalone function so the full `fetch_hrrr_field()` chain can be unit-tested without the herbie/cfgrib/eccodes stack installed.
- Cache eviction is opportunistic (triggered on each `fetch_hrrr_field()` call) — no background thread needed at current fetch cadence.
- `GRIB_CACHE_TTL_HOURS`/`GRIB_CACHE_DIR` are env-var-only (not in DB-backed CONFIG_DEFAULTS) because they control infrastructure layout rather than trading strategy, consistent with `HTTP_TIMEOUT_SECONDS`, `LOG_DIR`, etc.
- 6 pre-existing test failures in `test_portfolio_api.py` (py_clob_client_v2 import error) are unrelated to this PR.


P0K5:
## Tech Lead PM review — changes required before merge

Good structural work on the registry pattern, regional applicability, cold-start, and group cap. Three issues must be fixed.

---

### [blocker] `refresh_weights` crashes with `TypeError` at runtime

`compute_weights` now returns `dict[str, float]` but `refresh_weights` still does:

```python
weights, rmse = compute_weights(db, station, city)
```

When `DEB_ENABLED=true` this immediately raises `TypeError: cannot unpack non-iterable dict object`.

**Fix** — in `refresh_weights`, change to:
```python
weights = compute_weights(db, station, city)
for model, weight in weights.items():
    db.upsert_model_weight(..., rmse=0.0)
```

---

### [blocker] `get_forecast_log_by_lead` removed — regression from #422

The new `compute_weights` always calls `db.get_forecast_log(station, since_date)`. The old code had:

```python
if hasattr(db, "get_forecast_log_by_lead"):
    log_rows = db.get_forecast_log_by_lead(station, since_date, lead_hours=24)
else:
    log_rows = db.get_forecast_log(station, since_date)
```

Issue #422 specifically migrated DEB to `lead_hours=24` slices so it compares like-for-like forecasts and doesn't mix nowcast snapshots with genuine ahead-of-event forecasts. Reverting this degrades DEB accuracy on the live DB.

**Fix**: Restore the `hasattr` guard in `compute_weights` before the row-building loop.

---

### [medium] `check_weight_quality` return type changed without updating callers

Original contract: `-> list[str]` (violations list). New contract: `-> dict[str, object]`. Existing tests at lines 173–212 of the test file import and call the old interface — they will fail under CI.

**Fix**: Either restore the `list[str]` violations return (simplest), or update all test assertions to use the new dict interface.

---

Please fix these three issues, push to the branch, and re-request review. Do not request review until `pytest` and `flake8` are green.

---

Issue #548, opened 2026-07-01
2026-07-01 [P0] gfs forecast channel is a byte-identical duplicate of open_meteo — fix before DEB activates


P0K5:
## Context (audit 2026-07-01, finding §1.1 — highest severity)

`fetch_gfs_with_spread()` at `src/data/open_meteo.py:162-181` simply returns `fetch_open_meteo_with_spread()` — the docstring admits it is "a semantic distinction (logged under model=\"gfs\") rather than a different API call". Verified in DB: **all 1,140 `gfs` rows in `model_forecast_log` are identical to `open_meteo` rows** (same mu, same sigma, every station, every lead).

Neither channel has a `group_id` in the DEB registry (`src/model/deb_weighting.py:100-101`), so `GROUP_WEIGHT_CAP` does **not** protect against this. Once DEB calibrates, one signal gets 2× weight, and EMOS's ensemble μ double-counts it too.

**Timing is critical**: gfs/open_meteo/nws cross `MIN_SAMPLES=10` at lead=24 within ~1-2 days (log restarted June 25). This must merge before DEB activation.

## Fix options (decide in PR; preferred = a)

a) **Real GFS data**: point `fetch_gfs_with_spread` at Open-Meteo `models=gfs_seamless` only (single-model; document sigma sourcing — e.g. reuse the calibrated-sigma path or leave sigma NULL and let #EMOS-sigma issue handle it).
b) Drop the `gfs` channel from the registry and stop logging it.
c) Interim: keep the duplication but give `gfs` + `open_meteo` a shared `group_id` so the cap binds (acceptable stopgap if (a) slips past DEB activation).

## Acceptance criteria

- [ ] New `gfs` rows in `model_forecast_log` are no longer identical to `open_meteo` (verify on 1 day of fresh data: mu differs for the overwhelming majority of station/lead rows) — or the channel is removed/capped per (b)/(c)
- [ ] DEB registry (`deb_weighting.py`) reflects the decision; weights still sum to 1.0 in all region configs
- [ ] Historical duplicate-era `gfs` rows are excluded from DEB matched-pair training (or purged) so calibration doesn't learn from the duplicated period — decision documented in the PR
- [ ] `src/tests/test_gfs_forecast.py` updated; add a regression test asserting gfs and open_meteo fetchers do not return the same object/values for the same inputs (if (a))
- [ ] CI green

## Dependencies / sequencing

[...]

PR #562, opened 2026-07-01
fix(deb): read DEB_ENABLED from live config instead of env var

### LINKED ISSUE #549: 2026-07-01 [P0] envelope.py reads DEB_ENABLED from env instead of live config — dashboard toggle silently ignored
## Context (audit 2026-07-01, finding §2.3)

`src/model/envelope.py:91`:

```python
elif state.deb_mu_f is not None and os.getenv("DEB_ENABLED", "false").lower() == "true":
```

Everything else reads the DB live-config, and the flag was flipped in the **dashboard** on June 26. If the bot process env doesn't also set `DEB_ENABLED`, `deb_mu_f` is silently ignored whenever `corrected_mu_f` is None — meaning the DEB rollout we think is on may not be.

## Fix

Read the flag via the live-config helper (`get_live_config` / same pattern as other strategy flags), with the existing env var as fallback default only. Do not change the μ precedence order (`corrected_mu_f` > `deb_mu_f` > baseline).

## Acceptance criteria

- [ ] `envelope.py` resolves `DEB_ENABLED` from DB live config, consistent with the other flags wired through `CONFIG_DEFAULTS`/`_CONFIG_META`
- [ ] Unit tests cover: flag on via DB (env unset) → deb_mu_f used; flag off → ignored; corrected_mu_f still takes precedence (`src/tests/test_envelope.py`)
- [ ] Log line at startup (or first envelope evaluation) stating the effective DEB_ENABLED value and its source, so silent mismatches are visible
- [ ] CI green

## Dependencies / sequencing

Must land before/with DEB activation this week. One-file change + tests — good Junior issue.

P0K5:
## Summary

Fixes #549 — `envelope.py` now reads `DEB_ENABLED` from the live database config instead of only checking the environment variable. This ensures that the dashboard toggle is respected during execution.

## Changes

- **src/model/envelope.py**: Modified `true_probability_yes()` to accept an optional `db` parameter. When provided, reads `DEB_ENABLED` from live config via `get_live_config(db)`. Falls back to environment variable when `db` is None (for backward compatibility with tests).
- **src/strategy/scanner.py**: Updated both calls to `true_probability_yes()` to pass the `db` parameter when available.
- **src/tests/test_envelope.py**: Added 5 new tests in `TestDebMuFLiveConfig` class to verify:
  - DB-based reading works when DEB_ENABLED=true
  - DB-based reading works when DEB_ENABLED=false
  - DB value takes precedence over environment variable
  - Environment variable fallback works when db=None
  - corrected_mu_f still takes precedence over deb_mu_f

## Implementation Details

- Added module-level flag `_deb_enabled_logged` to ensure the DEB_ENABLED status and source are logged only once per process startup
- Log line format: `DEB_ENABLED={value} (source: {live config | env var | default})`
- Maintains μ precedence order unchanged: `corrected_mu_f > deb_mu_f > ensemble_forecast`

## Testing

- All 36 tests in `test_envelope.py` pass
- No flake8 style violations
- New tests verify both DB and env var code paths
- Backward compatibility with existing tests maintained

## How to Test

1. Run tests: `pytest src/tests/test_envelope.py -xvs`
2. Verify log output when scanner starts shows DEB_ENABLED status and source
3. Toggle DEB_ENABLED in dashboard and verify envelope probabilities respond accordingly (when deb_mu_f is available and corrected_mu_f is None)

---
_Generated by [Claude Code](https://claude.ai/code/session_01W95oJtjsSvBZSMjRNDucXs)_


P0K5 (commented):
**Review by Tech Lead PM — CHANGES REQUESTED** (posted as comment; the API won't allow a formal request-changes from the same token identity).

The flag semantics are correct (verified `get_live_config` coerces `DEB_ENABLED` to bool via `CONFIG_DEFAULTS`), tests are a good spread, and μ precedence is preserved. One blocker before this can merge:

**[blocker] Per-bracket DB read in the hot loop.** `true_probability_yes()` now calls `get_live_config(db)` on every invocation. The scanner evaluates it per bracket per market per scan cycle — that's hundreds of redundant `SELECT key, value FROM bot_config` reads plus the full coercion loop each cycle. The established pattern is to read live config **once per scan**: `scanner.py:358` already has `_live = get_live_config(db)` in scope in `scan_markets`.

Requested change: don't pass `db` into `true_probability_yes` at all. Change the signature to accept the resolved flag, e.g. `deb_enabled: bool | None = None`:
- `None` → fall back to the env var exactly as before (backward compatibility for existing callers/tests);
- scanner passes `deb_enabled=_live["DEB_ENABLED"]` from the existing line-358 dict at both call sites.

This also removes the DB coupling and the `try/except` fallback from `envelope.py` entirely — the coercion already happened in `get_live_config`.

**[suggestion]** `log.warning(f"...")` / `log.info(f"...")` — the codebase uses lazy `%`-style logging (`log.info("... %s", x)`); match it.

**[suggestion]** `test_env_var_fallback_when_no_db` only asserts `0 <= p <= 1`, which passes whether or not the DEB path fired. Compare against a no-deb baseline like your other tests do.

The `_deb_enabled_logged` once-per-process log is fine as is.

Sequencing note (no action needed from the author): even after approval this PR merges **after** #548, since it activates the DEB path that #548 decontaminates.

---

P0K5:
Review feedback addressed in 91c1b52 — ready for re-review.

**[blocker] Per-bracket DB read** — fixed. `true_probability_yes` no longer touches the DB: the `db` parameter, `get_live_config` import, and try/except are gone. New signature: `true_probability_yes(..., deb_enabled: bool | None = None)`. `None` falls back to the `DEB_ENABLED` env var exactly as before (backward compat); otherwise the passed bool is used. `scan_markets` resolves the flag once per scan from the existing `_live` dict (line 358) — `_deb_enabled = bool(_live.get("DEB_ENABLED", CONFIG_DEFAULTS["DEB_ENABLED"]))` — and passes it at both call sites. When `db is None`, the scanner passes `None` so the env-var fallback still applies. No new `get_live_config` calls added.

**[suggestion] Lazy %-style logging** — done: `log.info("DEB_ENABLED=%s (source: %s)", deb_enabled, deb_source)`. Source is now "caller" vs "env var". The once-per-process guard is unchanged.

**[suggestion] Weak env-fallback test** — strengthened: `test_env_var_fallback_when_deb_enabled_none` now compares against a no-deb baseline and asserts the probabilities differ. Also tightened `test_corrected_mu_f_precedence_over_deb` to assert the result equals a corrected-mu-only baseline (proving deb_mu_f was ignored), and replaced the MockDb tests with direct `deb_enabled=True/False` passing.

**Test results:** test_envelope.py 36/36 pass; full suite 1814 passed, 9 skipped; flake8 clean on all three touched files.

Noted on sequencing after #548 — no merge action from my side.

---

P0K5 (commented):
**Review by Tech Lead PM — APPROVED** (posted as comment; same-token identity prevents a formal approval).

Rework verified at 91c1b52: flag resolved once per scan next to the existing `_live` block (no per-bracket DB reads), `envelope.py` fully decoupled from the DB with the env-var fallback intact, μ precedence unchanged, lazy `%`-logging, and the strengthened tests now assert against baselines. All review items addressed.

CI gate verified on head 91c1b52: Test (pytest) ✅, Lint (flake8) ✅, gitleaks ✅ — all `completed: success`.

**Merge is deliberately deferred until PR #563 (#548 gfs decontamination) merges** — this PR activates the DEB path in the envelope, and the gfs duplicate fix must be in first so the newly-opened valve serves clean weights. I'll merge both in order once #563 clears review.

---

Issue #551, opened 2026-07-01
2026-07-01 [P0] MODEL_PROB_CAP saturation: p_yes clamps at 0.05/0.95 on ~99% of candidates — model probability never reaches trade selection


P0K5:
## Context (audit 2026-07-01, loophole §3.1 — "the big one")

In yesterday's candidates file, 104/159 rows have `p_yes = 0.05` and 54/159 have `p_yes = 0.95` — the `MODEL_PROB_CAP` clamp (`src/config.py:346`, default 0.95, also in `CONFIG_DEFAULTS` at :493; `cap_applied` fired ~169,000 times this week). With sigma fixed at ~2-2.5°F and `MIN_FORECAST_BRACKET_MARGIN_F=2.5`, essentially every candidate saturates before the edge computation sees it.

**Consequence: edge = f(price) only.** Whether the model thinks a bracket is 4.9% or 0.1% likely, the selector sees 5%. DEB weights, intraday corrections, residual corrections, future EMOS — all flattened at the last step. Station selection, not model quality, has been the actual PnL engine.

Related: #420 tracks the *post-EMOS* removal of this cap. This issue is the **now** fix — making model probability influence trades again without loosening protected gates.

## Constraint (non-negotiable)

Lowering the floor (cap → 0.97-0.98) raises computed NO edges and **effectively loosens the NO entry gate**. NO is the protected profitable side. Therefore:

- The **live entry gate keeps consuming the capped p** until a shadow comparison proves a change safe.
- The uncapped p may immediately be used for **ranking and sizing** (which candidate to prefer, how much to stake) — that doesn't loosen any gate.

## Implementation plan

1. Plumb **both** values through the candidate pipeline: `p_yes_capped` (current behaviour, feeds the entry gate) and `p_yes_raw` (uncapped). Log both in the candidates file and trades table.
2. Switch candidate **ranking/sizing** to `p_yes_raw` behind a config flag (`RANK_ON_RAW_PROB`, default false), wired through `CONFIG_DEFAULTS` + `_CONFIG_META` + dashboard (per `feedback_strategy_params_in_dashboard`).
3. Shadow-validate a cap change: run shadow with cap 0.97/0.98 alongside live 0.95 for ≥7 days; produce a comparison report (win rate, avg edge, trade volume, NO-side specifically).
4. Only after the report: separate decision (new issue or comment sign-off here) on changing the live cap.

## Acceptance criteria

[...]

PR #564, opened 2026-07-01
feat: log raw p_yes alongside capped; optional raw-prob ranking flag (stage 1 of #551)

P0K5:
## Part of #551

Stage 1 only, per PM guidance: **dual logging + a default-off ranking flag**. No `MODEL_PROB_CAP` value change, no live entry-gate change. The shadow-window comparison and the actual cap-value decision stay open in #551 for a follow-up.

### What changed

- **`Candidate` dataclass** (`src/strategy/scanner.py`) gains `p_yes_raw`, `ev_yes_raw`, `ev_no_raw` — the pre-clamp probability and its derived edges. Populated on both the high-side scan path (symmetric clamp around L480-490) and the low-side scan path (upper-only clamp around L679). These fields are additive/nullable-default and never read by any entry gate.
- **`RANK_ON_RAW_PROB` config flag** — added to `CONFIG_DEFAULTS` (`src/config.py`) and `_CONFIG_META` (`src/dashboard/api.py`), default `false`, dashboard-wired via the existing generic `/api/config` GET/PATCH endpoints (same pattern as `RESIDUAL_CORRECTION_ENABLED` / `DEB_ENABLED`). When `true`, `scan_markets()` sorts the returned candidate list by the raw-probability-derived edge on each candidate's flagged side (`ev_yes_raw` for YES, `ev_no_raw` for NO), descending. When `false` (default), **no sort is applied at all** — the candidate list order is byte-for-byte the same as today.
- **Persistence**: `p_yes_raw` is now logged to the candidates CSV, and threaded through to the DB:
  - New nullable `p_yes_raw REAL` column on `candidates` and `trades` (migration added to `_migrate()` in `src/data/db.py`, following the existing `ALTER TABLE` pattern).
  - `insert_candidate()`, `insert_trade()`, `upsert_shadow_trade()` all gain an optional `p_yes_raw` kwarg.
  - Live trades: `LiveTrader.place_order()` → `order_executor._execute_live()` now passes `candidate.p_yes_raw` through to the DB insert and into the live-trades JSONL record (and the order-manager reconciliation fallback insert).
  - **Bug fix found along the way**: the legacy `trades` table CHECK-constraint rebuild (the `trades_new` DROP/CREATE block that adds `'shadow'` to the mode CHECK for old DBs) only copied a fixed column list forward and would have silently dropped `p_yes_raw` for any DB that still needed that rebuild. Fixed by adding `p_yes_raw` to the `trades_new` schema and the `INSERT ... SELECT`, coalescing to `NULL` when the source table doesn't have it yet (same pattern already used for `direction`).

### Why

Issue #551: `MODEL_PROB_CAP` saturates ~99% of `p_yes` values to 0.05/0.95, so today only the market price drives edge — model probability (DEB weights, intraday/residual corrections) never reaches trade selection. NO is the protected profitable side, so the live entry gate must keep consuming the capped probability until a shadow comparison proves a cap change is safe. This PR makes the *raw* probability visible (dual logging) and lets it optionally drive *candidate ordering* only — never gates — behind a flag that's off by default.

### How to test

```
python -m pytest src/tests/test_model_prob_cap.py src/tests/test_rank_on_raw_prob.py src/tests/test_p_yes_raw_column.py src/tests/test_scanner.py src/tests/test_db_shadow.py src/tests/test_direction_column.py -q
python -m pytest src/tests/ -q     # full suite
python -m flake8 .
```

New/updated tests:
- `src/tests/test_model_prob_cap.py` — `TestCandidateRawProbability`: `p_yes_raw`/`ev_no_raw` equal the capped values when no clamp fires.
- `src/tests/test_rank_on_raw_prob.py` (new) — raw fields populated & diverge correctly when the clamp fires; flag-off preserves scan order and capped edges exactly (regression); flag-on reorders by raw edge while the candidate set, sides, and capped `edge_cents` are unchanged.
- `src/tests/test_p_yes_raw_column.py` (new) — fresh-install schema, idempotent migration, insert/read round-trips for `candidates`/`trades`/`upsert_shadow_trade`, and a dedicated regression test for the legacy CHECK-constraint rebuild bug described above.

Full suite: **1828 passed, 9 skipped**. `flake8 .`: clean.

### Notes / things worth a second look

- The low-side scan path (`src/strategy/scanner.py` ~L679) only ever applied an **upper** clamp (`min(p_yes, MODEL_PROB_CAP)`), not the symmetric `max(p, 1-cap)` the high-side path uses. I preserved this asymmetry exactly (it's an existing low-side gate behaviour, and low-side entries are shadow-only) rather than changing it, per the "no live-gate changes" constraint — flagging it here in case it's worth its own follow-up issue.
- The ranking formula (`ev_yes_raw`/`ev_no_raw` on the flagged side, descending) was a judgment call since there was no pre-existing candidate sort to mirror (today's execution order is simply scan/market order — no ranking exists yet at all). Happy to adjust the exact tie-break/formula if you have a different preference before this flag is ever flipped on.
- Sizing-on-raw-probability is explicitly out of scope here (per the issue plan) — noting it as a possible stage-1b follow-up for a separate PM decision.

---
_Generated by [Claude Code](https://claude.ai/code/session_01W95oJtjsSvBZSMjRNDucXs)_


P0K5 (commented):
**Review by Tech Lead PM — APPROVED** (as comment; same-token identity).

Verified in review:
- Entry gates read only the capped `p_yes`/`ev_*` on every path (high, NO, low); raw fields are additive and gate-invisible.
- Flag off → no sort applied at all; regression test asserts identical candidate set, edges, and ordering.
- DB migration follows the established ALTER pattern; the legacy CHECK-rebuild coalesce fix is a genuine catch — good find, and it has its own regression test.
- `RANK_ON_RAW_PROB` wired through CONFIG_DEFAULTS + _CONFIG_META per project convention, default off.

Decisions on the author's flagged judgment calls:
1. **Ranking formula signed off**: raw edge on the flagged side, descending, is the right stage-1 semantics. Revisit tie-breaks only if the shadow window shows order-sensitivity.
2. **Low-side upper-only clamp asymmetry**: preserving it was correct under the no-gate-changes constraint. Logging it as a Backlog issue so it isn't lost.
3. **Sizing-on-raw (stage 1b)**: deferred, tracked in #551 which stays open for the shadow window and cap decision.

CI gate verified on head 6fce372: Test (pytest) ✅, Lint (flake8) ✅, gitleaks ✅. Post-review merge commits (master syncs + graphify reset) were mechanical; the graphify-out diff is now empty relative to master, which ends the graph-bot conflict cycle for this PR.

Merging. #551 remains open: next step is the ≥7-day shadow comparison once this deploys.

---

Issue #307, opened 2026-06-16
Per-city warm-bias residual correction + monitoring (Busan/Singapore/Tokyo under-prediction)


P0K5:
## Context

`intraday_corrections.delta_f` (observed − model) reveals a **systematic warm bias** — the model under-predicts the daily high — concentrated at the stations losing money:

| City | n | Mean error | MAE | Max miss |
|---|---|---|---|---|
| **Busan** | 90 | **+4.6°F** | **10.1°F** | +21.1°F |
| Tokyo | 14 | +5.9°F | 8.4°F | +16.0°F |
| **Singapore** | 70 | +2.1°F | 3.7°F | +10.1°F |
| Seoul | 89 | +1.4°F | 2.4°F | +8.7°F |

For a NO book ("high won't reach the bracket"), under-predicting the high is the exact bias that busts NO. Singapore (−€8.46) and Busan (−€4.61) are the two worst P&L stations — this is a correctable bias, not noise. The existing intraday correction is clearly not removing the residual.

## Goal

Apply and monitor a per-city residual bias correction so the forecast mean used for entries reflects the observed warm bias, and gate live NO entries where residual MAE is unsafe.

## Scope

1. Compute a rolling per-city residual (mean signed `delta_f` over a trailing window) and add it as an explicit bias term to the forecast mean feeding `compute_envelope`.
2. Track rolling MAE per city; expose on the dashboard.
3. **Interim safety gate:** new config `MAX_RESIDUAL_MAE_F_FOR_LIVE` — when a station's rolling MAE exceeds it (e.g. Busan at 10°F), suppress new live NO entries (shadow only) until it recovers. This operationalizes the "pause Busan/Singapore" recommendation as a data-driven toggle rather than a hardcoded exclusion.
4. Guardrails: clamp the applied correction to a sane range; require a minimum sample before applying.

## Acceptance criteria

- [ ] Rolling per-city signed-residual bias term computed and applied to the forecast mean (flag-gated, default-on with a conservative clamp)
- [ ] Rolling per-city MAE surfaced on the dashboard
- [ ] `MAX_RESIDUAL_MAE_F_FOR_LIVE` suppresses new live entries above threshold; shadow continues
- [ ] Backtest: apply correction to the settled history and report win-rate / P&L delta at Busan, Singapore, Tokyo, Seoul
- [ ] Unit tests: correction applied with sufficient sample, skipped below minimum, clamp binds, MAE gate suppresses live entries

## Dependencies

[...]

Issue #553, opened 2026-07-01
2026-07-01 [P1] refresh_weights hardcodes rmse=0.0 — log real per-model RMSE so calibrated weights are distinguishable from cold-start


P0K5:
## Context (audit 2026-07-01, finding §2.4)

`refresh_weights()` persists `rmse=0.0` for every model (`src/model/deb_weighting.py:397` — hardcoded, not computed). Every row in `model_weights` is currently uniform (all-cold-start fallback), and with the hardcoded 0.0 you **cannot tell calibrated weights from fallback in the dashboard**. DEB crosses `MIN_SAMPLES=10` for gfs/open_meteo/nws in ~1-2 days — this week that distinction is exactly what we need to watch.

## Fix

`compute_weights()` already assembles per-model matched-pair error lists (`deb_weighting.py:266` area). Return (or expose) per-model RMSE and sample count alongside the weights, and persist the real values in `upsert_model_weight`.

Rules:
- Models below `MIN_SAMPLES` (cold-start): persist `rmse=NULL` (not 0.0) so fallback is unambiguous. If the schema requires a value, add a `sample_count` column instead and treat `sample_count < MIN_SAMPLES` as cold-start.
- Dashboard weights view shows RMSE + sample count per model per city.

## Acceptance criteria

- [ ] `model_weights` rows carry computed RMSE (and sample count) for calibrated models; NULL / explicitly marked for cold-start
- [ ] Dashboard exposes the values (weights table/tooltip)
- [ ] Unit tests: calibrated model → real RMSE persisted; cold-start model → NULL/marked (`src/tests/test_deb_weighting.py`)
- [ ] CI green

## Sequencing

Small, well-scoped — good Junior/Mid issue. Land before DEB activation so the first calibrated rows are already distinguishable.

P0K5:
Continuing work — plan:
1. Update API response model (DebWeightOut) to include sample_count field
2. Update get_city_deb() endpoint to pass real sample_count from database instead of hardcoding 0
3. Add weights display section to station cards in frontend (similar to perf/bias panels)
4. Add test coverage for API endpoint exposing rmse and sample_count
5. Full pytest + flake8 validation

Picking up from previous session which implemented backend logic (real RMSE computation, sample_count column, and unit tests — all verified green). This session completes the dashboard exposure requirement.

---
_Generated by [Claude Code](https://claude.ai/code)_

P0K5:
PR #575 ready for review

[...]

PR #576, opened 2026-07-02
Use live DEB weights as intraday consensus basis

### LINKED ISSUE #572: 2026-07-02 [P1] Intraday correction measures deviation against open_meteo only, but applies it to the DEB blend — basis mismatch goes live when DEB calibrates
## Context (found auditing the residual/intraday correction chain, 2026-07-02)

`src/model/intraday_correction.py` (~line 179) builds its hourly consensus with a hardcoded basis:

```python
consensus = build_consensus(lat, lon, weights={"open_meteo": 1.0})
```

then computes `delta_f = obs − model` against that open_meteo-only curve, and applies the decayed delta **on top of `deb_mu_f`** (line ~191: `corrected_mu_f = state.deb_mu_f + delta_f * decay_factor`). Downstream, `residual_correction.py` (#307) learns its 30-day rolling bias from these same deltas.

While DEB is uniform cold-start this mismatch is second-order. **DEB starts calibrating away from uniform this week** — from then on we'll be correcting a DEB-weighted blend using open_meteo's point-in-time bias. If DEB down-weights open_meteo for a city (which is the whole point), the correction increasingly measures the wrong model's error.

## Fix

Pass the city's current DEB weights into `build_consensus` (they're available via `get_weights(db, city, region)`), falling back to `{"open_meteo": 1.0}` only when weights are unavailable. Consider logging the basis used alongside the stored delta (`intraday_corrections` has a `source` column; a basis tag or the weight snapshot would let the residual layer segment pre/post-fix deltas rather than mixing two regimes in one 30-day window).

## Note

The broader "does residual correction double-correct once EMOS calibrates the mean" question stays with #420 — this issue is only the basis mismatch, which is timely now.

## Acceptance criteria

- [ ] Consensus basis follows live DEB weights with a documented fallback
- [ ] Regime change is distinguishable in stored deltas (tag or weights snapshot) so the rolling window can exclude mixed-basis samples
- [ ] Tests cover: calibrated weights used; fallback path; regime tagging
- [ ] CI green

P0K5:
Closes #572

## What changed

`src/model/intraday_correction.py` built its hourly consensus curve with a hardcoded basis (`build_consensus(lat, lon, weights={"open_meteo": 1.0})`), then measured `delta_f = obs − model` against that open_meteo-only curve — even though the correction is applied **on top of the full DEB blend** (`deb_mu_f`). As DEB starts calibrating away from uniform this week, this becomes a basis mismatch: if DEB down-weights open_meteo for a city, the correction increasingly measures the wrong model's error, and `residual_correction.py` (#307) learns its 30-day rolling bias from these same (now wrong-basis) deltas.

Changes:
1. **`src/model/intraday_correction.py`**
   - New `_get_station_region(city)` — resolves a city's DEB registry region ("us"/"eu"/"global"), mirroring `src.weather.builder._deb_station_region`'s rules. Duplicated locally rather than imported, since `builder.py` imports `compute_correction` from this module and importing back would create a circular import.
   - New `_get_consensus_weights(city, db)` — fetches the city's live DEB weights via `get_weights(db, city, station_region)` and passes them straight into `build_consensus`. Falls back to `{"open_meteo": 1.0}` only when weights are genuinely unavailable (the lookup raises, or the returned dict has no `"open_meteo"` key). DEB's own equal-weight cold-start/disabled dict is treated as a legitimate live basis (not overridden), since it reflects DEB's actual current state rather than an error condition.
   - `compute_correction` now calls `build_consensus(lat, lon, weights=basis_weights)` with the resolved weights instead of the hardcoded dict, and persists the exact weights snapshot used via a new `basis_weights` field on `db.upsert_intraday_correction`.

2. **`src/data/db.py`**
   - Added a `basis_weights TEXT NOT NULL DEFAULT '{"open_meteo": 1.0}'` column to `intraday_corrections` (JSON snapshot of the weights dict used to build that row's consensus basis), so pre/post-fix regimes are distinguishable and the residual layer can later exclude mixed-basis samples from its rolling window rather than blending two different regimes.
   - Wired the column into: the fresh-install `CREATE TABLE`, the legacy pre-station/source-column rebuild migration path (existing historical rows get the legacy open_meteo-only basis), and a simple `ALTER TABLE ADD COLUMN` for already-migrated DBs.
   - `upsert_intraday_correction` gained a `basis_weights: str = '{"open_meteo": 1.0}'` kwarg (defaulted so existing callers are unaffected).

3. **`src/tests/test_intraday_correction.py`** — added coverage for:
   - `_get_station_region` (us/eu/global/unknown-city resolution)
   - `_get_consensus_weights`: calibrated weights passed through unchanged, correct `station_region` forwarded, fallback on `get_weights` exception, fallback when `"open_meteo"` key is missing, and confirmation that DEB's equal-weight cold-start dict is *not* treated as a fallback
   - `compute_correction`: live weights are forwarded to `build_consensus`; the persisted `basis_weights` snapshot matches the calibrated weights used; the persisted snapshot for the fallback regime is distinguishable from a calibrated-regime snapshot (regime tagging)

Out of scope (tracked separately in #420): whether residual correction double-corrects once EMOS calibrates the mean. This PR only fixes the basis mismatch and makes the regime distinguishable in stored data — it does not change `residual_correction.py`'s rolling-window logic itself.

## Why

DEB is calibrating away from uniform weights this week. Without this fix, the intraday correction (and everything downstream that learns from its stored deltas) would silently start measuring deviation against the wrong model once DEB down-weights open_meteo for a city, corrupting the correction signal right as DEB goes live.

## How to test

```
python -m pytest src/tests/test_intraday_correction.py src/tests/test_intraday_correction_edge_cases.py src/tests/test_deb_hourly_consensus.py src/tests/test_deb_weighting.py -q
python -m pytest src/tests -q -n auto --dist loadfile --cov=src --cov-report=term-missing --cov-fail-under=60
python -m flake8 src/
```

All 1868 tests pass (9 skipped, pre-existing), coverage 88.31%, flake8 clean.

Manual DB migration verification (see PR description commands run locally): confirmed `basis_weights` is added correctly for (a) a fresh DB, (b) a legacy pre-station/source-column DB via the rebuild path, and (c) an already-migrated DB via the simple `ALTER TABLE` path — in all three cases historical/default rows get the legacy `{"open_meteo": 1.0}` tag.

---
_Generated by [Claude Code](https://claude.ai/code/session_01DWuwREYtoHXaGvbMPEsSdD)_


P0K5 (commented):
**APPROVED**

Reviewed against #572's acceptance criteria:
- Live DEB weights now feed the consensus basis via `_get_consensus_weights`, with a documented fallback (lookup exception or missing `open_meteo` key) — DEB's own cold-start equal-weight dict is correctly treated as a legitimate basis, not a fallback condition.
- `basis_weights` snapshot column added to `intraday_corrections`, wired through all three migration paths (fresh install, legacy rebuild, ALTER TABLE), defaulting historical rows to the legacy tag — regime is now distinguishable for the residual layer's future rolling-window segmentation.
- Tests cover calibrated weights, both fallback triggers, and confirm the cold-start dict isn't misclassified as fallback.
- Rebased on top of #553/#571/#555/#570 myself (PM) — resolved a `db.py` migration-list conflict with #553's `sample_count` addition (both appended a tuple at the same anchor point; kept both). Re-verified: 1877 passed, flake8 clean, and CI is green on the rebased head.

Merging — this is the last piece before DEB goes non-uniform this week.

---

PR #579, opened 2026-07-02
Make EMOS min_samples configurable; enforce shadow-only guardrail

P0K5:
## Summary

Closes #556

Implements configurable EMOS min_samples thresholds to enable early shadow calibration:
- Shadow fits now run at 35 settled days (vs. hardcoded 60), pulling first CRPS rows to ~late July
- Promotion eligibility remains at 60 days (manual promotion required)
- Hard guardrail enforces ready_for_promotion=0 for all <60-sample fits (shadow-only, avoids overfitting)
- GFS_DATA_VALID_FROM filter added to exclude duplicate-era rows (issue #548 duplicate-era GFS exclusion)

### Config (src/config.py)
- Added `EMOS_MIN_SAMPLES_SHADOW` (default 35) to CONFIG_DEFAULTS
- Added `EMOS_MIN_SAMPLES_PROMOTION` (default 60) to CONFIG_DEFAULTS
- Both are dashboard-visible for live adjustment

### Core Logic (src/model/emos_calibration.py)
- Added `GFS_DATA_VALID_FROM = "2026-07-02"` constant (matches deb_weighting.py logic)
- Updated `fetch_training_data()` to filter out "gfs" rows before GFS_DATA_VALID_FROM
- Updated `save_coefficients()` to accept optional `sample_count` parameter
- Enforced hard guardrail: <60 samples → ready_for_promotion=0 (shadow-only fit)

### Scripts
- **run_emos_shadow.py**: Passes `sample_count=len(training_data)` to save_coefficients
- **auto_retrain_probability_calibration.py**: Passes `sample_count=len(data)` to save_coefficients

### Documentation (docs/OPERATIONS.md)
- Added new section "EMOS Shadow Fit Threshold (Issue #556)"
- Documented two-tier threshold: 35 days shadow, 60 days promotion
- Added timeline: first shadow fit expected ~late July (July 28–31)
- Documented both guardrails: sample_count < 60 → ready_for_promotion=0, GFS duplicate-era filter

### Tests (src/tests/test_emos_calibration.py)
- Added `test_reduced_sample_guardrail()` to verify <60-sample fits always have ready_for_promotion=0

## How to Test

1. **Verify config parameters are live-readable:**
   ```bash
   python - <<'EOF'
   from src.config import CONFIG_DEFAULTS, get_live_config
   from src.data.db import Database
   assert "EMOS_MIN_SAMPLES_SHADOW" in CONFIG_DEFAULTS
[...]

PR #580, opened 2026-07-02
fix(#552): switch Edge tab ensemble distribution to model_weights, add staleness guard

### LINKED ISSUE #552: 2026-07-01 [P1] deb_weight_log stale since June 26 while model_weights updates daily — Edge tab renders 5-day-old weights
## Context (audit 2026-07-01, finding §1.3)

`deb_weight_log` has no rows since **June 26**, while `model_weights` keeps updating daily. The Edge tab's ensemble distribution reads `deb_weight_log` (`src/model/ensemble_distribution.py:85`, falling back to equal weights when no row is found), so the new Edge tab renders 5-day-old weights — which also (pre-195f08c) include `nws` for non-US cities.

Whatever wrote that table stopped. With DEB activation days away, the Edge tab must show *current* weights or it will mislead exactly when it matters.

## Scope

1. Find the `deb_weight_log` writer, determine why it stopped on June 26 (crashed job? removed call site? migration?), and fix it.
2. Decide whether `ensemble_distribution` should read `model_weights` instead (one source of truth) or whether `deb_weight_log` serves a distinct purpose (per-snapshot history) worth keeping — document the decision.
3. Purge or ignore stale pre-195f08c rows containing `nws` for non-US cities so the fallback never serves them.
4. Add a staleness guard: if the freshest weight row for a city is older than N days, log a warning and fall back to equal weights explicitly (so silent stalls surface in logs/dashboard).

## Acceptance criteria

- [ ] Root cause of the June 26 stall documented in an issue comment
- [ ] Edge tab's ensemble distribution reflects weights no older than 1 day (or explicit equal-weight fallback with a visible warning)
- [ ] Staleness warning covered by a unit test (`src/tests/test_ensemble_distribution.py`)
- [ ] CI green

P0K5:
Closes #552

## Root cause

`deb_weight_log`'s only writer was the once-per-day logging side effect inside `deb_weighting.get_weights()`. It was **silently deleted** in commit `94a2ece` (2026-06-26 17:29 UTC, PR #465 "[Epic A1] … DEB model registry"), as part of the `get_weights()`/`refresh_weights()` refactor to the declarative model registry. The commit's own test rewrite (`TestDebWeightNotLoggedTwiceSameDay`) was changed to check `get_weights()` idempotency instead of `log_deb_weights.call_count`, so nothing caught that the actual write had disappeared. `model_weights` kept updating daily because `refresh_weights()` — the separate write path used by the live trading loop (`src/weather/builder.py`) — was untouched by that refactor.

Full diagnosis posted on the issue: https://github.com/P0K5/MeteoEdge/issues/552#issuecomment-4864971178

## What changed

- **`src/model/ensemble_distribution.py`** — `_deb_weights_for_city()` now reads `model_weights` (via `db.get_model_weights`) instead of `deb_weight_log`, restricted to the single freshest date per city so weights from different refresh cycles never blend. This is the same table `deb_weighting.get_weights()` reads for live trading decisions, so the Edge tab now shows literally the weights used by the trading logic — one source of truth instead of two write paths that can (and did) silently diverge.
- **Staleness guard** — if the freshest `model_weights` row for a city is older than `EDGE_DEB_WEIGHT_STALENESS_DAYS` (default 1 day, env-overridable), `ensemble_distribution` now falls back to equal weights *explicitly*, logging a `log.warning` that names the city and the stale date. A silent stall like this one will now show up in logs immediately instead of quietly serving 5-day-old data.
- **`src/data/db.py`** — one-time idempotent startup migration purges `deb_weight_log` rows logged before `195f08c` (2026-06-30, the per-region DEB routing fix). Every row currently in the table predates that fix and may attribute `nws` weight to non-US cities; `get_emos_shadow_city_status()` (a separate EMOS shadow-monitoring endpoint) still surfaces the latest `deb_weight_log` row, so this ensures its fallback never serves the contaminated backlog either.
- **`src/tests/test_ensemble_distribution.py`** — updated the DEB-weighted-mean tests to write `model_weights` instead of `deb_weight_log`, and added: a "1-day-old row is not stale" boundary test, a "5-day-old row triggers warning + equal-weight fallback" test (mirrors the real incident), and a "fresh row with no overlap with the active stack still falls back to equal weights" test.

### Design decision (documented on the issue)

I chose to **switch the read path to `model_weights`** rather than resurrecting the `deb_weight_log` write call. Restoring the old writer would keep two redundant persistence paths for the same data — the exact setup that broke silently once already. `model_weights` is already correctly region-routed (post-`195f08c`) and gained real per-model RMSE + `sample_count` in #553. `deb_weight_log` / `log_deb_weights()` / `get_latest_deb_weights()` are left in place (still covered by their existing DB-level unit test) since `get_emos_shadow_city_status()` still reads the table for shadow monitoring — that's out of scope for this issue, but the purge protects it too.

## How to test

```bash
python -m pytest src/tests/test_ensemble_distribution.py -q
python -m pytest src/tests -q -n auto   # full suite
flake8 src
```

Locally: 1927 passed, 9 skipped (full suite); flake8 clean.

To see the guard fire manually: seed `model_weights` for a city with a `date` more than 1 day in the past (`db.upsert_model_weight(city=..., model=..., date=<old date>, weight=..., rmse=0.0, sample_count=20)`), then call `get_ensemble_distribution(station, date, db)` — `ensemble_mean` should reflect equal weights and a `WARNING` naming the city and stale date should be logged.

---
_Generated by [Claude Code](https://claude.ai/code/session_01DWuwREYtoHXaGvbMPEsSdD)_


P0K5 (commented):
**APPROVED**

Reviewed against #552's acceptance criteria:
- Excellent root-cause diagnosis — traced the silent writer removal to a specific commit (94a2ece, PR #465's DEB registry refactor) and identified exactly how it slipped past review (the test was rewritten to check idempotency instead of write call-count). Posted as an issue comment before the fix, as required.
- Design decision to switch the read path to `model_weights` (single source of truth, already correctly region-routed post-195f08c) rather than resurrecting a second writer is the right call — restoring the old writer would just recreate the two-write-paths setup that broke silently once already.
- Staleness guard is real: falls back to equal weights with a named warning when the freshest row exceeds `EDGE_DEB_WEIGHT_STALENESS_DAYS`, with boundary tests on both the 1-day (not stale) and 5-day (stale, mirrors the actual incident) sides.
- Purge migration for pre-195f08c contaminated `deb_weight_log` rows is idempotent and correctly scoped — protects the still-live `get_emos_shadow_city_status()` fallback without deleting a table still in use elsewhere.
- No file overlap with #579 (also in this batch) — clean merge.
- CI green: Test/Lint/gitleaks all passing.

Merging.

---

PR #577, opened 2026-07-02
Wire low-side shadow weather builder into poll_once (fixes #554)

P0K5:
Closes #554

## What changed

Root cause (full diagnosis posted as a comment on #554 before this PR): the low-side scanner block (`scanner.py` ~660-740, Epic C, shadow-only) was fully implemented and unit-tested in isolation — `is_lowest_temp_market`, the low-side probability model (`true_probability_low_in_bracket` in `src/model/envelope_low.py`), and `direction='low'` persistence in `src/data/db.py` were all correct. But `scan_markets()`'s low-side block only runs `if weather_low and prob_low_fn is not None:`, and the production polling loop (`poll_once()` in `src/scripts/run.py`) never passed either argument — it called `scan_markets(weather, markets, db=db, orderbooks=shared_orderbooks)` with no `weather_low`/`prob_low_fn` at all. On top of that, **no function anywhere in the codebase ever built a `WeatherStateLow`** from live data — `src/weather/builder.py` only had high-side builders. The entire low-side block was unreachable dead code in production, which is why all 600 logged trades are `direction='high'`.

This PR closes the gap:

1. **`src/data/metar.py`** — added `low_window_bounds()` and `compute_daily_low_window()`, the low-side counterpart of the existing `sunset_local()`/`compute_daily_high()` pair. Computes the overnight sunset→sunrise observation window (per epic #452 notes — lows settle ~12:00 UTC, window logic differs from highs) and the running minimum temperature observed within it.
2. **`src/weather/builder.py`** — added `build_weather_low_for_scanning()` (+ `_build_one_station_low()` helper), the low-side counterpart of `build_weather_for_scanning()`. Builds a `WeatherStateLow` per station for every station with a "lowest temperature in ..." Polymarket mapping (`POLYMARKET_CITY_TO_STATION_LOW`). `forecast_low_f`/`secondary_forecast_low_f` are deliberately left `None` for v1 — `true_probability_low_in_bracket()` already falls back to the envelope midpoint when no forecast is available, so this is a safe minimal v1. Wiring a real NWS/Open-Meteo low-temperature forecast source is left to a follow-up issue (not required for shadow candidates to start flowing).
[...]

Issue #582, opened 2026-07-02
Low-side weather builder duplicates METAR HTTP calls with the high-side builder


P0K5:
## Context

Found during PR review of #554 (#577), logged as non-blocking at the time but never filed — filing now.

`build_weather_low_for_scanning()` (added in #577, `src/weather/builder.py`) fetches METAR observations independently per poll, with no caching shared with the existing high-side `build_weather_for_scanning()`. Both builders run every poll (`src/scripts/run.py::poll_once()`) and both need the same underlying METAR data per station — currently that means ~2x the METAR HTTP calls per poll cycle across all 30 stations.

## Why it matters

Not correctness-affecting today (#577's tests pass, low-side shadow trades are recording correctly), but it's wasted request volume against `aviationweather.gov` per poll, and will matter more as poll frequency or station count grows.

## Suggested fix

Share the fetched METAR observation set between `build_weather_for_scanning()` and `build_weather_low_for_scanning()` in a single poll — either have one call the other's underlying fetch, or hoist the shared fetch up into `poll_once()` and pass it into both builders.

## Labels
backend, Simple

P0K5:
Starting work — plan:
1. Add a per-poll `metars_cache: dict[str, list] | None` parameter threaded through `_build_one_station`/`build_weather_for_scanning`/`_build_weather` and `_build_one_station_low`/`build_weather_low_for_scanning` in `src/weather/builder.py` — when provided, a station's METAR fetch is served from the cache after the first fetch instead of re-hitting `aviationweather.gov`.
2. Hoist a single `metars_cache = {}` dict into `poll_once()` (`src/scripts/run.py`) and pass it into both the high-side and low-side builder calls for that poll.
3. Add a caller-level test on `poll_once()` proving `fetch_all_metars_today` is called only once per station and that both builders operate on the identical fetched METAR data for a station present in both the active-hours set and the low-market mapping.
4. Run full pytest + flake8, then open PR.

Note: this branch builds on top of `fix/583-low-side-forecast-low` (PR #593, not yet merged) since both issues touch the same files. Will rebase onto `master` once #583 merges, per Tech Lead PM's sequencing instructions.

---
_Generated by [Claude Code](https://claude.ai/code)_

P0K5:
[...]

PR #593, opened 2026-07-02
fix(#583): wire NWS forecast into low-side forecast_low_f

### LINKED ISSUE #583: forecast_low_f left None in low-side shadow candidates (v1 gap from #554)
## Context

Found during PR review of #554 (#577), logged as non-blocking at the time but never filed — filing now.

`build_weather_low_for_scanning()` (added in #577) builds `WeatherStateLow` without populating `forecast_low_f` — it's left `None` for all low-side shadow candidates. The probability model (`true_probability_low_in_bracket`) already handles the `None` case correctly (falls back to observation-only signal), so this isn't a correctness bug, but it means low-side shadow pricing never benefits from forecast information the way the high-side path does.

## Why it matters

Once low-side shadow trades accumulate enough history to evaluate promotion (the same ≥20-30 trade bar used elsewhere in this project for shadow→live decisions), the missing forecast signal could be masking or distorting the low-side edge estimate compared to what live promotion would actually see.

## Suggested fix

Wire a low-side forecast source into `build_weather_low_for_scanning()` — mirror however the high-side builder sources `forecast_high_f` (check `build_weather_for_scanning()` / `build_weather_for_pricing()` for the pattern) and adapt for the low/overnight-window case.

## Labels
backend, Simple

P0K5:
Closes #583

## What changed

`build_weather_low_for_scanning()` (added in #577) built `WeatherStateLow` with `forecast_low_f` always left `None`. This PR wires a real forecast signal into it:

- **`src/data/nws.py`**: added `fetch_nws_forecast_low(lat, lon)` — mirrors the existing `fetch_nws_forecast_high()` exactly (same `/points` permanent cache, same 30-min-TTL forecast cache, same forecast-period window), but takes the **minimum** instead of the maximum over the periods.
- **`src/weather/builder.py`**: `_build_one_station_low()` now calls `fetch_nws_forecast_low(lat, lon)` and passes the result into `WeatherStateLow.forecast_low_f`, instead of hardcoding `None`. `secondary_forecast_low_f` is deliberately left `None` for v1 (no low-side equivalent of the Open-Meteo secondary source exists yet — same scope as the issue, which asks for "a low-side forecast source," singular). Updated the function's docstring to reflect this.
- If NWS is unavailable (non-US station, outage), `fetch_nws_forecast_low` returns `None` and `forecast_low_f` stays `None` — `true_probability_low_in_bracket` / `ensemble_forecast` already handle that fallback (unchanged behavior, still not a correctness bug either way).

## Why

Low-side shadow candidates never benefited from forecast information the way high-side candidates do. This isn't a correctness bug today (the prob model handles `None`), but once enough low-side shadow trades accumulate for a promotion decision, the missing forecast signal could distort the low-side edge estimate versus what live promotion would see.

## How to test

- `python3 -m pytest src/tests/test_nws.py src/tests/test_weather_low_builder.py -q` — new/updated unit tests, all calling the real production functions (`fetch_nws_forecast_low`, `build_weather_low_for_scanning`), no re-implemented logic.
  - `test_nws.py` (new): min-over-window, 18-period cutoff (mirrors the existing high-side test asymmetry), non-Fahrenheit filtering, `None` fallbacks (no URL / fetch failure / malformed payload), and a sanity check that low ≤ high for the same payload.
  - `test_weather_low_builder.py`: updated `_patch_low_build_deps()` to mock `fetch_nws_forecast_low`; updated the existing "builds state" assertion (`forecast_low_f` is now populated, not `None`); added two new tests — forecast populated from NWS via the real builder call path, and fallback to `None` preserved when NWS is unavailable.
- Full suite: `python3 -m pytest -q` → **1952 passed, 9 skipped**.
- `python3 -m flake8 .` → clean, no output.

## Trade-offs / notes

- Scope kept to the NWS source only (matches "Simple" label and the issue's singular "a low-side forecast source" framing) — did not add a secondary/Open-Meteo low-side source or DEB weighting, which would be a larger follow-up if wanted.
- Branch based directly on `origin/master` (869e0a6); no `graphify-out/` changes included.

Requesting review from the Tech Lead PM.

---
_Generated by [Claude Code](https://claude.ai/code/session_018pBBubYS8JBySsV1CjhatM)_


P0K5:
## Tech Lead PM review — APPROVED

Merge gate satisfied: all checks green on HEAD `ab8f9756` (pytest, flake8, gitleaks), backend-only (no Designer review needed).

- `fetch_nws_forecast_low` mirrors `fetch_nws_forecast_high` exactly (same `/points` + 30-min caches, same 18-period **hourly** window) and takes `min` — a sound overnight-low proxy, consistent with the deployed high-side sourcing.
- Wiring verified by a real-function test: `mock_fetch_low.assert_called_once_with(lat, lon)` plus the `None`-fallback case; no re-implemented logic.
- Scope correctly kept to the single NWS source (secondary/Open-Meteo low left as documented v1 follow-up), matching the Simple label.

Merging now. Note for the author of #597 (same agent): once this lands, rebase `fix/582-shared-metar-fetch` onto master so its diff reduces to the #582-only changes.

---

PR #595, opened 2026-07-02
fix(#586): exclude mixed-basis deltas from residual correction window

P0K5:
Closes #586

## What changed

`src/model/residual_correction.py`'s 30-day rolling mean over `intraday_corrections.delta_f` was mixing rows recorded under the legacy hardcoded open_meteo-only consensus basis with rows recorded under the live DEB-weighted basis introduced by #576 (#572). The `basis_weights` column PR #576 added recorded the regime per delta, but the exclusion was never applied to the rolling window query.

- Added `DEB_BASIS_DEPLOY_DATE` (env-overridable, defaults to `2026-07-02`, the #576 deploy date) to `src/model/residual_correction.py`. Documented why we gate on **date** rather than comparing `basis_weights` content: the column's `DEFAULT` value (`'{"open_meteo": 1.0}'`) is textually identical to the legacy hardcoded basis, so a genuinely-legacy row and a post-deploy row where DEB's live weights happen to equal the fallback are indistinguishable by content alone.
- Added an optional `min_date` param to `Database.get_trailing_deltas` (`src/data/db.py`) that raises — never lowers — the effective lower bound of the trailing window.
- Threaded a `deploy_date` param through `_query_trailing_deltas`, `compute_residual_stats`, `compute_residual_stats_per_pair`, and `apply_residual_correction`, so **both** the pair-scoped query and the city-wide fallback query exclude pre-deploy rows (city-wide fallback semantics preserved).
- **Design decision** (per the issue): if excluding mixed-basis rows drops the sample below `RESIDUAL_MIN_SAMPLES`, we prefer **no correction** over a mixed-basis correction. This required no new branching — the existing under-sample guard already returns `None` in that case, since filtering now happens at the query level before the count check. Documented inline as a `CRITICAL DESIGN DECISION` comment in `compute_residual_stats`.
- Extended the existing "applying bias correction" log line to add `basis=deb-since:<deploy_date>` alongside the existing sample count (`n=%d`).

## Why

Per the issue: transitional blur in per-city bias corrections until pre-fix rows age out of the 30-day window (~early August). The corrections applied during exactly the weeks DEB first calibrates were noisier than necessary because the rolling mean blended two different bias regimes.

## How to test

```
python3 -m pytest src/tests/test_residual_correction.py src/tests/test_db.py -q
python3 -m pytest -q
python3 -m flake8 .
```

[...]

PR #597, opened 2026-07-02
fix(#582): share METAR fetch between high-side and low-side builders

### LINKED ISSUE #582: Low-side weather builder duplicates METAR HTTP calls with the high-side builder

P0K5:
Closes #582

**Base note:** this branch is built on top of `fix/583-low-side-forecast-low` (PR #593, not yet merged) since both issues touch the same files (`src/weather/builder.py`, `src/scripts/run.py`). It also includes a merge of `origin/master` (as of `3be11a5`, PR #595) to stay current. The diff will shrink to just the #582-specific changes once #593 merges and this branch is rebased onto `master`.

## What changed

`build_weather_for_scanning()` and `build_weather_low_for_scanning()` (`src/scripts/run.py::poll_once()`) each independently called `fetch_all_metars_today(station)` every poll. For any station present in both the active-hours-gated high-side set and the low-market `POLYMARKET_CITY_TO_STATION_LOW` mapping, that meant ~2x METAR HTTP calls to `aviationweather.gov` per poll for the same station.

- **`src/weather/builder.py`**: added `_get_metars_for_station(station, metars_cache)` — a small helper that fetches once per station and memoizes into `metars_cache` when one is provided (falls back to a plain fetch, unchanged behavior, when `metars_cache=None`). Threaded an optional `metars_cache` parameter through `_build_one_station`, `build_weather_for_scanning`, `_build_weather`, `_build_one_station_low`, and `build_weather_low_for_scanning`.
- **`src/scripts/run.py`**: `poll_once()` now creates a single `_metars_cache: dict = {}` per poll and passes it into both `_build_weather(...)` and `build_weather_low_for_scanning(...)`, so a station fetched by one builder is reused by the other instead of triggering a second HTTP call.
- `build_weather_for_pricing()` is untouched — it's already scoped to 1-3 open-position stations and isn't part of the duplicated hot-loop fetch this issue targets.
- No live config is read per-call in the hot loop; `metars_cache` is a plain in-memory dict scoped to a single poll, created fresh each `poll_once()` call.

## Why

Wasted request volume against `aviationweather.gov` — not correctness-affecting (per the issue), but avoidable duplication that will matter more as poll frequency or station count grows.

## How to test

- **Caller-level test (mandatory per fetch-shape change):** `src/tests/test_run_wiring.py::TestPollOnceSharesMetarFetch::test_metar_fetched_once_and_shared_across_both_builders` drives the real `poll_once()` with a single controlled station (KORD/Chicago, present in both the active-hours set and the low-market mapping) and asserts:
  - `fetch_all_metars_today` is called **exactly once** for KORD across the whole poll (`fetch_calls == ["KORD"]`).
  - The high-side (`compute_daily_high`) and low-side (`compute_daily_low_window`) builders both receive the **identical** fetched METAR list object (`is` check, not just equal-valued) — proving the fetch is genuinely shared, not coincidentally deduplicated.
- Existing `test_weather_low_builder.py` and `test_run_wiring.py` tests updated where their mocked function signatures needed the new `metars_cache` kwarg (backward compatible — defaults to `None`, unchanged behavior for direct/test callers that don't opt in).
- Full suite: `python3 -m pytest -q` → **1958 passed, 9 skipped**.
- `python3 -m flake8 .` → clean, no output.

## Trade-offs / notes

- Chose "hoist the shared fetch into `poll_once()` and pass it into both builders" (per the issue's suggested fix) over having one builder call the other's fetch directly, since the two builders have different station-selection logic (active-hours gate vs. low-market mapping) and a poll-scoped cache dict cleanly decouples "who decides which stations to fetch" from "don't fetch the same station twice."
- `metars_cache` defaults to `None` everywhere so all existing call sites (tests, `build_weather_for_pricing`) are unaffected unless they opt in.

Requesting review from the Tech Lead PM.

---
_Generated by [Claude Code](https://claude.ai/code/session_018pBBubYS8JBySsV1CjhatM)_


P0K5:
## Tech Lead PM review — APPROVED

Merge gate satisfied: all checks green on HEAD `13a479a` (pytest, flake8, gitleaks), backend-only. With #593 now merged, the effective delta here is the #582 METAR-sharing change.

- Poll-scoped `_metars_cache` created fresh in `poll_once()` and passed into both builders — correct lifetime, no per-call config/DB read in the hot loop.
- `_get_metars_for_station` is a clean memoize; `metars_cache=None` preserves prior fetch-every-call behavior for non-opt-in callers (`build_weather_for_pricing`, tests).
- The fetch-shape change carries the required **caller-level** test: real `poll_once()`, `fetch_all_metars_today` called exactly once for KORD, and both builders consume the same list object (`is` check). Exactly the standard that would have caught this class of bug.

Merging now.

---

Issue #609, opened 2026-07-03
2026-07-03 [P0] Live held-to-expiry settlement broken since ~June 16: settle rewrites frozen live_trades.jsonl instead of the DB — closed wins missing from dashboard


P0K5:
## User-visible symptom (2026-07-03)

Two live NO trades won and closed today (WMKK 91.4–93.2°F id 1163+, ZGGG 96.8–98.6°F id 1189) but do not appear in the dashboard's closed section, and their PnL is nowhere. In the DB they sit at `outcome='filled'`, `pnl=NULL`, `settled_at=NULL`.

## Root cause

`settle_live_trades()` (src/scripts/settle.py:254-351) settles live trades by **rewriting `logs/live_trades.jsonl` in place**. That file has been frozen at **2026-06-16** since log rotation moved writes to dated files (`live_trades.YYYY-MM-DD.jsonl`). Every nightly run since then logs `[settle] no live trades to update for <date>` while unsettled rows accumulate in the DB.

The shadow path does this correctly — `settle_shadow_trades()` reads/writes the `trades` table via `db.get_unsettled_shadow_trades()` / `db.update_trade_by_id()`.

## Blast radius

- **Every `outcome='filled'` live trade since 2026-06-20 is unsettled — 42+ rows** (8 on 06-20, 15 on 06-30, 7 KATL on 07-02, 4 on 07-03, …). Held-to-expiry results (full wins at ~+€1 AND full losses at −€5) are entirely absent from live PnL.
- All reported live PnL since ~June 20 includes only `sold`/`timeout` exits → the "+€42 all-time live" figure is not trustworthy in either direction.
- Dashboard closed-positions panel and daily PnL are blind to these trades (portfolio enrichment also reads live_trades.jsonl, src/dashboard/api.py:635-745 — the rotated-file reader helps display but settlement never wrote outcomes at all).
- Likely shares a root cause with the duplicate-entry stacking bug (order reconcile logs `[reconcile] wallet token … has no JSONL history`), filed separately.

## Fix

1. Rewrite `settle_live_trades()` to settle from the **DB trades table**, mirroring the shadow path: select live `filled` rows with `settled_at IS NULL` for the target date, resolve bracket vs daily high, write `pnl`/`settled_at` via `update_trade_by_id`. Keep JSONL writing as optional enrichment only (via the rotated-file reader), never as the source of truth.
[...]

PR #712, opened 2026-07-12
Dashboard closed-positions panel: read settled trades from the DB, not live_trades.jsonl

### LINKED ISSUE #617: Dashboard closed-positions panel should read from the DB, not live_trades.jsonl
## Context

Found during review of #609. Settlement is now DB-first, but the dashboard's closed-positions panel (`src/dashboard/api.py`, `_settled_jsonl_positions` area ~635-745) still renders from `live_trades.jsonl`. #609 bridges this with a best-effort `_enrich_jsonl_with_settlements()` write-back into the rotated JSONL files — which works, but keeps JSONL as a load-bearing display dependency and duplicates state.

Migrating the closed panel to read settled rows from the `trades` table (`mode='live' AND settled_at IS NOT NULL`) would remove the write-back and make the DB the single source of truth end to end.

## Acceptance criteria

- [ ] Closed-positions panel and daily PnL read from the trades table
- [ ] `_enrich_jsonl_with_settlements()` removed (or reduced to a no-op shim) once no consumer needs the JSONL pnl fields
- [ ] Rendering parity verified against the current JSONL-driven output
- [ ] CI green

Refs #609. Frontend-adjacent: if any UI changes result, Designer review is required.

P0K5:
## Linked issue
Closes #617 (refs #609)

## What changed

**Read-path-only migration.** No changes to trade execution, settlement writing, entry gates, `MODEL_PROB_CAP`, or risk logic.

- Added `Database.get_settled_live_trades()` (`src/data/db.py`): `SELECT * FROM trades WHERE mode='live' AND outcome='filled' AND settled_at IS NOT NULL`.
- Added `_db_settled_positions()` (`src/dashboard/api.py`), which converts those rows into `ClosedPositionOut` — entry price from `actual_price`, exit price 100¢/0¢ for win/loss, shares from `size_eur`/`capital_before` over entry price, `closed_at` from `end_date` (fallback `ts`). This replaces `_settled_jsonl_positions()` as the source feeding `_positions_from_wallet()`'s closed_positions list.
- Removed `settle.py`'s `_enrich_jsonl_with_settlements()` (the #609 JSONL write-back) and all references to it — it had no other callers, and `_db_settled_positions()` no longer needs the patched-back JSONL fields.
- **Scope decision:** only the held-to-expiry (`outcome='filled'`) branch moved to the DB. Early exits (`outcome='sold'`, stop-loss/take-profit/manual) stay sourced from `live_trades.jsonl` via the existing `_stopped_positions()` / `_cached_live_trades()` path — those JSONL records are written directly by `order_manager` at sell time, independent of `settle.py`, so they were never broken by the #609/#617 write-back issue and still carry a usable `token_id` for the dashboard's per-position chart. Migrating them too would drop that chart link with no corresponding bug to fix, so I left them as-is. Flagging this scope call explicitly for review — happy to extend to `sold` rows in a follow-up if you'd rather have zero JSONL dependency in this panel.
- **Wallet-dedup fix:** since the `open_positions` row (and its `token_id`) is deleted by `settle.py` at settlement time, DB-sourced closed positions can't carry a `token_id`. The existing `_positions_from_wallet()` logic used `token_id` to skip Polymarket wallet rows that are still `redeemable` for a market we've already settled (avoiding double-counting via the wallet API's unreliable `curPrice`-based win/loss guess — see existing comment in that function). I extended this dedup to also match on market condition ID (`trades.ticker` for real markets vs. the wallet API's `conditionId` field, confirmed present on `/positions` responses per Polymarket's docs) so settled-but-unredeemed positions aren't double-listed.
- `_today_pnl()` / `/status` daily PnL: already DB-first via `_dashboard_load_trades()` (falls back to JSONL only if the DB is unavailable) — no change needed here, confirmed by reading the code.
- `_settled_jsonl_positions()` left in place (not deleted) since it's still directly unit-tested and technically valid if ever pointed at historical/legacy JSONL; just no longer wired into the live render path. Docstring updated to say so.

## Frontend-adjacent — Designer review required

One visible change: **DB-sourced closed (held-to-expiry) positions no longer show the original Polymarket market question text** (`question=""`) or a per-position price chart (`token_id=""`), because neither survives past settlement in the current schema (the `open_positions` row that held them is deleted at settle time). The frontend already has fallbacks for both — `buildQuestion(station, side, bracket_low, bracket_high)` synthesizes a question string, and a plain non-expandable row renders when `token_id` is empty — so nothing crashes or renders blank, but the *wording* of the market description differs from the literal Polymarket question, and the chart disappears for these entries specifically (open positions and early-exit/sold closed positions are unaffected — they keep the real question text and chart). Asking Designer to confirm the synthesized-question wording and no-chart fallback are acceptable for held-to-expiry closed positions, or advise if we should invest in fetching the real question via Gamma (I deliberately avoided adding per-row synchronous Gamma API calls to the `/api/portfolio` request path for this).

## How to test
- `pytest src/tests/test_dashboard.py src/tests/test_settle_live.py src/tests/test_dashboard_jsonl_cache.py -v` — includes new `TestDbSettledPositionsParity` (6 tests) verifying the DB-driven output reproduces the same `exit_reason`/`pnl`/`shares`/`entry_price`/`exit_price` shape as the pre-existing JSONL-driven fixtures in `TestClosedPositionsExitReason`, plus exclusion of `sold`/`shadow`/`paper` rows and graceful `_db is None` handling.
- `pytest src/tests/test_settle_live.py -v` — replaced `TestJsonlEnrichment` (tested the removed write-back) with `TestDbSettledLiveTrades`, asserting settled rows are selectable via the new `get_settled_live_trades()` query.
- Full suite: `pytest src/tests/ -q` — 2293 passed, 9 skipped (pre-existing skips), no failures.
- Lint: confirmed via `ruff check` that the changed files have the same pre-existing warning count before/after this diff (no new lint issues introduced).

## How to deploy
- [x] **Code only** — `git pull` + restart these services: `meteoedge-dashboard` (and `meteoedge-settle.timer`/settle process, since `settle.py` changed — no behavior change, just removed the dead JSONL write-back)
- [x] **Post-deploy validation** — Load the dashboard's closed-positions panel; confirm held-to-expiry wins/losses since deploy show up with correct PnL and no duplicate entries for any market still sitting unredeemed in the wallet. Watch settle logs for any unexpected errors (none expected — settlement logic itself is unchanged).

## Checklist
- [ ] Project board status updated (In review) — **not available in this environment**: no `gh` CLI, no raw GraphQL tool, and `list_issue_fields` returns empty. Noting intended status ("In review") via issue comment instead; Tech Lead PM has the project-board GraphQL access to apply the transition.
- [x] Docs updated if this PR adds/changes env vars, DB schema, API endpoints, or run modes — no schema/API changes; `get_settled_live_trades()` is a new read-only query method, not a schema change.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Xqtthhd4h3XixSs6UvjQuM)_


P0K5:
## Designer review — UX impact of #712 (closed-positions DB migration)

**Overall: APPROVE for UX.** Verified against `src/dashboard/api.py::_db_settled_positions()` and the frontend fallback paths in `src/dashboard/static/index.html` — the PR description matches what's actually in the diff.

**What I verified:**
- `_db_settled_positions()` sets `question=""` and `token_id=""` for every held-to-expiry closed row (lines ~172482-172494 of the diff). Confirmed only `outcome='filled'` rows go through this path; `outcome='sold'` rows are untouched and still carry real `question`/`token_id` from `live_trades.jsonl` via `_stopped_positions()`.
- Frontend fallback for empty `question`: `buildQuestion(station, side, bracket_low, bracket_high)` at `index.html:1034` → `"Will daily high at ${code} ${dir}be between ${bracket_low}°F–${bracket_high}°F?"`. This is **not new for this PR** — it's the same fallback already used for open positions with missing `question` (line 1099); the PR just makes it the *guaranteed* path for a whole category (held-to-expiry closed) instead of a rare edge case.
- Frontend fallback for empty `token_id`: `_renderClosedPage()` at `index.html:1348-1374` — `hasChart` gates the expandable row/chevron/chart-section; when false it renders a plain `<div class="closed-row">` with no crash, no visual break, no blank space.

**Judgment on the trade-off:**

[suggestion] `buildQuestion()` uses the raw station code (e.g. `KORD`) rather than the friendly name already available via `stationMeta(code).name` (e.g. "Chicago O'Hare"). This was a minor gap before (rare edge case), but since it's about to become the *standard* wording for every held-to-expiry closed position, it's worth the one-line fix: `buildQuestion(stationMeta(code).name, ...)`. Please log as a follow-up issue.

[suggestion] Losing the per-position chart specifically for held-to-expiry positions is a real (if graceful) loss of review depth — these are arguably the trades a user most wants to inspect after the fact. Recommend a follow-up issue to explore snapshotting `question` text and enough chart data (or the `token_id`) onto the `trades` row at settlement time, before `db.close_position()` deletes the `open_positions` row — that would close this gap permanently without adding synchronous Gamma calls to `/api/portfolio`. Not a blocker now: the current fallback doesn't mislead or break, and the author correctly avoided a heavier fix outside this PR's read-path-only scope.

Neither item blocks merge. The fallback behavior is legible, doesn't crash or render blank, and the scope decision to leave `sold` positions on JSONL (keeping their charts) is the right call.

**Verdict: APPROVE** (UX/frontend impact). No blockers.

---

Issue #782, opened 2026-07-21
Edge tab: anchor Today/D+1 to market settlement window, not UTC calendar date


P0K5:
## Problem

The Edge tab's day partitions ("Today" vs "D+1") are anchored to the **UTC calendar date**, but every Polymarket temperature market closes at **noon UTC (`<label>T12:00:00Z`)** on its label date. These two facts interact badly:

- A label-date-`D` market is "same-day" (`endDate.date() == today_utc`) **only during 00:00–12:00 UTC** of day `D`.
- After 12:00 UTC it is closed, and the only *open* market for each station is label-date `D+1`, which the scanner classifies as next-day (`is_next_day=1`) and keys to tomorrow's date.

**Consequence:** for the entire **12:00–24:00 UTC** half of every day, the "Today" partition is empty *by construction* and all live/tradeable markets appear under "D+1". For an operator in a positive-UTC-offset timezone (e.g. WEST/UTC+1), this empty-"Today" window covers most of their waking/evening hours, and the currently-open market's *label date* often equals their **local** calendar date — so the tab reads as "today's data is filed under tomorrow."

This was diagnosed live on 2026-07-21: the entire freshly-populated `scan_decisions` table (created at that day's ~18:04 UTC restart, all writes in the 19:57–23:07 UTC window) held **only** `is_next_day=1` / `date=<tomorrow>` rows, zero same-day rows. Data was internally consistent — not corrupted — but the default view was empty.

## Current mitigation (option A, already shipped)

`loadEdgeStationData()` now auto-advances to the D+1 partition when "Today" is empty but D+1 has data, so the operator always lands on live data. This is a **UX band-aid**, not a semantic fix — the day-boundary confusion remains if you inspect the partitions directly.

## Proposed proper fix (option B)

Stop defining the Edge tab's "day" by `datetime.now(timezone.utc).date()`. Instead anchor the partitions to the **market's own settlement window / nearest open market**, so that:

- The "current" view always tracks the nearest open, still-settling market for a station regardless of UTC wall-clock.
- The date label shown to the operator matches the market's effective trading day (consider surfacing the market's local/label date rather than the raw UTC date).
- "D+1" means the genuinely-next settlement window, not "the same market that flipped across UTC midnight."

Touch points:
- `src/dashboard/api.py` `analysis_station()` — date resolution (`api.py:2949–2959`) and the `is_next_day` partition filter.
[...]

PR #779, opened 2026-07-21
feat(#758): rebuild Edge tab as per-bracket decision view

ClaudioSenior:
## Linked issue
Closes #758

## What changed
Rebuilds the Edge tab per `docs/design/edge-tab-bracket-decisions.md` (final story of epic #754), served entirely by the #757 endpoint (`GET /api/analysis/{station}?next_day=<bool>`, reading the persisted `scan_decisions` snapshot — never a live recompute).

- **Removed**: the ensemble-distribution histogram chart and the 5-card KPI strip (Ensemble Mean / **Bias-Corrected** / Members / Range / Best Edge).
- **New scan summary bar**: poll-freshness pill (`HH:MM:SS (Nm ago)`, amber "stale" styling when `poll_ts` is older than 2× `poll_interval_seconds`) + model-forecast-high / current-observed-high context; "No scan recorded for {station} today" empty state.
- **New collapsed "Forecast inputs" `<details>`**: ensemble mean / range / members, closed by default.
- **New per-bracket decision table** — one row per evaluated bracket, exactly the 5 AC columns: `Bracket / Market (YES·NO) / Model p_yes / EV (YES·NO) / Gate`.
  - Model p_yes cell shows the capped value bold, a muted `raw X%` caption when the 0.05/0.95 clamp fired, and the `emos_mode` inline chip.
  - Market/EV cells underline the `side` the scanner flagged (when populated) so the operator doesn't have to infer it.
- **Gate-verdict chip** (11 verdicts, 4 color families exactly per the design spec, reusing the existing `.status-badge`/`--yes`/`--no`/`--warn`/`--muted` tokens) — `title` + `aria-label` built from `gate_actual`/`gate_threshold`/`gate_unit`/`gate_detail`, showing e.g. `edge 12.4¢ < min 15¢`. Satisfies the intent of #529.
- **Client-side row emphasis**, mutually exclusive, precedence per spec: `row-traded` → `row-near-miss` (entry_guard/timeout_today, highest EV on its own flagged side) → `row-best-signal` (highest `max(ev_yes, ev_no)` across all rows, only when that value is actually > 0 — never manufactured from an all-negative table). Independent `row-thin` de-emphasis when both sides' EV is near zero.
[...]

Issue #687, opened 2026-07-11
[P0?] Daily ~12h evaluation blackout (11:45-00:04 UTC) from wrong_date gate vs shared ~12:00 UTC settlement — every day since at least 2026-06-11


ClaudioSenior:
## Problem

`wrong_date` gate (`src/strategy/scanner.py:440-441`) rejects any market whose settlement date
(UTC calendar date) isn't today's UTC calendar date, even when tomorrow's market is already
present in the fetched Gamma feed and today's market is past/near its `MIN_MINUTES_TO_SETTLEMENT`
cutoff. Because non-US weather markets settle once daily at a shared ~12:00 UTC instant, this
produces a **permanent, daily ~12h evaluation blackout from ~11:45 UTC to ~00:00-00:04 UTC**,
confirmed present in every day of available `bot.log` history (2026-06-11 through 2026-07-10,
30/30 days). Full evidence and mechanism: #686 (comment 2026-07-11).

US station active-hours windows (`STATION_ACTIVE_HOURS`) overlap heavily with this UTC dead
zone, so US stations' effective evaluable time is compressed to roughly a quarter of their
configured local active window.

## Proposed direction (needs design, not scoped/decided here)

Something like: once today's market for a station drops below `MIN_MINUTES_TO_SETTLEMENT`,
allow evaluating the *next* market for that station (by close time) rather than requiring
`end_dt.date() == today_utc`. Needs care around the METAR-validity assumption in the existing
comment ("our METAR data is only valid for today") — may need forecast-only evaluation (no
METAR-anchored intraday correction) for the just-opened next-day market until its own
observation window begins.

## Constraint

This is a live entry-gate change and must NOT be bundled into #686's watchdog PR or any Phase 2
data-plumbing PR. Deliberate, reviewed change only, per repo governance (CLAUDE.md: no
entry-gate changes without explicit scope).

## Acceptance criteria (draft — refine before implementation)

- [ ] Design proposal reviewed (Tech Lead PM) before code
- [ ] Fix (or explicit decision not to fix) implemented behind review
- [ ] Confirms via `snapshot_archive`/`bot.log` that the daily blackout is closed or
      intentionally narrowed
- [ ] No regression to the wrong-date safety property (never trade a bracket using another
      day's weather observations)

Refs #686, #682 diagnosis thread.

P0K5:
## Design note posted — requesting review before implementation

[...]

Issue #602, opened 2026-07-03
Light-theme --yes/--warn badge contrast is just under WCAG AA (design-system token audit)


P0K5:
## Context (non-blocking, found during Designer review of PR #598 / issue #559)

Designer review of the new Promotion tab computed WCAG contrast for the status badges in both themes:
- Dark theme: comfortable (5.0–7.8:1).
- **Light theme:** green (`--yes` on `--yes-bg`) ≈ **4.47:1** and amber (`--warn` on `--warn-bg`) ≈ **4.60:1** for the ~11px bold badge text — green is just under the 4.5:1 AA threshold, amber is borderline.

## Important distinction from #584

This is **not** a repeat of #584 (which was a bespoke `#6a7290` badge at ~3.35:1). PR #598 correctly **reused the app's existing design tokens** (`--yes`/`--warn` + `-bg`), which already carry this same marginal light-theme contrast elsewhere in the app (e.g. `.badge-active`, `.bias-badge-applied`). A local fix in #598 would be the wrong place — this is a shared-token issue.

## Fix

Audit and nudge the light-theme `--yes`/`--warn` foreground/background token pair app-wide so badge text clears 4.5:1, then verify existing consumers (`.badge-active`, `.bias-badge-applied`, promotion badges) still read correctly in both themes.

## Labels
frontend

ClaudioSenior:
New consumer of the `--warn`/`--warn-bg` pair: the Edge tab's persistent D+1 indicator (`.edge-d1-indicator`, issue #787 / PR #783) — a banner shown whenever the D+1 partition is being viewed, reusing the same amber pair already backing `gate-shadow_only`, `badge-shadow`, `badge-disabled`, etc.

Flagging it specifically because it's the highest-stakes consumer of this pair so far: unlike a supplementary badge sitting next to other content, this element's entire purpose is *being noticed* — it's the sole in-content signal that the operator is looking at a different day's data. Contrast independently re-verified at `#9a6010` on `#fef0d8` = 4.60:1, clearing the 4.5 AA threshold for normal text, so no action needed here — just noting it as a data point for whatever token-level fix eventually lands on this issue, since it'll propagate to this indicator too.

PR #783, opened 2026-07-21
fix(#782): auto-advance Edge tab to D+1 when Today is empty

### LINKED ISSUE #787: Edge tab: auto-advance to D+1 when Today partition is empty
## Context

The Edge tab's Today/D+1 partitions are anchored to the UTC calendar date, but
every Polymarket temperature market closes at noon UTC on its label date. So a
label-date-`D` market is "same-day" only during 00:00-12:00 UTC; after 12:00 UTC
the station's only open market is label-date `D+1`.

The consequence is that for the **entire 12:00-24:00 UTC half of every day** the
Today partition is empty by construction and every live market sits under D+1.
Operators in positive UTC offsets see "No scan recorded for … today" while D+1
quietly holds all the data, which reads as broken. Confirmed live 2026-07-21:
`scan_decisions` held only `is_next_day=1` rows, zero same-day rows — internally
consistent, just an empty default view.

This issue tracks the **UX mitigation only**: land the operator on the partition
that actually has data. The semantic fix — re-anchoring the day boundary to the
market's settlement window instead of the UTC calendar date — is tracked
separately in #782 and stays open.

Implemented by PR #783, which currently needs an issue to close and test
coverage for the new branch.

## Acceptance criteria

- When the Today partition returns no brackets and D+1 has brackets,
  the Edge tab renders the D+1 view with the D+1 date button active.
- The already-fetched D+1 payload is reused — no additional network request.
- When Today has brackets, behaviour is unchanged (lands on Today).
- When both partitions are empty and the fetch failed, the existing error
  banner path still runs.
- "Today" remains reachable in one click via the existing date toggle.
- No backend or data-model change; frontend only.

## Technical notes

`src/dashboard/static/index.html`, `loadEdgeStationData()` (~line 2659).
Both partitions are already fetched in parallel, so the auto-advance renders
from the in-hand `d1Data` local — not from `edgeD1Cache` — and returns early,
skipping the Today-rendering path.

## Test requirements

`src/tests/test_edge_tab_js_logic.py` (executes the inline script under Node):

- auto-advance fires when Today is empty and D+1 has data
- auto-advance does NOT fire when Today has data
- auto-advance does NOT fire when both are empty (falls through to error path)

## Dependencies

Depends on #786 (flaky gate tests) landing first so CI is trustworthy.
Related: #782 (the real fix, intentionally left open).

P0K5:
Closes #787
Refs #782

## What

Auto-advance the Edge tab to the **D+1** partition when **Today** comes back empty but D+1 has data. `loadEdgeStationData()` already fetches both partitions in parallel; it now renders the D+1 data it already has instead of showing an empty "Today" view.

**Update (Designer change request, addressed in this PR):** the toggle button's `.active` tint alone wasn't enough of a signal that the view had changed on trading UI, especially once landing on D+1 became the default for ~12h/day rather than a deliberate click. Added:
- A **persistent in-content D+1 indicator** — a textual banner ("Today's markets closed — showing D+1 (shadow-only)") inside the content area itself, shown whenever D+1 is displayed (auto-advanced *or* manually toggled), surviving every re-render.
- Fixed the **"Best signal today" caption** to be day-aware ("Best signal D+1" when viewing D+1). This was a pre-existing bug (reachable since v1 via the manual D+1 toggle), not introduced by this PR — but this PR raises its severity from "minor, deliberate click" to "material, default view for ~12h/day," so it's fixed here since it's now in this PR's blast radius.
- `docs/design/edge-tab-bracket-decisions.md` updated to reflect both.

## Why

The Edge tab's Today/D+1 partitions are anchored to the **UTC calendar date**, but every Polymarket temperature market closes at **noon UTC (`<label>T12:00:00Z`)** on its label date. These interact badly:

- A label-date-`D` market is "same-day" only during **00:00–12:00 UTC** of day `D`.
- After 12:00 UTC it's closed, and a station's only open market is label-date `D+1` (classified `is_next_day=1`) until UTC midnight.

So for the **entire 12:00–24:00 UTC half of every day**, the Today partition is empty *by construction* and all live markets sit under D+1. Operators in positive UTC offsets (e.g. WEST/UTC+1) hit this for most of their waking/evening hours and saw "No scan recorded for … today" while D+1 quietly held all the data — which reads as broken.

Diagnosed live 2026-07-21: the freshly-populated `scan_decisions` table held only `is_next_day=1` rows, zero same-day rows — internally consistent, just an empty default view.

## How

- [`index.html`](../blob/fix/edge-tab-auto-advance-d1-782/src/dashboard/static/index.html) `loadEdgeStationData()`: after fetching both partitions, if Today has no brackets and D+1 does, set `edgeIsNextDay = true`, activate the D+1 button, and render from the `d1Data` local already in hand (no extra request), then return. This renders the freshly-fetched payload directly, not the `edgeD1Cache` object — that cache is only read later, by `selectEdgeDate()`, and is already guarded there by `edgeD1Cache.station === edgeSelectedStation`.
- Unchanged paths: 00:00–12:00 UTC (Today has data → renders Today), and the fetch-error path (`todayData === null` with no D+1 → error banner).
- "Today" remains one click away via the existing date toggle.
- New `renderEdgeD1Indicator(isNextDay)` toggles `#edge-d1-indicator`'s visibility/text; called from the auto-advance branch, the Today fallthrough (hides it), and `selectEdgeDate()` (manual toggle, both directions) — every render path that can change which partition is showing.
- `renderEdgeDecisionTable`'s best-signal caption (line ~3092) now reads the `isNextDay` param it already receives, instead of hardcoding "today".

## Scope / safety

- **Frontend only** (+ one design doc), no backend or data-model change.
- No impact on trading logic, EMOS/DEB weights, or promotion — those read `model_forecast_log`/`candidates`/`trades`/`emos_calibration`, never `scan_decisions`.
- This is a **UX mitigation**, not the semantic fix. Reframing the day boundary around the market's settlement window (rather than the UTC calendar date) is tracked in **#782** — intentionally left open.

## How to test

1. Load the dashboard Edge tab during **12:00–24:00 UTC** (Today empty, D+1 populated) → tab lands on the **D+1** view with the D+1 button active, no empty "No scan recorded today" state, and the in-content banner "Today's markets closed — showing D+1 (shadow-only)" is visible above the scan summary bar.
2. Load during **00:00–12:00 UTC** (Today populated) → lands on **Today** as before, no D+1 banner.
3. Click **Today** from the auto-advanced D+1 view → the empty Today partition renders (still reachable), D+1 banner disappears.
4. Manually click **D+1** on a station whose Today already has data → D+1 banner appears; if any bracket is the best-signal row, its caption reads "Best signal D+1" (not "today").
5. Automated: `python -m pytest src/tests/test_edge_tab_js_logic.py` (executes the full inline script under Node — covers the three auto-advance cases, the manual-toggle case, the day-aware caption, and the D+1 indicator's actual rendered text/visibility; confirms existing gate-chip/row-emphasis logic is unaffected).

## How to deploy

- [x] **Code only** — `git pull` + restart `meteoedge-dashboard.service` (static file served via `StaticFiles`, which reads from disk per request, so a restart isn't strictly required — but it clears any browser/proxy cache confusion, matching the convention used in #779).


P0K5:
## Tech Lead review — request changes

Triaged against the GLM-5.2 review. Verdict stands as **BLOCK**, but two of the
three AI blockers need correcting, and the red CI is not this PR's fault.

### CI: the 3 pytest failures are unrelated to this PR

`test_scan_decisions.py::TestGateVerdicts` failures are **not** caused by this
diff — it is 20 lines of frontend JS in `index.html` and cannot affect scanner
gating. They are time-of-day flaky:

- `test_below_min_confidence_no_side` builds its market at `_today_end()`
  (23:59:59Z) and, unlike its sibling tests, does not patch
  `MIN_MINUTES_TO_SETTLEMENT`/`MODEL_PROB_CAP`. This CI run was at 23:40Z —
  ~19 min to settlement — so the edge grows and `above_max_edge` fires first.
- Both `mae_gate` tests use `now + 3h`, which at 23:40Z resolves to 02:40Z the
  *next* UTC day, so the market is classified next-day and filtered out —
  hence `len(candidates) == 0`.

All 12 pass locally at 15:54Z. Master is green only because its runs landed in
safe UTC hours. Tracked and delegated as **#786**; that lands on master first,
then this PR rebases so its CI result is trustworthy at any hour.

### Blocker 1 — missing `Closes #N`: valid, with a correction

Governance rule 3 applies, but `Closes #782` would be **wrong**. #782 is the
semantic fix (re-anchoring the day boundary to the market settlement window)
and must stay open; this PR is an interim UX mitigation. Created **#787** to
represent the mitigation. Use `Closes #787` + `Refs #782`.

### Blocker 2 — no tests: valid

The new `!todayHasData && d1HasData` branch is operator-facing trading UI and
needs the three cases the reviewer listed, in
`src/tests/test_edge_tab_js_logic.py`.

### Blocker 3 — "stale `edgeD1Cache` rendered as live data": **false positive**

The auto-advance block renders `d1Data` — the local just fetched in this same
call, assigned to the cache on the line immediately above — not `edgeD1Cache`.
The cache *read* path at ~line 2700 is already guarded by
`edgeD1Cache.station === edgeSelectedStation`. No cache-invalidation guard is
warranted; adding one would be redundant code. Not a blocker.

### Additional blocker not raised by the AI reviewer

**Missing "How to deploy" section.** The PR body has What/Why/How/Scope/How to
test but no deploy section, which `.github/pull_request_template.md` requires.
Add the real ship steps or an explicit "No deploy needed".

### Also required

Frontend PR → **Designer approval is mandatory** before merge. Designer review
requested in parallel. Flagging honestly: the open UX question is whether
silently switching the operator's default view from Today to D+1 is right for
trading UI, versus an explicit empty state that explains *why* Today is empty.
If the Designer rejects the interim, this PR changes shape.

Fixes delegated. Moving #787 to In progress.

ClaudioSenior (changes requested):
## Designer review — Requesting changes

Verdict: the auto-advance mechanism (reuse the already-fetched D+1 cache, no extra request) is fine engineering, but as implemented it trades one UX problem for a worse one. Requesting changes before merge.

### Why the silent switch is worse than the empty state it replaces

The board thread frames this as "silent auto-switch vs. explicit empty-state" — I checked what actually renders in each, not just the toggle:

1. **The only visual signal that the view changed is the toggle button's `.active` class** (`index.html:2680-2685`) — a small pill going from grey to tinted/bold. Everything else in a populated D+1 view (scan bar, forecast inputs, decision table) renders through the *exact same* functions as Today, with no in-content "you are viewing D+1" marker.
2. **Most rows carry zero day signal.** Only brackets that reach candidate status get the `next_day_shadow` → "Next-day" gate chip (`scanner.py:934,992`). The routine majority case — rows rejected on the numbers (`below_min_edge`, `margin_gate`, etc., per the legend at `index.html:2833-2834`) — renders an *identical* chip whether it's Today or D+1. An operator scanning rows sees nothing that says "D+1."
3. **The "Best signal today" caption is actively wrong in this state.** `index.html:3054` hardcodes the word "today" regardless of `isNextDay`. In the auto-advanced case this literally mislabels a D+1 evaluation as today's. That's not a hypothetical misread risk — it's a shipped false statement on the page.
4. Before this change, an operator landing on an empty Today view got an unambiguous (if plain) message: `No scan recorded for {station} today`. That's boring but it cannot be misread as live data. After this change, for the ~half of every day that Today is empty by construction, the *default* view an operator opens the dashboard into is full of real-looking numbers that are D+1, distinguished only by a toggle-button tint most people won't consciously register on load.

Given this is trading UI and Today vs. D+1 genuinely changes what the numbers mean (D+1 rows are `is_next_day_eval` / always shadow-routed, per `scanner.py:922-928` — they can never correspond to a live position today), a misread here has real money-decision consequences, not just cosmetic confusion. Priority-1 in our design principles is clarity about "where am I" — this regresses it for the majority of the day.

### Requested changes

Pick one of these two directions — either is acceptable, silence is not:

**A. Keep the auto-advance, but make the D+1 state impossible to miss.** Add a persistent, textual indicator inside the content area itself (not just the toggle), e.g. a line in the scan bar or a small banner above it: `"Today's markets are closed for {station} — showing D+1 (shadow-only, not live)."` This should render whenever `edgeIsNextDay` is true, including on manually-selected D+1 (good for consistency, not just the auto-advance path). Also fix the `"Best signal today"` caption at `index.html:3054` to read `"Best signal for D+1"` when `isNextDay` — right now it's a straightforward copy bug that this PR's own new code path newly makes reachable in the default view.

**B. Skip the silent full-content switch; upgrade the empty state instead.** Keep landing on Today, but replace the bare `"No scan recorded for {station} today"` message with an informative empty state that explains *why* (markets close at noon UTC) and offers a one-click link to D+1 — e.g. `"Today's markets are closed. {N} bracket(s) are open for D+1." [View D+1 →]`. This preserves an unambiguous "you clicked to get here" mental model for the money-relevant view, at the cost of one extra click during the empty half of the day.

I lean toward **A** for efficiency (matches the PR's stated goal, doesn't cost operators a click), but it is not optional to ship it without the explicit in-content label — the toggle-button-only signal is not sufficient on its own, especially for a state that's the *default* for 12 of 24 hours.

### Non-blocking

- Worth a follow-up issue (not a blocker here): the spec at `docs/design/edge-tab-bracket-decisions.md:34` currently documents Today as the unconditional default on load — that's now stale given this PR's behavior change and should be updated once the direction above is settled, so implementation and spec don't drift.

Scope/engineering (cache reuse, no extra fetch, error-path untouched) all look correct — this is purely about the missing "where am I" signal.

P0K5:
## Tech Lead decision — Designer change-request accepted, Direction A

Designer's review is upheld. I verified both load-bearing claims independently:

- **Caption bug is real.** `index.html:3054` hardcodes "Best signal today".
  `renderEdgeDecisionTable` already receives `isNextDay` (line 3007) and already
  uses it correctly for the empty state (lines 3019-3020) — line 3054 simply
  never got the same treatment.
- **"Shadow-only" is factually accurate.** `scanner.py:928` and `:986` force
  `shadow=(shadow_* or is_next_day_eval)`, so next-day candidates are never live
  positions (per #687). The proposed indicator wording is safe to state as fact.

**One correction to the framing:** the caption bug is *pre-existing*, not
introduced by this PR — it is already reachable today by manually clicking the
D+1 toggle. What this PR changes is its severity: from "minor, requires a
deliberate click" to "material, it is the default view for ~12h of every day".
It gets fixed here because it sits squarely in this PR's blast radius and makes
the auto-advance actively assert something false on screen — not because it is
this PR's regression.

**Direction chosen: A** (keep auto-advance, label it in-content). It matches the
PR's stated goal, costs no extra click, and is the Designer's own lean. Direction
B remains valid but loses the benefit the PR exists to deliver.

Required before re-review:

1. Persistent in-content D+1 indicator — visible on *both* the auto-advanced and
   manually-toggled D+1 states, surviving re-render. The toggle `.active` tint
   alone is insufficient for trading UI.
2. Make the `Best signal today` caption day-aware.
3. Update `docs/design/edge-tab-bracket-decisions.md` (stale load-default).
4. Test coverage extended to the indicator: renders when auto-advanced, renders
   on manual D+1, absent on Today.

Plus the items from my earlier review: `Closes #787` + `Refs #782`, the three
auto-advance branch tests, the **How to deploy** section, and the written
rebuttal of AI blocker 3.

Delegated to the assigned mid-dev. CI remains gated on #786 landing first.

ClaudioSenior:
Re: blocking item 3 (`edgeD1Cache` staleness) — verified, this is a false positive.

The auto-advance branch at `loadEdgeStationData()` line ~2678 calls `renderEdgeDecisionTable(d1Data.brackets, unit, metar, true)`, where `d1Data` is the local from the `Promise.all` fetch a few lines above — the freshly-fetched payload for the station being loaded, not `edgeD1Cache`. `edgeD1Cache` is written right after that same fetch (`edgeD1Cache = { station: metar, data: d1Data }`) but isn't *read* in this code path at all.

The only read of `edgeD1Cache` is in `selectEdgeDate()`, and it's already guarded there: `(wantNextDay && edgeD1Cache.station === edgeSelectedStation) ? edgeD1Cache.data : await fetchEdgeData(...)` — falls back to a fresh fetch if the cached station doesn't match the currently selected one, so it can't render stale cross-station data.

No code change here; addressed items 1 (`Closes #787` added, keeping `Refs #782` since that's the real semantic fix and stays open) and 2 (three new test cases in `src/tests/test_edge_tab_js_logic.py` covering the auto-advance branch) in the latest push.

ClaudioSenior:
Scope-change addendum pushed (commit 13d4b15) per the Tech Lead PM / Designer's change request:

- **Persistent in-content D+1 indicator** — new `#edge-d1-indicator` banner (`renderEdgeD1Indicator()`), shown whenever D+1 is the active view (auto-advanced *or* manually toggled via `selectEdgeDate()`), text "Today's markets closed — showing D+1 (shadow-only)". Survives every re-render, no interaction needed to see it.
- **Day-aware "Best signal" caption fix** — `renderEdgeDecisionTable` (line ~3092) now reads its own `isNextDay` param instead of hardcoding "today". Pre-existing bug, not introduced by this PR (already reachable via the v1 manual D+1 toggle) — fixed here because auto-advance raises its severity to the default view for ~12h/day.
- `docs/design/edge-tab-bracket-decisions.md` updated (annotated corrections, matching this doc's existing convention) to document both.
- Test coverage extended: asserts the indicator's actual DOM text/visibility and the caption's actual rendered string for both `isNextDay` states — not the date-toggle button's `.active` class, per the explicit re-review criterion.

ClaudioSenior:
Rebased onto latest master (includes #788's frozen-clock fix for the day-boundary-adjacent `test_scan_decisions.py` tests) and force-pushed — head is now `c0f0341`. CI is fully green on this rebased head: lint, pytest (including the previously-flaky `TestGateVerdicts` cases), gitleaks, AI / NVIDIA NIM review (verdict PASS, no blocking issues). This green run is no longer clock-window luck.

ClaudioSenior:
Final review-fix round pushed (commit 63e275f, on top of the rebased c0f0341):

1. `loadEdgeStationData()` now resets the D+1 indicator synchronously at the top of the function (alongside the other state resets), not only in the post-fetch fallthrough — fixes the previous station's D+1 banner lingering through a station switch's fetch round-trip. Added a regression test that checks indicator state before the fetch promise resolves.
2. Indicator icon swapped `alert-triangle` → `info` (background/text unchanged) — `alert-triangle` is already used elsewhere on this tab for a genuine adverse-position signal; `info` correctly reads as orientation for this routine, ~12h/day state.
3. Posted a comment on #602 noting this indicator as a new `--warn`/`--warn-bg` consumer — no code change, contrast independently re-verified at 4.60:1 by the Designer.

CI running on 63e275f now.

ClaudioSenior:
CI on 63e275f: all green. Note the first "AI / NVIDIA NIM review" run hit a transient NIM API timeout (`Read timed out. (read timeout=120)` calling `integrate.api.nvidia.com`) — infrastructure flake, not a code finding. Re-ran the job (`gh run rerun --failed`) and it came back clean: verdict PASS, no blocking issues. Ready for final review.

P0K5:
## Tech Lead technical sign-off — Designer approval still required

> **Governance note:** this is recorded as a comment, not a GitHub `APPROVED`
> review, because this PR was authored by `P0K5` — the same account the Tech Lead
> PM operates under — and GitHub refuses self-approval. Treat this comment as the
> Tech Lead technical sign-off of record. See the process gap noted at the bottom.

Verified against the file at head `63e275f` directly, not the author's summary.
All three follow-up items confirmed present:

- **Icon swap** — `index.html:946` is `data-lucide="info"`. The only remaining
  `alert-triangle` occurrences are the explanatory comment at 939 and the
  legitimate adverse-position signal at 1263 ("market moved against model").
  Collision resolved.
- **Indicator reset** — `loadEdgeStationData()` line 2677 calls
  `renderEdgeD1Indicator(false)` alongside the other synchronous state resets,
  with the post-await fallthrough call left intact. The reset-vs-decision split
  is correct.
- **#602 comment** — posted, correctly scoped, no code change.

**Test quality meets the re-review criterion.** Assertions read the indicator's
actual `style.display` and `textContent`, and the caption's rendered HTML string
— not toggle `.active`, which would pass with the indicator missing entirely.
Case 4 covers manual toggle in both directions; case 5 asserts the reset lands
*synchronously* before the fetch resolves; caption assertions are both positive
and negative. Author verified the suite fails against a reverted implementation.

**CI green on head `63e275f`:** `Lint (flake8)`, `Test (pytest)`,
`AI / NVIDIA NIM review`, `gitleaks`. This run is meaningful — the branch was
rebased onto master after #788 merged, so the previously flaky
`test_scan_decisions.py` cases are genuinely fixed rather than dormant. Earlier
green runs on this branch were clock-window artifacts.

Branch is 2 commits behind master (#790 shadow-pnl fix + a graph chore). Both
backend-only, no overlap with this diff; GitHub reports `MERGEABLE/CLEAN`. No
rebase required.

**Not merging** — per governance rule 5 this is a frontend PR requiring Designer
approval in addition to the Tech Lead's. Over to the Designer for UX sign-off on
the indicator wording, placement, and the `info` icon.

---

**Process gap for follow-up:** governance rule 5 assumes the Tech Lead can post a
formal approving review, but that is impossible on any PR the PM account itself
authored. Worth deciding whether PM-authored PRs should be reassigned, or whether
a comment-of-record is the accepted substitute.

ClaudioSenior (approved):
## Designer re-review — Approved

Verified independently against `63e275f` (not just the summary): checked out the head commit, read the actual diff, and ran the test suite myself.

- **Icon:** `info` at `index.html:946`, with an in-file comment explaining the `alert-triangle` conflict. Confirmed the only remaining `alert-triangle` uses are the explanatory comment and the pre-existing "market moved against model" case — no dilution of that signal.
- **Contrast (#602):** comment posted on #602 naming this consumer, correctly recording 4.60:1 and flagging it as the highest-stakes consumer of the pair without proposing a local override. Matches what I asked for — no action needed on this PR.
- **Synchronous reset fix:** verified by breaking it locally (removed the early `renderEdgeD1Indicator(false)` call) and confirming `test_edge_tab_js_logic.py::test_edge_tab_gate_and_emphasis_logic` fails on `case 5` with exactly the expected assertion (`indicator must reset before the fetch resolves, not after`). Restored and re-ran clean — passes.
- **Caption fix:** `index.html:3101` is day-aware, and the test asserts both directions positively and negatively against the real rendered table HTML, not just presence/absence.
- **Test coverage:** all 5 cases (auto-advance, no-advance, fetch-error path, manual toggle both directions, sync-reset regression) assert on `style.display`/`textContent` read off the actual DOM stub, not the toggle's `.active` class — this was my explicit criterion and it holds.
- **Spec doc:** updated using the PR #779 annotated-correction convention, covers the auto-advance behavior change, the indicator, the day-aware caption, and carries an explicit `[Tech constraint: ... agreed approach: ...]` note on #602 in the accessibility section.
- **CI:** green on `63e275f` (lint, pytest, gitleaks, NIM review).

All four items from my change request, plus the two follow-up calls (icon, contrast), are in and verified. Approving.

PR #788, opened 2026-07-22
fix(#786): freeze wall clock in day-boundary-adjacent gate tests

### LINKED ISSUE #786: test_scan_decisions gate tests are time-of-day flaky (fail when CI runs late UTC)
## Context

Three tests in `src/tests/test_scan_decisions.py::TestGateVerdicts` pass during
most UTC hours but fail when CI runs late in the UTC day. Observed on PR #783
(CI run at 23:40 UTC), whose diff is 20 lines of frontend JS in
`index.html` and cannot possibly affect scanner gating.

```
FAILED test_below_min_confidence_no_side - assert 'above_max_edge' == 'below_min_confidence'
FAILED test_mae_gate_forces_no_to_shadow - assert 0 == 1
FAILED test_mae_gate_does_not_fire_below_threshold - assert 0 == 1
```

All 12 tests in the class pass locally at 15:54 UTC. Master appears green only
because its runs have happened to land in safe UTC hours — this is a latent
CI-integrity bug that will randomly block unrelated PRs.

## Root cause

**`test_below_min_confidence_no_side` (line 129)** builds its market with
`_today_end()` = today at 23:59:59Z. At 23:40 UTC that is ~19 minutes to
settlement. Unlike its sibling tests in the same class (lines 84-85, 96-97,
222-223) this test does **not** patch `MIN_MINUTES_TO_SETTLEMENT` or
`MODEL_PROB_CAP`. As settlement approaches, the remaining-climb/time term
shrinks, the distribution sharpens, `p_yes` goes extreme and the edge grows —
so `above_max_edge` fires before `below_min_confidence` is ever reached.

**`test_mae_gate_forces_no_to_shadow` (line 152)** and
**`test_mae_gate_does_not_fire_below_threshold` (line 187)** both set
`end_time = now + timedelta(hours=3)`. Run at 23:40 UTC that resolves to
02:40Z on the *following* UTC day, so the market is classified as next-day
rather than same-day and is filtered out of the candidate set — hence
`len(candidates) == 0` instead of the expected 1.

Note `_kmia_state()` already documents this exact hazard and pins `now_local`
30 minutes past sunset "instead of time-of-day flaky" — the same discipline was
not applied to the settlement timestamps.

## Acceptance criteria

- The three named tests pass deterministically at **any** wall-clock UTC hour,
  including 23:00-23:59Z and 00:00-00:59Z.
- The fix removes the wall-clock dependency rather than widening assertion
  tolerances — the gate semantics under test must stay exactly as asserted
  today (`below_min_confidence`, `mae_gate`, `traded_live`).
- No other test in `test_scan_decisions.py` regresses.
- No production code in `src/strategy/scanner.py` is modified — this is a test
  determinism bug, not a scanner bug.

## Technical notes

Preferred approach: make the market settlement timestamps independent of the
current wall clock rather than relative to `datetime.now()`.

- For the two `mae_gate` tests, anchor `end_time` so it always lands inside the
  same UTC day as `now` (e.g. derive a fixed same-day settlement instant, or
  freeze `now` to a mid-day UTC value and build both the `WeatherState` and the
  market from that single frozen instant).
- For `test_below_min_confidence_no_side`, either patch
  `MIN_MINUTES_TO_SETTLEMENT`/`MODEL_PROB_CAP` to match its sibling tests, or
  use a settlement time far enough out that the confidence gate is the binding
  constraint. Verify which by checking that `gate_actual > gate_threshold ==
  0.05` still holds, per the existing assertion on line 137.
- Whichever mechanism is chosen, apply it consistently and leave a short
  comment in the same spirit as the `_kmia_state()` docstring.

## Test requirements

- Demonstrate determinism by running the class with the clock forced to a late
  hour (e.g. `freezegun`, or a monkeypatched `datetime.now`) for 23:40Z, 23:59Z,
  00:10Z and 12:00Z, all green.
- Full `python -m pytest src/tests/test_scan_decisions.py` green.

## Dependencies

None. This should land on master **before** PR #783 is re-run, so that PR's CI
result is trustworthy at any hour.

ClaudioSenior:
Closes #786

## What changed

Three tests in `src/tests/test_scan_decisions.py::TestGateVerdicts` built their market/state timestamps relative to the real `datetime.now()`:

- `test_below_min_confidence_no_side` used `_today_end()` (today at 23:59:59Z) — late in the UTC day the remaining-time-to-settlement term shrinks, the distribution sharpens, and `above_max_edge` fires before `below_min_confidence` is ever reached.
- `test_mae_gate_forces_no_to_shadow` / `test_mae_gate_does_not_fire_below_threshold` used `now + timedelta(hours=3)` — run late enough in the UTC day, that rolls onto the next calendar date, so the market gets classified as next-day and filtered out of the candidate set entirely (`len(candidates) == 0` instead of 1).

Fix: freeze `src.strategy.scanner`'s internal `datetime.now(timezone.utc)` reads (the same-day filter and the minutes-to-settlement math) to a fixed mid-day UTC instant for just these three tests, via a local `datetime` subclass + context manager (`_FrozenDatetime` / `_frozen_scanner_now`) — same spirit as the file's existing `_kmia_state()` sunset-pinning trick. Both the market end time and the `WeatherState` for the mae-gate tests are derived from that same frozen instant, so scanner's two internal wall-clock reads and the test's own timestamps always agree, independent of the hour CI actually runs in.

No production code changed — `src/strategy/scanner.py` is untouched, and the asserted gate verdicts (`below_min_confidence`, `mae_gate`, `traded_live`) are exactly as before.

## Why

CI runs late in the UTC day (e.g. PR #783 at 23:40 UTC) were failing these three tests despite the diff being unrelated to scanner gating — a latent CI-integrity bug that would randomly block unrelated PRs.

## How to test

```
python -m pytest src/tests/test_scan_decisions.py -q
```

All 22 tests pass. Determinism of the three previously-flaky tests was verified by monkeypatching the ambient wall clock to 23:40Z, 23:59Z, 00:10Z and 12:00Z and confirming all three stay green at every hour (and, as a sanity check on the harness itself, that they reproduce the original failures at 23:40Z/23:59Z against the pre-fix version of the file).

## How to deploy

- [x] No deploy needed (test-only change)

## Trade-offs / uncertainties

- Went with a manual `datetime` subclass rather than adding `freezegun` as a new test dependency, to stay within "no new major dependencies without approval" — happy to switch if the team prefers a proper freezing library.
- Blocked on the board-status "In progress" transition — `GITHUB_TOKEN_OPERATIONAL` can't resolve the Project board via GraphQL (separate issue, flagged to Tech Lead PM on #786).


P0K5:
## Tech Lead review — approve pending one fix

Verified all checks green on head `8427bcaa`: `Lint (flake8)`, `Test (pytest)`,
`AI / NVIDIA NIM review`, `gitleaks`. **How to deploy** present and correct
("No deploy needed — test-only"). Acceptance criteria on #786 met: the three
named tests are deterministic across the day boundary, gate semantics unchanged,
`src/strategy/scanner.py` untouched.

Approach is sound. Freezing scanner's `datetime.now()` to a fixed mid-day UTC
instant and deriving the market end time from that *same* instant fixes the real
defect — the two clock reads inside `scan_markets` disagreeing with the
test-constructed settlement time. Choosing a `datetime` subclass over adding
`freezegun` was the right call on dependency discipline.

Particular credit for confirming the harness actually **reproduces** the original
failures at 23:40Z/23:59Z against the pre-fix file before validating the fix.
That is the step that separates a real fix from one that merely makes tests pass.

### `[suggestion]` — resolve before merge

`_frozen_scanner_now` has no `try/finally`. If a test body raises inside the
context, `_FrozenDatetime._fixed = None` never runs and class-level state leaks
into whatever executes next. Impact is low today (`patch` still unwinds, and each
call site re-sets `_fixed` on entry), but this is mutable global state in a
fixture whose entire purpose is guaranteeing determinism, and the
`assert cls._fixed is not None` guard stops protecting anything once leaked.
Wrap the `yield` in `try/finally`. Requested from the author directly.

### Verified independently — no action needed

- I suspected `test_next_day_shadow` was also day-boundary flaky
  (`now + 1d6h` rolls to D+2 when run at ≥18:00Z). It is **not**:
  `eligible_next_day_date` is `min(_dates)` over the same market list passed in,
  so a single-market test always self-matches. Correctly left alone.
- The remaining six `_today_end()` tests all patch `MIN_MINUTES_TO_SETTLEMENT=0`
  and `MODEL_PROB_CAP=1.0`, so they are protected from the same failure mode.
  Scope of this PR is right.

Merging once the `try/finally` lands and CI is green. This unblocks the rebase of
PR #783, whose recent clean pytest run was luck of the clock rather than evidence
the flakiness was gone.

P0K5 (approved):
Approved. `try/finally` fix verified in 4b80a4d (+5/-3, scoped to the context manager). All checks green on head `4b80a4d`: Lint (flake8), Test (pytest), AI / NVIDIA NIM review, gitleaks. Acceptance criteria on #786 met — three previously-flaky tests now deterministic across the UTC day boundary, gate semantics unchanged, `src/strategy/scanner.py` untouched.

PR #836, opened 2026-07-24
feat(#798): replace EMOS's hard 60-sample cutover with a shrinkage blend

P0K5:
## Linked issue
Closes #798

## What changed
- `src/model/emos_calibration.py`: new `shrinkage_weight(n_city, full_weight_samples=60)` (linear ramp `clamp(n_city/60, 0, 1)`, exactly 1.0 at the legacy 60-sample threshold) and `blend_coefficients(city_coeffs, pooled_coeffs, weight)` (componentwise weighted average; convexity preserves sigma positivity automatically).
- `scripts/run_emos_shadow.py`: replaced the hard two-pass cutover (pure per-city fit if `n≥60`, else 100% pooled-group fallback) with a single blended pass. Cities at/above 60 samples get their pure per-city fit, bit-for-bit identical to pre-#798 output (verified by test). Cities in `[5, 60)` get a genuine blend weighted by their own sample count. Cities below 5 are still skipped (unchanged floor from #659).
- `src/tests/test_emos_shrinkage.py` (new, 16 tests): unit tests for the weight/blend functions, boundary-preservation at `n=60` and `n<5`, and an LLBG (Tel Aviv)-scale backtest at the 21-triple window size from #767 (`weight = 21/60 ≈ 0.35`), verified against the issue's own text.

## Design trade-off flagged for review
Chose a **linear ramp** (`n/60`, capped at 1.0) over an asymptotic Empirical-Bayes curve (`n/(n+k)`) specifically so the weight is exactly 1.0 at the legacy 60-sample threshold — guaranteeing zero behavior change for any city that already cleared the old bar. This is less statistically "principled" than an EB curve; documented inline in `shrinkage_weight`'s docstring as an explicit alternative. Worth a second opinion before merge.

## How to test
`pytest src/tests/test_emos_shrinkage.py -v` (16 passed). Full suite: 2555 passed, 9 skipped, 0 failed (171 pre-existing EMOS tests pass unmodified, confirming zero regression at both legacy boundaries). `flake8 src/` clean.

## How to deploy
- [ ] **One-off script(s) to run after deploy** — next `run_emos_shadow.py` invocation will use the blended path automatically; no manual step, but first post-deploy run should be spot-checked against a known city's prior coefficients to confirm the ≥60-sample cities are unchanged.
- [x] `git pull` + restart `meteoedge-run` (which invokes `run_emos_shadow.py`)

---
[...]

Issue #798, opened 2026-07-22
EMOS: partial-pooling/shrinkage blend instead of hard 60-sample per-city cutover


P0K5:
## Background

Follow-up from #767 (Tel Aviv/LLBG diagnosis, accepted). `run_emos_shadow.py`'s Pass 1/Pass 2 fit selection is a hard cutover: cities with \<60 own training triples fall through entirely to their pooling group's shared `(a,b,c,d)` coefficients (`fetch_training_data_pooled`), cities with ≥60 get a pure per-city fit. #767 showed this hurts low-sample cities inside heterogeneous pooling groups: LLBG's raw forecast is already near-unbiased (mean residual -0.03°F), but `midlat_c`'s shared `(a,b)` — fit to a very heterogeneous group spanning Warsaw to Wellington to Tokyo to Shanghai — injects a +1.67°F systematic overshoot when applied to it, reproducing ~100% of the observed CRPS regression.

A partial-pooling / shrinkage-weighted blend between the per-city fit and the pooled-group fit (weighted by `n_city` own-sample count, e.g. Bayesian shrinkage or a simple linear blend ramping from 0% per-city weight at `n=0` to 100% at `n=60`) would soften exactly this failure mode for any low-sample city in a heterogeneous group — not just LLBG — without needing a full per-city fit.

## Task

- Design and implement a shrinkage/partial-pooling blend as an alternative (or default) to the hard `min_samples=60` cutover in `run_emos_shadow.py` (`src/model/emos_calibration.py`'s `fetch_training_data` / `fetch_training_data_pooled` split).
- Validate against LLBG's own 21-triple backtest window (from #767's investigation) — the blend should measurably close some of the ~0.53 CRPS gap without needing to wait for the full 60-sample bar.
- No regression to cities currently well-served by either pure per-city or pure pooled fits.

## Acceptance criteria

- [ ] Blend formula documented with rationale (weighting scheme, edge cases at n=0 and n≥60)
- [ ] Backtest comparison: hard-cutover CRPS vs blended CRPS, at minimum for LLBG and one healthy comparison city
- [ ] Unit tests for the blend weighting at boundary sample counts
- [ ] No change to cities currently using a pure per-city fit (n≥60 unaffected, or negligibly so)

## Dependencies

Informed by #767 (closed) and #762. Part of epic #70.

P0K5:
[...]

Issue #658, opened 2026-07-08
Define the EMOS-era μ-pipeline layer contract (train/serve parity) before any EMOS promotion


P0K5:
## Context — 2026-07-08 composition audit

The scan-time forecast mean currently passes through four stacked correction layers (DEB weighting → intraday delta → residual bias → obs_bias offset), and `emos_primary` would apply its `a + b·μ` transform **on top of all of them**. But EMOS trains on something else entirely: `fetch_training_data()` fits against the **plain equal-weight average** of the stack models vs METAR truth.

Promoting EMOS as-is creates train/serve skew:
- EMOS's `a` coefficient learns the ensemble's static bias — which the **residual correction** also learns and removes at serve time → static bias removed twice.
- Training μ is equal-weight; serving μ is DEB-weighted + intraday-corrected. The (a, b) fit is for a different variable than the one it transforms. (`FORECAST_STACK_MODELS`'s docstring claims training "must mirror the live envelope's blend exactly" — it currently doesn't.)
- σ side: EMOS fits `c + d·σ` per lead bin, but serving passes a constant σ=2 at every hour. #653's remaining-climb floor mitigates the worst intraday case but is not the design.

## Proposed contract (to be decided on this issue)

1. **EMOS input = plain equal-weight mean of the active stack, per lead bin** — exactly what it trains on. Serve-time code computes that same mean for the EMOS path rather than feeding it `corrected_mu_f`.
2. **EMOS replaces the residual correction** when a city is `emos_shadow`/`emos_primary` — they learn the same static bias; only one may run.
3. **Intraday delta layers on top of μ_cal** (it's a genuine nowcast signal EMOS's fixed-lead training cannot capture), with the obs_bias duplication already removed by #657.
4. **DEB weighting either enters training too, or stays out of the EMOS serving path** — no half-and-half. Recommendation: keep DEB for the legacy path only in v1; revisit weighted-mean EMOS as v2 once weights mature past cold-start.
5. **σ served per lead bin** (nearest bin to minutes_to_settlement), floored by ENVELOPE_SIGMA_CLIMB_FRACTION as today.

## Why now
Blocking #659 (pooled shadow training) is NOT necessary — shadow rows can accumulate under the current wiring — but the contract MUST be settled before any promotion to `emos_primary`, or CRPS shadow comparisons will be measuring a moving target.

## Acceptance criteria
- [ ] Contract decided and documented (docs/design or this issue)
[...]

Issue #448, opened 2026-06-26
[Epic B] Replace fixed FORECAST_STDDEV_F in envelope with conditional sigma


P0K5:
Parent: #445

## What

Plumb `ensemble_sigma_f` through `WeatherState` and consume it in `true_probability_yes()` and the EMOS-shadow path.

## Acceptance criteria

- [ ] WeatherState has new optional field `ensemble_sigma_f: float | None`
- [ ] When set, `true_probability_yes` uses it instead of `FORECAST_STDDEV_F`
- [ ] When `None` (e.g., GEFS unavailable), legacy fixed-sigma fallback
- [ ] EMOS-shadow path picks up the new sigma
- [ ] Unit tests covering both code paths

## Technical notes

- Strict feature-flag rollout: `USE_ENSEMBLE_SIGMA=false` by default; promote per station via the existing promotion gate

ClaudioSenior:
Starting work — plan:
1. Add `WeatherState.ensemble_sigma_f: float | None = None` field (envelope.py)
2. Resolve `USE_ENSEMBLE_SIGMA` feature flag (default off) through `CONFIG_DEFAULTS` + `_CONFIG_META` + live-read, same pattern as `DEB_ENABLED`
3. Consume it in `true_probability_yes()` (replaces `forecast_stddev` before the climb-fraction floor) and in the EMOS-shadow/primary serving path (`resolve_sigma_raw()` feeding `emos_serving_mu`'s `sigma_raw`) — both fall back to legacy `FORECAST_STDDEV_F` unchanged when the flag is off or `ensemble_sigma_f` is `None`
4. Unit tests for both code paths in `test_envelope.py` and `test_emos_mode.py`

Note: this issue only wires the field + consumption path (WeatherState field, `true_probability_yes`, EMOS-shadow). Actually populating `ensemble_sigma_f` from GEFS at scan time (the producer/builder wiring) is scoped to the follow-up issue #449/#665 per the Tech Lead PM's assignment.

ClaudioSenior:
PR #692 submitted for review — https://github.com/P0K5/MeteoEdge/pull/692

[...]

PR #839, opened 2026-07-24
fix(#799): couple EMOS train/serve sigma_source under USE_ENSEMBLE_SIGMA

### LINKED ISSUE #799: EMOS: switch on ensemble sigma (USE_ENSEMBLE_SIGMA) and retrain, coupling train and serve
## Scope correction (2026-07-24 — see docs/REMEDIATION_PLAN.md, PR #827)

This issue's original body (preserved below) proposed reparametrizing the
**fixed**-sigma path (`sigma_source='fixed'`) to drop the unidentifiable `d`
coefficient while staying on the constant `FORECAST_STDDEV_F=2.0`.

The remediation plan (M2, "highest leverage") supersedes that scope. The
actual fix is to **switch on the ensemble sigma predictor that already
exists and is unused**:

- `sigma_f` is already populated from real ensemble spread: 3,546 `open_meteo`
  rows, 2,717 `gefs` rows. `USE_ENSEMBLE_SIGMA` is `False` in production —
  the signal is captured and discarded.
- Flip `USE_ENSEMBLE_SIGMA=True` and retrain EMOS so `sigma_raw` is the real
  per-row ensemble spread instead of a constant. This is what makes `c`/`d`
  identifiable in the first place — a non-constant predictor, not a
  reparametrization of the constant one.
- **Main review risk: train/serve coupling.** The training data used to fit
  `(a,b,c,d)` and the data path serving live/shadow predictions MUST use the
  same `sigma_source`. Decoupling them (training on ensemble sigma, serving
  on fixed, or vice versa) reproduces the **#658** train/serve skew.

## Acceptance criteria (revised)

- [ ] `USE_ENSEMBLE_SIGMA=True` in the config path used by both training and
      serving — verify there is exactly one source of truth for
      `sigma_source`, not independently-set flags on each side.
- [ ] EMOS retrained with `sigma_raw` sourced from `sigma_f` (ensemble
      spread) rather than the `FORECAST_STDDEV_F` constant.
- [ ] `c`/`d` no longer walk to an arbitrary/floored value as an artifact of
      a collinear fit on a constant predictor — confirm against a couple of
      pooled groups (e.g. `midlat_c`, `tropics_c`) that previously floored.
- [ ] No train/serve skew introduced: add a test or check asserting training
      and serving read the same `sigma_source` config.
- [ ] Document before/after `sigma_cal` behavior for at least one city whose
      fit changes materially.

## Dependencies

Part of epic #70. M2 in `docs/REMEDIATION_PLAN.md`. Related: #798 (partial
pooling, same EMOS fit path), #823 (promotion bars — should not be trusted
until #822's skill test passes), #824 (ECMWF sigma capture — a further
sigma-source improvement, not blocking this), #658 (train/serve parity —
the failure mode to avoid).

---

### Background

Follow-up from #767 (Tel Aviv/LLBG diagnosis, accepted) — separate,
fleet-wide, lower priority; **not** the cause of LLBG's own regression (that
was the pooled mu-correction, see #767).

Production trains and serves EMOS exclusively under `sigma_source='fixed'`
(`bot_config.EMOS_SIGMA_SOURCE='fixed'`, `USE_ENSEMBLE_SIGMA=False`; zero
`sigma_source='ensemble'` rows exist in `emos_calibration`). Under `'fixed'`,
`sigma_raw` is the *same constant* (`FORECAST_STDDEV_F=2.0`) for every
training triple, so `sigma_cal = c + d·sigma_raw = c + d·2.0` has infinitely
many `(c,d)` pairs giving an identical fit — `c` and `d` are
collinear/unidentifiable along that ridge. `L-BFGS-B`'s path from
`x0=(0,1,0.5,1)` happens to walk to the `d` lower bound (`d=0.001`) for some
pooled groups (`midlat_c`, `tropics_c`) and not others (`us_f` settles at
`d=0.6748`) — this is optimizer-path noise on a flat objective, not a real
dispersion signal, but it *reads* as "sigma correction disabled /
under-dispersed," which is misleading for anyone diagnosing a city's fit (as
#762/#767 initially assumed for LLBG before the real cause was found).

*(Original "Task" section proposed reparametrizing the fixed-sigma fit
in-place — superseded above by switching to the ensemble sigma source
instead.)*

P0K5:
## Linked issue
Closes #799

## The bug this fixes (beyond the original scope)
Before this change there were **two independent `bot_config` flags** governing ensemble sigma: `USE_ENSEMBLE_SIGMA` (gated the serving-side raw sigma input) and `EMOS_SIGMA_SOURCE` (gated which `emos_calibration` coefficient row was read/written). Worse: the daily retrain (`run_emos_shadow.py::_run_calibration`) never read *either* flag — it relied on `fetch_training_data()`'s own hardcoded default (`sigma_source="ensemble"`) while its `save_coefficients()` call persisted under whatever `EMOS_SIGMA_SOURCE` resolved to (default `"fixed"`). Training silently used one track while the write landed on the other — exactly the #658-style train/serve skew this issue was scoped to avoid.

## What changed
- `Database._active_sigma_source()` (`src/data/db.py`) now derives directly from `USE_ENSEMBLE_SIGMA` — the single source of truth for both which coefficient row and what sigma_raw input get used. `EMOS_SIGMA_SOURCE` is kept as a legacy no-op `bot_config` key (for existing seeded rows / dashboard/API back-compat) — dashboard description updated to say so.
- `run_emos_shadow.py` and `auto_retrain_probability_calibration.py` now resolve `sigma_source` once from `USE_ENSEMBLE_SIGMA` and thread the identical value through every `fetch_training_data`/`save_coefficients`/`upsert_emos_coefficients` call (per-city and pooled paths). Added a `--sigma-source` CLI override to the offline script.
- `CONFIG_DEFAULTS["USE_ENSEMBLE_SIGMA"]` flipped `False → True`.
- New regression tests proving fetch and save always land on the same track (`TestRunnerSigmaSourceCoupling`, `TestSigmaSourceCoupling`, `TestResolveSigmaRawAgreesWithActiveSigmaSource`, `test_emos_sigma_source_key_is_now_a_noop`).

## ⚠️ Interaction risk with PR #836 (#798, partial-pooling blend) — read before merging either
Both PRs modify `scripts/run_emos_shadow.py`'s calibration path substantially. #836 replaces the hard 60-sample cutover with a shrinkage blend; this PR couples the sigma_source used for fetch/save. They're currently based on the same pre-#798/#799 master and don't conflict with each other yet, but **merging one will require rebasing and re-verifying the other** before it merges — don't merge both back-to-back without a rebase+retest pass in between. Recommend merging whichever is reviewed/green first, then rebasing the other on top and re-running its full suite before that one merges.

## ⚠️ Production deploy note — flag prominently
Flipping `CONFIG_DEFAULTS` in code does **not** retroactively change an already-seeded production `bot_config` row (`seed_config()` only seeds unset keys). Required sequencing on the production host:
1. Run `auto_retrain_probability_calibration.py --sigma-source ensemble` (or let the cron `run_emos_shadow.py` pick it up once the flag is set) to produce `sigma_source='ensemble'` calibration rows.
2. Verify shadow CRPS improves/holds.
3. Only then set `USE_ENSEMBLE_SIGMA=true` on the production DB (dashboard PATCH or `db.set_config`) — do not rely on the code default alone.

## Trade-off flagged for review
`USE_ENSEMBLE_SIGMA` was already (pre-existing, #448 design) wired into the **legacy, non-EMOS** `true_probability_yes` sigma input too. Flipping it changes live probability computation for any city still in legacy mode, not just EMOS-shadow cities. That's existing, already-reviewed behavior — not new in this PR — but flipping the flag activates it, worth a second look before the production `set_config` step.

## How to test
Full suite: 2593 passed, 9 skipped, 0 failed. `flake8 src/` clean.

## How to deploy
- [ ] **Config / env change** — see the 3-step production sequencing above; do not just deploy code and flip the flag in one step.
- [ ] **Post-deploy validation** — confirm `emos_calibration` rows are being written with `sigma_source='ensemble'` after step 1, and shadow CRPS before flipping the live flag.

---
_Generated by [Claude Code](https://claude.ai/code/session_01XNxbdqVvJAs4iKKobp2bQo)_


P0K5:
Closing — GitHub won't allow reopening after a rebase force-push ("branch was force-pushed or recreated"). Replaced by #846, same content, rebased onto current master (now includes #798/#836's merge; no conflicts, full relevant test set re-run green).

---

Issue #861, opened 2026-07-25
Establish the correct bracket-boundary convention (adjacent brackets both resolve YES)


P0K5:
## Problem

`resolve_outcome()` in `src/scripts/resolve_bracket_outcomes.py` uses `bracket_low <= observed_high <= bracket_high` — inclusive on **both** ends. So a value sitting exactly on the shared edge of two adjacent brackets resolves YES for **both** of them. A station-day has exactly one daily high, so at most one bracket can contain it; two YES brackets is a logical impossibility.

Confirmed in the 2026-07-25 production dry-run: ZGSZ on 2026-06-12 with `observed_high = 86.0` resolved YES for both `84.2–86.0` and `86.0–87.8`.

`src/scripts/settle.py` (`_write_db_settlements`) uses the **same** `lo <= actual <= hi` expression, so whatever the right answer is, both call sites need it — changing one alone would silently diverge the two resolution paths.

## Why this wasn't just fixed in #860

The obvious fix is a half-open interval, but **the production data is consistent with neither form**, so picking one would be guessing:

| ticker | bracket | observed / `actual_high_f` | `settlements.resolved_yes` | `[lo,hi)` predicts | `(lo,hi]` predicts |
|---|---|---|---|---|---|
| `0x13b2981b…` RKSI | 73.4–75.2 | 75.2 | NO | NO ✓ | YES ✗ |
| `0x8b3730f9…` ZGSZ | 84.2–86.0 | 86.0 | NO | NO ✓ | YES ✗ |
| `0x957fb44f…` ZGSZ | 86.0–87.8 | 86.0 | NO | YES ✗ | NO ✓ |

The last two are adjacent brackets on the same station-day with the same recorded high, and `settlements` says **NO to both** — impossible under any interval convention if the high really were 86.0. So at least some of these rows are not METAR-resolved at all (they are `resolution_source='gamma'`, i.e. Polymarket's official outcome, which #644 documented as disagreeing with METAR ~22% of the time). The boundary question is therefore entangled with the gamma-vs-metar question and cannot be settled from this table alone.

## Likely underlying cause worth checking first

[...]

Issue #858, opened 2026-07-25
Validate resolve_bracket_outcomes() directly against settlements, not just bracket_evals overlap


P0K5:
## Context

#850 added `src/scripts/resolve_bracket_outcomes.py`, which resolves bracket outcomes from observed highs independent of `settlements`, plus a correctness check (`cross_check_against_settlements()`) comparing its resolution against `settlements.resolved_yes` on the ticker-overlap between `bracket_evals.*.jsonl` and `settlements`.

First production dry-run (2026-07-25) result: **385 resolved bracket-rows, 35 station-days, but 0 overlap with `settlements`** (`n_overlap=0, n_mismatch=0`). This is not a correctness pass — it's an untested check, since there was nothing to compare.

Root cause (confirmed via `logs/bracket_evals*` listing and `settlements` date range): `bracket_evals.jsonl` only started logging 2026-07-24 (issue #826). `settlements` spans 2026-07-05 → 2026-07-25, almost entirely *before* that. #820 (the evening false-certainty fix) merged 2026-07-24, the same day M0 intentionally throttled live trading to near-zero — so the two populations have almost no time-window overlap by construction: bracket_evals starts right where the old live-trading-heavy settlement history ends. This isn't a bug, but it means the ticker-overlap cross-check may stay empty for a long time (however long until a newly-logged bracket both appears in bracket_evals and settles).

## The fix

`settlements` already carries everything `resolve_bracket_outcomes()`'s core logic needs per row — `station`, `bracket_low`, `bracket_high`, `actual_high_f`, `resolved_yes` — so the correctness check does not need to go through `bracket_evals` at all. Add a direct validation path: for every row in `settlements`, independently recompute the observed daily high from `observations` (station's local calendar day, same `STATION_TZ`-grouping logic already in the module) and compare `bracket_low <= observed_high <= bracket_high` against `settlements.resolved_yes`. This validates the resolver's core correctness against the full 156-row `settlements` population **today**, without waiting on bracket_evals/settlements time-window overlap.

## Acceptance criteria

- A new function (e.g. `cross_check_against_settlements_direct()`, or extend the existing one to accept `settlements`-only mode) that resolves outcomes directly from `settlements` rows' own `station`/`bracket_low`/`bracket_high`/`ts`-derived settlement date, using the same observed-high logic as `resolve_bracket_outcomes()` — not routed through `bracket_evals`.
[...]

Issue #822, opened 2026-07-24
[GO/NO-GO] Market-vs-model skill test: do our bracket probabilities beat the market's implied prices?


P0K5:
## Purpose

This is the **go/no-go test for the entire trading thesis.** Everything else in the roadmap is engineering; this issue answers the only question that matters:

> **Do our bracket probabilities beat the market's implied probabilities at predicting the actual outcome?**

If they do not, there is no edge, and no amount of EMOS/σ/pooling work will create one — we would be optimising a model that is worse than the free public price. We should know this **before** re-enabling live trading, not after.

## Why this is answerable without trading

We do not need P&L to measure skill. For every evaluated bracket we already store:

- **our probability** — `candidates.p_yes_raw` / `scan_decisions.raw_p_yes`
- **the market's implied probability** — `yes_ask` / `no_ask` (or `market_price`)
- **the realised outcome** — `settlements.resolved_yes`

That is exactly the input to a proper scoring rule. P&L is a noisy, sample-starved proxy; a scoring rule uses every evaluated bracket, not just the handful we traded.

## Metric

**Brier Skill Score of our model against the market as baseline:**

```
BS_model  = mean( (p_model  - outcome)^2 )
BS_market = mean( (p_market - outcome)^2 )
BSS       = 1 - (BS_model / BS_market)
```

- `BSS > 0` → we are more accurate than the market. An edge plausibly exists.
- `BSS ≤ 0` → the market is at least as good as us. **No edge. Stop.**

Report alongside: reliability diagram (10 bins), sharpness histogram, and the same decomposition restricted to the **0.00–0.20 model-probability band**, since that is the NO entry zone and where we already know we are miscalibrated (raw_p_yes under-predicts YES by 2–3× there).

## Required exclusions (critical)

- **Exclude `p_yes_raw = 0.0` rows** — these are certainty-shortcut artifacts, not forecasts (see #820). Including them would flatter the model on rows where it wasn't really forecasting. ~14.5% of settled shadow rows overall, up to 53% for west-of-UTC stations.
- Exclude rows where the market price is at the 1¢/99¢ rail (no information).
- De-duplicate: one row per (station, bracket, settlement date) — the same candidate is re-flagged every poll and would otherwise be counted dozens of times.
- Segment results by same-day vs next-day, and by UTC-offset bucket.

## Two passes

[...]

Issue #644, opened 2026-07-07
Settlement books wrong outcomes: METAR truth disagrees with Polymarket resolution (~22% sign-flips); scanner ignores dashboard NO-gate config


P0K5:
## Problem (from 2026-07-07 ground-truth audit)

Verified every filled live trade against Polymarket Gamma resolutions (\`outcomePrices\` with \`closed=true\`). Findings:

1. **Settlement truth is wrong ~22% of the time.** \`settle.py\` runs at 12:00 UTC before markets resolve; the Gamma lookup returns empty (only 20/139 settlement rows ever got \`market_final_price\`), so it silently falls back to METAR-derived truth, which disagrees with Polymarket's official resolution source. 28 of 126 booked outcomes have the wrong **sign**. 52/139 settlement rows reused a stale previous-day high. Once \`settled_at\` is set the row is never re-checked.
   - Net effect: dashboard/risk_state shows **+95 EUR** since Jun 7; on-chain wallet flow is **−120 USDC** (weather net −86.64). The paper profit is a settlement artifact.
   - Example: KMIA 94–95°F Jun 16 — booked **+0.95** (METAR high 93.92 → \"NO won\"), market resolved **YES** → real **−5.00**.

2. **Scanner ignores dashboard NO-gate config.** \`bot_config\` has \`MIN_PRICE_CENTS=75\`, but \`scanner.py\` reads the module constant (env default 60) at import time — July still shows 74c entries. Ground truth: NO bought at 60–69c has a 46% win rate vs ~65% breakeven (−47.70 EUR on 28 trades).

3. Historical rows need repair: booked pnl, \`settlements.resolved_yes\`, and \`risk_state.daily_pnl\` must be recomputed from actual market resolutions.

## Acceptance criteria
- [ ] Live/shadow settlement uses the Gamma resolution as the sole authoritative source; a market that has not definitively resolved is left unsettled and retried on subsequent runs (no premature METAR settle).
- [ ] Overdue unsettled rows from previous days are picked up by every settle run (bounded lookback), and settled PnL is credited to the row's own trade date in \`risk_state\`.
- [ ] Scanner reads \`MIN_EDGE_CENTS\`, \`MAX_EDGE_CENTS\`, \`MIN_PRICE_CENTS\`, \`MAX_CONFIDENCE_YES_FOR_NO\` from live DB config (dashboard edits take effect without restart).
[...]

PR #862, opened 2026-07-25
Gamma-first ground truth for bracket outcome resolution (#860)

### LINKED ISSUE #860: Gamma-first ground truth for bracket outcome resolution (#822 M3 gate)
## Context

#858's settlements-direct correctness check ran in production (2026-07-25) and found **30 mismatches / 109 checked (27.5%)** between `resolve_bracket_outcomes()`'s METAR-derived outcome and `settlements.resolved_yes`.

Root cause traced to `src/scripts/settle.py:296-305`: `settlements.resolved_yes` is **not** purely METAR-derived. It prefers Polymarket's official resolution when the market closed decisively, falling back to METAR only otherwise:

```python
if market_final_price is not None and (market_final_price >= 95 or market_final_price <= 5):
    resolved_yes = market_final_price >= 95
    resolution_source = "gamma"
else:
    resolved_yes = lo <= actual <= hi
    resolution_source = "metar"
```

The comment above it records why (#644): *"the METAR-derived comparison disagreed with the official resolution in ~22% of audited settlements."* Our 27.5% is that same documented divergence, now visible for the first time because #858 built the check that could see it. So the mismatches are **mostly not a defect in #850/#858** — they are two different ground truths, one of which the team already concluded is the less trustworthy proxy.

## Why this matters for M3

#822 is the go/no-go gate for the entire trading thesis, with a decision rule fixed in advance specifically so the result "cannot be rationalised afterwards." Scoring against a ground truth that is wrong ~22% of the time can plausibly flip which side of the `BSS > 0.05` threshold the verdict lands on. The better source is already in the codebase and one function call away. `resolve_bracket_outcomes()` should use the same resolution precedence `settle.py` does, so there is **one** resolution policy in this repo rather than two competing ones.

## Required change

Make `resolve_bracket_outcomes()` Gamma-first with METAR fallback, mirroring `settle.py`'s precedence exactly:
1. Official Polymarket resolution via `src.data.polymarket.fetch_market_final_price()` when decisive (`>= 95` or `<= 5` cents) → `resolution_source='gamma'`.
2. Otherwise the existing observed-daily-high logic → `resolution_source='metar'`.

**Network cost must be bounded by a persistent cache.** A resolved market's outcome never changes once closed, so each ticker is fetched at most once ever. Populate incrementally; by the M3 target (2026-08-22) most of the population is already cached, amortising API load over weeks instead of one burst.

## Acceptance criteria

- Outcome resolution prefers a decisive Gamma resolution over METAR, with identical thresholds/precedence to `settle.py` — no second, divergent policy.
- Every resolved row carries a `resolution_source` (`gamma` | `metar`) and the dry-run report breaks down counts by source.
- A persistent on-disk cache means any given ticker is fetched from Gamma at most once across runs; a cache hit performs no network call.
- An explicit offline mode (`--no-network`) resolves purely from cache + METAR, making the script fully usable with no network and never silently blocking on API calls.
- Network failure on any single ticker degrades to METAR for that ticker and never aborts the run.
- The #858 settlements-direct correctness check reports `settlements.resolution_source` alongside each mismatch, so expected Gamma-vs-METAR divergence is distinguishable from unexplained disagreement.
- After this lands, re-running the dry-run should show the settlements-direct mismatch count drop substantially from 30/109 — remaining mismatches are the ones actually worth investigating.

## Known separate sub-issue (fix here too)

At least one mismatch is not explained by Gamma-vs-METAR: RKSI ticker `0x13b2981b…`, bracket 73.4–75.2, `observed_high = 75.2`, `settlements.actual_high_f = 75.2` (identical), yet resolver says YES and settlements says NO. Boundary inclusivity: the resolver treats `bracket_high` as inclusive on both ends (`lo <= x <= hi`), so a value sitting exactly on the shared edge of two adjacent brackets resolves YES for **both** — visible in the 2026-07-25 report where ZGSZ 2026-06-12 has `observed_high = 86.0` resolving YES for both 84.2–86.0 and 86.0–87.8. Determine the real convention and make it resolve exactly one bracket. Note `settle.py` shares the same `lo <= actual <= hi` expression, so a change here may need mirroring there — call that out rather than silently diverging.

## Technical notes

- Reuse `src.data.polymarket.fetch_market_final_price()` — do not hand-roll a second Gamma client.
- Keep the read-only `mode=ro` DB discipline from #850 (never instantiate `Database()`; its `__init__` runs migrations + a `deb_weight_log` purge).
- Self-gating rules from #850/#858 still apply: no fabricated numbers when data or network is unavailable.

## Test requirements

- Gamma decisive-YES and decisive-NO override METAR (including a case where the two disagree, proving precedence).
- Non-decisive Gamma price (e.g. 50c) falls back to METAR.
- Network failure / `None` return falls back to METAR and does not raise.
- Cache hit performs zero network calls (assert the fetch function is not called).
- Cache persists across runs and round-trips correctly.
- `--no-network` performs zero network calls and still resolves via cache + METAR.
- Boundary case: a value exactly on a shared bracket edge resolves YES for exactly one of the two adjacent brackets.
- All network mocked — no test may make a real API call.

## Dependencies

Builds on #850 and #858 (both merged).

## Board
Registration deferred — this session's GitHub scope blocks all Projects V2 GraphQL except a pinned PR-review set (`bootstrap_session.sh` and the Status mutation both 403). Functionally Ready; will sync once GraphQL access is available.

Complexity: **Complex** — network dependency, new persistent cache, and it defines what ground truth the M3 decision gate is scored against.

P0K5:
## Linked issue
Closes #860

## What changed

#858's settlements-direct check reported **30 mismatches / 109 checked (27.5%)** on its first production run. Root cause traced to `src/scripts/settle.py:296-305`: `settlements.resolved_yes` is not purely METAR-derived — it prefers Polymarket's official on-chain resolution and falls back to METAR only when the market hasn't settled decisively, *because* the METAR comparison booked the wrong outcome in ~22% of audited settlements (#644). So the mismatches were largely two different ground truths, not a defect in #850/#858.

Since #822's M3 gate is the go/no-go decision for the entire trading thesis, it must be scored against what actually happened. This makes `resolve_bracket_outcomes()` use the same precedence `settle.py` does, so the repo has **one** resolution policy rather than two:

- **Gamma-first, METAR fallback.** Reuses `src.data.polymarket.fetch_market_resolution()` (which already encapsulates the `>= 95` / `<= 5` cent decisive thresholds) rather than re-implementing them. Every resolved row carries `resolution_source` (`gamma` | `metar`), and the report breaks the population down by it.
- **Persistent cache bounds the network cost.** A settled market's outcome never changes, so each ticker is fetched at most once ever. Indecisive results are deliberately *not* cached — a market that hasn't settled yet may settle tomorrow, and caching "unknown" forever would permanently freeze it out of the M3 population.
- **`--no-network`** serves from cache only (zero HTTP requests); **`--no-gamma`** reverts to pre-#860 METAR-only, with the report stating loudly that such a run is not M3-suitable.
- **Failure isolation.** A per-ticker fetch error degrades that one ticker to METAR and never aborts the run.
- **Population can only grow.** A Gamma-resolved row survives even with no observation on record for that station-day — an official outcome doesn't need METAR to corroborate it.
- **#858's check now attributes each mismatch** to the settlements row's own `resolution_source`, separating expected gamma-vs-metar divergence from genuinely unexplained disagreement, and reports `settlements.actual_high_f` next to our recomputed `observed_high` (so "we disagree about the temperature" is distinguishable from "we disagree about the rule").

## Deliberately *not* fixed here: the bracket-boundary bug

`resolve_outcome()`'s inclusive `lo <= x <= hi` makes a value on a shared bracket edge resolve YES for **both** adjacent brackets (ZGSZ 2026-06-12, `observed_high = 86.0`, brackets 84.2–86.0 and 86.0–87.8). The issue asked me to fix this too, but the production data fits **neither** half-open convention — two adjacent brackets with the same recorded high both say NO in `settlements`, which is impossible under any interval rule if that high were real, so the boundary question is entangled with gamma-vs-metar and can't be settled from that table. The colliding brackets are also all Celsius-derived (84.2/86.0/87.8°F = 29/30/31°C), suggesting the real fix may be resolving those markets in native units, not picking an interval. `settle.py` shares the same expression, so a one-sided guess would silently diverge the two paths.

Instead this PR adds `detect_multi_yes_station_days()`, which counts and reports affected station-days so the problem is quantified rather than silent, and I filed **#861** with the evidence table and a suggested approach. Note #860's Gamma-first precedence already removes most of the exposure — a Gamma-resolved bracket never consults the interval logic at all.

## How to test
```
pytest src/tests/test_resolve_bracket_outcomes.py -v
```
80 tests (30 new): gamma overrides METAR in both directions; non-decisive and error paths fall back to METAR without raising; cache hit performs **zero** network calls (asserted via mock); indecisive results retry on a later run; `--no-network` makes zero calls; corrupt/non-dict/unwritable cache degrades without raising; gamma resolves a row with no observation; multi-YES detection; mismatch attribution by source; and a DB *without* the migrated `resolution_source` column still runs the check instead of silently zeroing it.

A module-wide autouse fixture patches the fetch function and redirects the cache to a temp dir, so **no test in the file can make a real Polymarket call or touch `logs/gamma_resolution_cache.json`**.

Full suite: **2746 passed, 9 skipped, 0 failed**. `flake8` clean.

## How to deploy
- [x] **Code only** — no service restart needed; this is a standalone read-only CLI, same as #850/#858. `git pull` is sufficient.
- [x] **Post-deploy validation** — re-run `.venv/bin/python -m src.scripts.resolve_bracket_outcomes` and check the new "Ground truth used" section: the `gamma` share should be non-trivial, and the settlements-direct mismatch count should drop substantially from 30/109, with remaining mismatches attributed by source. First run fetches uncached tickers (slower); subsequent runs are cache-served.
- Note: the first production run writes `logs/gamma_resolution_cache.json`. It is regenerable — deleting it only costs re-fetch time, never correctness.

## Checklist
- [x] Tests pass locally (see above)
- [ ] Project board status updated (In review) — **blocked this session**: Projects V2 GraphQL is restricted to a pinned PR-review operation set (`bootstrap_session.sh` and the Status mutation both return 403). Will be reconciled once GraphQL access is available.
- [x] Docs updated — no env vars, DB schema, API endpoints, or run modes changed; the two new CLI flags are documented in `--help` and the module docstring.

---
_Generated by [Claude Code](https://claude.ai/code/session_017TYLSsDMqeHGwnkxoYKRP1)_


P0K5:
**Tech Lead PM verdict: Approve.** CI green (lint/pytest/gitleaks/AI-review) on head `8ed3ace`, mergeable_state clean. Self-approval is blocked since I authored this, so recording the review here.

Reviewed against #860's acceptance criteria: Gamma-first precedence reuses `fetch_market_resolution()` rather than re-implementing the decisive thresholds (one policy, not two); cache is asserted to make zero network calls on a hit; indecisive results are correctly left uncached so they retry; failure isolation and offline mode both covered by tests; the `resolution_source` probe degrades to `unknown` rather than zeroing the whole check on a pre-migration DB. Full suite run locally before review (2746 passed).

The boundary sub-issue is deliberately deferred to #861 with the evidence, not silently guessed — documented in the PR body and the code comments. Merging.

---

PR #881, opened 2026-07-28
fix(#861): use [lo, hi) interval convention (upper bound exclusive)

### LINKED ISSUE #861: Establish the correct bracket-boundary convention (adjacent brackets both resolve YES)

P0K5:
Closes #861

## What changed

Changed the bracket boundary convention from `[lo, hi]` (inclusive on both ends) to `[lo, hi)` (upper bound exclusive) in both:
- `src/scripts/resolve_bracket_outcomes.py` -- `resolve_outcome()`
- `src/scripts/settle.py` -- `_write_db_settlements()`

## Why

67% of station-days (70 of 105) had boundary collisions -- adjacent Celsius-derived brackets like 29-30C and 30-31C (converted to 84.2-86.0F and 86.0-87.8F) both resolved YES when the observed high sat exactly on the shared edge.

Live Polymarket Gamma API comparison across **2,619 decisive settlements** confirms:
- `[lo, hi)`: 96.9% match with official Gamma resolutions
- `[lo, hi]` (old): 48.8% match
- `(lo, hi]`: 3.1% match

## How to test

1. `python -m pytest src/tests/test_resolve_bracket_outcomes.py -v` -- 129 tests pass
2. `python -m pytest src/tests/test_settle_live.py -v` -- 18 tests pass
3. Re-run dry-run: `python -m src.scripts.resolve_bracket_outcomes --no-network`
   - Boundary collisions: 0 (was 70)
   - Disjoint collisions: 0 (unchanged)
4. New regression test `test_adjacent_brackets_only_one_yes` verifies adjacent brackets no longer both resolve YES

## Before/after dry-run boundary counts

| Metric | Before | After |
|---|---|---|
| Boundary collision station-days | 70 | **0** |
| Disjoint collision station-days | 0 | 0 |


P0K5:
## TechLead-PM review of AI reviewer findings

Both blocking concerns are **false positives**:

1. **settle.py IS changed.** The diff at `src/scripts/settle.py:304` shows:
   ```diff
   -            resolved_yes = lo <= actual <= hi
   +            resolved_yes = lo <= actual < hi
   ```
   Both call sites (`resolve_outcome` and `settle.py`) are migrated together, exactly as the issue requires.

2. **blast radius is verified.** The agent re-ran the actual resolver dry-run against production data and confirmed boundary collisions went 70 → 0. `classify_multi_yes()` correctly reports the new convention. No downstream breakage.

**Verdict**: The AI reviewer misread the settle.py change (it's a 1-line diff, easy to miss). Both blocking issues are resolved. Once Test (pytest) completes green, this is ready for merge.

Manual review confirms: all acceptance criteria met, both call sites changed, regression tests added, dry-run verified.

Issue #893, opened 2026-07-28
EMOS training σ source is order-dependent and `nws` (a per-lead constant) is wrongly eligible on the ensemble track


P0K5:
> **⚠️ Triage update 2026-07-28 — read [the triage comment](../issues/893#issuecomment-5109843865) first.**
> Structural claims confirmed, but severity is overstated and the mechanism is *insertion order*, not alphabetical. Verified on live `meteoedge.db`: **KORD 38%** (not 72%), other US stations **0%** (not ~21%), and KORD's `d` is currently **identifiable** (17 unique σ, stdev 1.20). Reclassified critical → **latent-correctness + hygiene**. KORD GEFS capture gap split out to **#897**. Approach revised (diagnostic now, resolver after M3). **Original report preserved below unchanged for history.**

---

## Problem

`fetch_training_data()` picks the training σ for a date by taking the **first non-NULL `sigma_f` it encounters**:

```python
if date_sigma.get(d) is None:
    raw_sigma = row.get("sigma_f")
```

and `get_forecast_log_by_lead()` returns rows with **`ORDER BY date ASC` and no secondary key**. In practice SQLite returns them alphabetically by model — `ecmwf, gefs, gfs, hrrr, icon, nbm, nws, open_meteo`.

**Nothing specifies this priority.** Which model's σ trains EMOS is an artifact of the scan order.

## Consequence: US stations silently degenerate to a constant

`nws` sigma_f is not a spread — it is a climatological lookup keyed on lead time (`_NWS_SIGMA_BY_LEAD`, documented in #555's sourcing policy):

```python
{3: 1.0, 6: 1.5, 12: 2.0, 18: 2.5, 24: 3.0}
```

`nws` sorts **before** `open_meteo`. So on any date where `gefs` is absent, a US station resolves to that constant and never reaches `open_meteo`'s genuine cross-model stdev.

Observed at lead 24, since the 2026-06-25 reset:

| Station group | gefs dates | Falls through to | Degenerate share |
|---|---|---|---|
| Non-US (LLBG, NZWN, OEJN, RCSS, RJTT, RPLL, SBGR, ZGGG, ZHCC) | 21 / 27 | `open_meteo` (varies) | none |
| US (KATL, KHOU, KLAX, KMIA) | 23 / 29 | **`nws` = 3.0 const** | ~21% of dates |
| **KORD** | **8 / 29** | **`nws` = 3.0 const** | **~72% of dates** |

[...]

Issue #887, opened 2026-07-28
GEFS ensemble spread is 63% below the 1°F floor — decide whether training consumes raw or calibrated sigma before serving is wired


P0K5:
## Problem

`model_forecast_log.sigma_f` carries the **unfloored, uncalibrated** member spread, and a large share of it is implausibly tight for a daily-high forecast:

| Model | rows with σ | σ < 1.0°F | % below floor |
|---|---|---|---|
| **gefs** | 3330 | 2095 | **62.9%** |
| ecmwf | 479 | 219 | 45.7% |
| gfs | 1170 | 232 | 19.8% |
| open_meteo | 4176 | 776 | 18.6% |
| nws | 705 | 0 | 0.0% |

Exact zeros are rare (11 total across all models). The problem is the bulk below the floor — a GEFS ensemble claiming <1°F uncertainty about tomorrow's daily high is the classic under-dispersive signature, not a real forecast of near-certainty.

Per station-day (933 total): **7** have a zero, **34** have *every* model below the floor.

## Why the values are unfloored — and why that is currently one-sided

This is deliberate at capture time. `raw_member_sigma()`:

> Unlike `compute_ensemble_sigma()`, this function performs NO calibration-regression and applies NO `SIGMA_FLOOR_F` clamp — the true (possibly near-zero) spread is returned as-is… the floor must never be baked into what gets persisted to `model_forecast_log`.

The reasoning is sound: `σ_cal = c + d·σ_ens` cannot learn `d` from a σ that was clamped before logging.

The counterpart is that **consumers are supposed to go through `compute_ensemble_sigma()`**, which applies `SIGMA_FLOOR_F = 1.0` plus a historical calibration regression. That function has **no production callers** (see #885). So the floor exists in code and is applied nowhere.

Meanwhile `fetch_training_data()` reads `model_forecast_log.sigma_f` **directly** for the ensemble track — raw, unfloored, uncalibrated. On the 34 all-sub-floor station-days, `σ_cal = c + d·σ` is fitted against a near-degenerate input.

## The decision to make

Three coherent options; the current state is none of them, it is an omission:

1. **Train raw, serve calibrated.** Keep `fetch_training_data()` as-is, and make the serving path call `compute_ensemble_sigma()`. Consistent with the capture-time docstring's intent, but train/serve then see different σ distributions — the #658 skew risk the plan explicitly warns about.
[...]

Issue #885, opened 2026-07-28
[M2 BLOCKER] WeatherState.ensemble_sigma_f is never populated — USE_ENSEMBLE_SIGMA=true has no effect on served probabilities


P0K5:
## Problem

`USE_ENSEMBLE_SIGMA` is **`true`** in production (`bot_config`, set 2026-07-25T07:57), and EMOS has ensemble-track coefficients fitted (27 rows, trained 2026-07-28T12:24). **Neither reaches a served probability.**

`WeatherState.ensemble_sigma_f` has **zero assignment sites** anywhere in `src/` outside tests. Confirmed by AST scan over every `Assign`, `AnnAssign`, and keyword argument:

```
assignment/kwarg sites: NONE -- always None at serving
```

It therefore stays at its dataclass default of `None` forever, and `resolve_sigma_raw()` can never take its ensemble branch:

```python
if use_ensemble_sigma and getattr(state, "ensemble_sigma_f", None) is not None:
    return state.ensemble_sigma_f      # unreachable
return fallback_sigma                  # always FORECAST_STDDEV_F = 2.0
```

This is by design *so far* — `src/model/ensemble_sigma.py` says so explicitly:

> DO NOT import this module from `src/trading/`, `src/strategy/`, or `src/model/envelope.py`. Integration is tracked in **#448 (Week 3)**.

`compute_ensemble_sigma()` — the consumption-time function that applies `SIGMA_FLOOR_F` and the historical calibration regression — has **no production callers**. Only docstring references.

## Why this is an M2 blocker, not a cleanup task

`docs/REMEDIATION_PLAN.md` records M2 as:

> **M2 · Make the model honest** — Switch on ensemble spread and retrain, **coupling train and serve**… **Outcome:** `c`/`d` become identifiable, so EMOS can finally express **sharpness** — the only lever that could legitimately reopen the 74–79¢ band.

Training was switched on. **Serving was not.** M2 delivered training infrastructure only, and the plan currently records it as complete.

Consequences that follow directly:

1. **#874's null result is explained.** Its CRPS delta of −0.0041 "within noise" is exactly what you get comparing two coefficient tracks when *neither* is served.
[...]

Issue #555, opened 2026-07-01
2026-07-01 [P1] EMOS sigma inputs are broken: GEFS spread floored at 1.00°F at logging time; sigma_f NULL for ecmwf/icon/hrrr/nbm


P0K5:
## Context (audit 2026-07-01, finding §1.4)

Two sigma problems that starve EMOS's spread coefficient (`d`) of signal:

1. **GEFS floor**: `SIGMA_FLOOR_F=1.0` is applied inside sigma computation (`src/model/ensemble_sigma.py:31`, applied at :117 and :121) and the floored value is what gets logged — **240 of 345 GEFS rows sit at exactly 1.00°F**. GEFS member spread is the one real ensemble spread we capture, and the floor destroys it before it reaches the log.
2. **NULL sigma_f**: `model_forecast_log.sigma_f` is NULL for ecmwf, icon, hrrr, nbm.

EMOS trains `σ_calibrated = c + d·σ_ensemble` — with a floored/missing σ input, `d` is unlearnable.

## Fix

1. **Log raw sigma**: persist the *unfloored* member-spread sigma in `model_forecast_log`. Apply `SIGMA_FLOOR_F` only at the consumption point (envelope/probability computation), never at capture. Keep the floor for live trading behaviour unchanged.
2. **sigma_f for deterministic channels**: for each of ecmwf/icon/hrrr/nbm, either derive a sigma (e.g. lagged-run spread or documented static per-model dispersion) or record an explicit decision that it stays NULL and EMOS trains per-channel without it. Don't invent numbers silently — a documented NULL beats a fake sigma.
3. Extend `scripts/check_emos_data_quality.py` to flag: % of rows at exactly the floor value, and % NULL sigma per channel.

## Acceptance criteria

- [ ] New GEFS rows log raw (unfloored) sigma; floor applied only at consumption; live probability output unchanged (regression test)
- [ ] Per-channel sigma decision table committed (derive vs NULL, with rationale)
- [ ] `check_emos_data_quality.py` reports floor-saturation and NULL-sigma rates
- [ ] Unit tests for raw-vs-floored paths (`src/tests/` around ensemble_sigma)
- [ ] CI green

## Sequencing

The sooner this lands, the more usable sigma history EMOS has when it first trains (see the min_samples issue). Not blocked by anything.

P0K5:
[...]

PR #896, opened 2026-07-28
fix(sigma): warm GEFS cache once per run; drop pre-#555 clamped training rows

### LINKED ISSUE #894: 329 legacy gefs sigma_f rows are clamped to SIGMA_FLOOR_F (2026-06-29→07-02) and sit inside the EMOS training window
## Problem

`model_forecast_log.sigma_f` for `gefs` contains **329 rows at exactly 1.0** — a point mass against an otherwise continuous distribution. Every neighbouring value has n = 1–2:

```
sigma   n
0.9982  1
1.0     329      <-- point mass
1.0014  2
1.0018  2
```

`1.0` is `SIGMA_FLOOR_F`. These rows were written with the floor applied — the thing #555 explicitly forbids for the persisted column.

## Not a live regression — the window is closed

| Group | n | Logged window |
|---|---|---|
| `exactly_1.0` | 329 | 2026-06-29 18:08 → **2026-07-02 13:14** |
| `other` | 3030 | 2026-06-29 18:04 → 2026-07-28 18:42 (ongoing) |

The clamped rows stop entirely on 07-02 while unclamped capture continues to today — consistent with #555 landing around then and removing the floor from the capture path. The current code is clean: `capture_forecasts.py` has a single `gefs` write path using `raw_member_sigma()` (unfloored by design), and no other writer touches that column.

**So there is nothing to fix in the capture code.** The issue is the residue.

## Why it still matters

The EMOS reset timestamp is **2026-06-25**, so the clamped window sits *inside* the training range: **5 of 27 distinct training dates (18.5%)** at lead 24.

That is load-bearing for #887, which just decided *train on raw σ*, reasoning:

> `σ_cal = c + d·σ_ens` cannot learn `d` from a σ_ens that has already been clamped to a floor before it was logged

For those 5 dates it was. Given ~63% of GEFS spread sits sub-floor, most of the affected dates had a true value below 1.0 flattened to exactly 1.0 — compressing precisely the low end of the range `d` needs to see.

Confirmed to reach the fit for non-US stations, where `gefs` wins the σ-source selection (see #893). For US stations the selection is a separate problem entirely.

## Options

1. **Exclude 2026-06-29→2026-07-02 from ensemble training.** A date-range filter on the ensemble track, with a comment pointing at this window. Cleanest — the window is precisely bounded and will never grow.
2. **Accept and document.** Defensible if the fit proves insensitive; the contamination is one-directional (compresses low σ), biasing `d` toward under-weighting spread — the conservative direction.
3. **Do NOT NULL the rows.** Tempting and worse: `fetch_training_data` falls back to `FORECAST_STDDEV_F = 2.0` on NULL, replacing a wrong-but-close 1.0 with a wrong-and-further 2.0.

Recommend option 1.

## Also worth adding

#892's sigma monitor watches for the below-floor share dropping toward 0% (a capture-side regression). It found this incidentally. A **point mass at exactly `SIGMA_FLOOR_F`** is the shape a clamp actually makes, so an explicit check for that would catch the same class of bug directly.

## Acceptance

- Clamped window excluded from ensemble training, or the decision to accept it recorded.
- `check_emos_data_quality.py` gains a point-mass-at-floor check.

## Board

Registration deferred — no Projects V2 GraphQL access in this session.

Complexity: **Simple** — bounded historical data issue with a known fix.

Refs: #555, #887, #892, #893, #824, #822.

### LINKED ISSUE #895: KORD loses GEFS on ~72% of dates because it sorts first in STATIONS and eats the cold-start 90s timeout
## Problem

`KORD` fails its GEFS capture **98 times** in the current logs. Every other station fails 8–10 times:

```
98 [capture] KORD
10 [capture] SBGR
10 [capture] RKPK
10 [capture] NZWN
10 [capture] LTAC
10 [capture] EGLC
 8 [capture] ZSPD
 ...
```

A 10× outlier, with a single uniform cause — never an HTTP error, never an empty member set:

```
2026-07-26T07:03:35 WARNING [capture] KORD gefs unavailable (lead=18h): gefs exceeded 90s timeout
2026-07-26T13:03:35 WARNING [capture] KORD gefs unavailable (lead=24h): gefs exceeded 90s timeout
2026-07-26T19:03:34 WARNING [capture] KORD gefs unavailable (lead=6h):  gefs exceeded 90s timeout
2026-07-27T07:03:35 WARNING [capture] KORD gefs unavailable (lead=18h): gefs exceeded 90s timeout
...
```

## Root cause: KORD is index 0 in `STATIONS`

```python
STATIONS = [
    ("KORD", 41.9742, -87.9073, "Chicago", "KORD", "F", "America/Chicago"),
    ("KMIA", ...),
```

`capture_forecasts.py` iterates `for station, lat, lon, city, *_ in STATIONS:`, so **KORD pays the cold-start cost on every capture run** — DNS, TLS handshake, GRIB index fetch, whatever the GEFS path warms — and exceeds `CAPTURE_FETCH_TIMEOUT_SECONDS = 90`. Every station after it runs against a warm path.

Three independent facts confirm it is positional, not station-specific:

1. **Magnitude is a step, not a gradient** — 98 vs a tight 8–10 cluster. Consistent with one station bearing a fixed setup cost.
2. **Cause is uniform** — 100% timeouts. A bad grid point would produce varied errors.
3. **Failures cluster at capture-run start times** (07:03, 13:03, 19:03 daily); the `lead=Xh` label just tracks the hour. It is the *first fetch of each run* that dies.

Nothing to do with Chicago's coordinates, and nothing to do with KORD having been the live-traded station — `capture_forecasts.py` has no live/shadow awareness at all.

## Why this is not just a capture nuisance

This is the **upstream cause of #893**. KORD loses GEFS on ~72% of dates at lead 24 (8 rows of 29, against 21–23 for every other station), so its EMOS training σ falls through to `nws` — a climatological per-lead constant (3.0 at 24h). With σ constant, `c` and `d` enter `σ_cal = c + d·σ_raw` only as `c + 3d`: a flat ridge, so **`d` is unidentifiable**.

**The most-traded station — the one `docs/REMEDIATION_PLAN.md`'s entire diagnosis is built around — has degenerate σ calibration because it sorts first in a Python list.**

Fixing #893's fallthrough priority alone will not help KORD: if GEFS never arrives, there is no ensemble spread to prioritise. This issue is a prerequisite.

## Likely compounding effect

Per `_call_with_timeout`'s docstring:

> The worker thread is **NOT forcibly killed** on timeout (Python has no safe way to do that); it is abandoned to finish or error out on its own in the background.

So KORD's abandoned GEFS fetch keeps running while subsequent stations fetch, competing for bandwidth and connections. That may well account for part of the residual 8–10 failures on other stations — meaning fixing KORD could reduce failures fleet-wide, not just its own.

## Fixes, cheapest first

1. **Raise the timeout.** `CAPTURE_FETCH_TIMEOUT_SECONDS` is already env-configurable (`os.getenv("CAPTURE_FETCH_TIMEOUT_SECONDS", "90")`) — a `.env` change with **no code change at all**. Given every other station clears comfortably, the cold start is the only thing near the limit. Likely clears most failures fleet-wide.
2. **Warm up before the loop** — one throwaway fetch, or a session/connection reused across stations, so no station pays setup cost.
3. **Retry once on timeout** — the retry hits a warm path and would very likely succeed.
4. **Do NOT fix by reordering `STATIONS`.** It only moves the problem to whichever station lands first, and silently changes which station has degenerate calibration.

Option 1 is worth trying immediately since it is reversible and requires no deploy; 2 or 3 is the durable fix.

## Acceptance

- KORD's GEFS capture rate matches the fleet (~21–23 of 29 dates at lead 24, ideally better).
- No station shows an order-of-magnitude excess of capture timeouts.
- Re-check `check_emos_data_quality.py`: KORD's resolved training σ should come from `gefs` (varying), not `nws` (constant).

## Note on M3 sequencing

This is capture-side only — it does **not** change served probabilities, so it is safe to ship during M3 collection, unlike #885/#893. It improves the data future retrains see; it does not retroactively repair the ~72% of KORD dates already trained against a constant.

## Board

Registration deferred — no Projects V2 GraphQL access in this session.

Complexity: **Simple** — root cause confirmed, several bounded fixes, one requiring no code change.

Refs: #893, #717 (the timeout mechanism), #555, #887, #822.

P0K5:
## Linked issues
Closes #895. Closes #894. Prerequisite for #893.

## #895 — KORD lost GEFS on ~72% of dates because it sorts first in `STATIONS`

`fetch_gefs_ensemble` downloads **31 member GRIB files**, cached by `(model, var, cycle, fxx, member)` — **not** by station. So on a fresh GEFS cycle the *first* station pays for all 31 downloads while stations 2–30 read from the warm cache. `STATIONS[0]` is KORD, which blew the 90s per-fetch timeout on nearly every run:

```
98 [capture] KORD
10 [capture] SBGR
10 [capture] RKPK
 8 [capture] ZSPD
 ...
```

100% of them `exceeded 90s timeout` — never an HTTP error, never an empty member set — clustered at capture-run start times (07:03, 13:03, 19:03), with the `lead=Xh` label just tracking the hour. It is the **first fetch of each run** that dies. Positional, not geographic, and unrelated to KORD having traded live (`capture_forecasts.py` has no live/shadow awareness at all).

**The downstream cost was not a missing row.** With no `gefs` row, EMOS training for KORD fell through to the next model with a non-NULL `sigma_f` — `nws`, whose sigma is a per-lead climatological **constant** (3.0 at 24h). With σ constant, `c` and `d` in `σ_cal = c + d·σ_raw` collapse onto a flat ridge and **`d` is unidentifiable** (#893).

> The most-traded station — the one `docs/REMEDIATION_PLAN.md`'s entire diagnosis is built around — had degenerate σ calibration because of its index in a Python list.

### Fix

`_warm_gefs_cache()` primes the shared per-cycle cache once before the station loop, so no station pays the cold start. It gets its own larger budget (`CAPTURE_GEFS_WARMUP_TIMEOUT_SECONDS`, default 600) because 31 member downloads is structurally ~31× a single-endpoint per-station fetch — holding it to the same timeout is precisely what caused the bug. `_call_with_timeout` grows an optional `_timeout` override to support that.

Two deliberate choices:
- **Not fixed by reordering `STATIONS`.** That only moves the penalty to whichever station lands first, and silently changes *which* station ends up degenerate.
- **Warm-up failure is non-fatal.** Every station still attempts its own fetch, so a failed warm-up restores exactly the previous behaviour rather than skipping GEFS entirely.

## #894 — 329 legacy clamped rows reached the fit

A **2026-06-29 → 07-02** window was persisted with `SIGMA_FLOOR_F` already applied, before #555 moved the floor from capture-time to consumption-time. 329 rows sit at exactly `1.0` against an otherwise continuous distribution — every neighbouring value has n=1–2:

```
0.9982  1
1.0     329   <-- point mass
1.0014  2
```

That spans **5 of 27** distinct training dates at lead 24, and breaks the premise #887's train-on-raw decision rests on: `σ_cal = c + d·σ_ens` cannot learn `d` from a σ clamped before it was logged. Since ~63% of GEFS spread is sub-floor, most affected dates had a true value *below* 1.0 flattened to exactly 1.0 — compressing precisely the low end of the range `d` needs to see.

### Fix

`fetch_training_data` drops the whole **date** on the ensemble track when it sees the exact `(model, date-range, value)` clamp signature.

**Dropping rather than un-setting matters:** leaving σ as `None` falls through to `default_sigma` (`FORECAST_STDDEV_F = 2.0`), which is *further* from the true sub-floor value than the clamped 1.0 it replaced. A dedicated sentinel keeps the date out of the fit entirely.

Only the exact signature is rejected. Untouched: an unclamped row inside the window, a `1.0` outside it, a non-`gefs` `1.0` (`nws`'s 3h climatological entry *is* 1.0), and the `fixed` track. The window is closed and cannot grow — capture has written raw unfloored σ via `raw_member_sigma()` since #555, through a single `gefs` write path.

## How to test

```bash
python -m pytest src/tests/test_capture_forecasts.py src/tests/test_emos_calibration.py -v
python -m flake8 src/
```

Full suite: **2912 passed, 9 skipped**; flake8 clean. 13 new tests covering: warm-up ordering (must precede the station loop, or the first station still pays the cold start), the warm-up budget exceeding the per-fetch budget, non-fatal failure, dry-run skip, first-station coordinates, explicit-vs-default timeout — and for #894, each of the four "must be kept" cases alongside the exclusion itself.

## How to deploy

- [x] **Code only** — `git pull` + restart. `capture_forecasts.py` runs under `meteoedge-capture-forecasts.timer` (oneshot), so it picks up the change on its next scheduled run with no restart needed. `emos_calibration.py` is used by retrain paths, not the live scan loop.
- [x] **Post-deploy validation** — after the next capture run:
  ```bash
  grep "gefs cache warmed" logs/capture_forecasts.log | tail -3
  grep -c "KORD gefs unavailable" logs/capture_forecasts.log
  ```
  Expect a `gefs cache warmed: N members in Xs` line and KORD's failure count to stop growing. On the next retrain, `[emos] ... dropped N date(s) with pre-#555 clamped gefs sigma_f` confirms #894.
- **Safe during M3 collection** — capture-side and training-side only; neither changes served probabilities. Unlike #885/#893.

## Note on scope

This does **not** retroactively repair the ~72% of KORD dates already trained against a constant, nor does it fix the fallthrough priority that made `nws` eligible as a σ source in the first place — that is #893, correctly held until after M3.

## Checklist
- [x] PR body includes `Closes #N` linking the issues
- [ ] Project board status updated (In review) — **blocked**: no Projects V2 GraphQL access in this session
- [x] Docs updated — rationale is in the module/function docstrings; no env-var, schema, or API changes beyond the new optional `CAPTURE_GEFS_WARMUP_TIMEOUT_SECONDS`

---
_Generated by [Claude Code](https://claude.ai/code/session_01GQDvbped6jLKS3ZhKAg6bz)_


P0K5:
## Correction to this PR's stated impact — the code is unchanged and still correct

Triaging #893 turned up something that invalidates part of this PR's description. Flagging before merge; **no code change needed**, but the claimed downstream impact is wrong.

`fetch_training_data` applies a **regime filter** before σ selection:

```python
if regime is not None:
    forecast_rows = [r for r in forecast_rows if r.get("model") in regime]
```

and production passes `FORECAST_STACK = baseline` → `frozenset({"nws", "open_meteo"})`. **`gefs` is not in that set.**

### What that means for each half

**#895 (GEFS warm-up)** — the fix is correct and worth merging, but this claim in the description is wrong:

> With no `gefs` row, EMOS training for KORD fell through to the next model with a non-NULL `sigma_f` — `nws` … The most-traded station had degenerate σ calibration because of its index in a Python list.

`gefs` was never a training σ candidate under the baseline stack, so the capture gap is **not** what causes KORD's degenerate σ. Per #893's corrected triage, `nws` wins unconditionally for all 5 US stations on 100% of dates, independent of whether `gefs` was captured.

The warm-up still earns its place — it fixes real capture failures (98 for KORD), and GEFS σ matters for the `full` stack and for `state.ensemble_sigma_f` if #885 wires serving, which reads GEFS directly rather than through the training regime. It just doesn't fix KORD's training σ.

**#894 (clamped rows)** — correct, but **inert under the baseline stack**: those `gefs` rows are filtered out before σ selection ever sees them. It becomes live if `FORECAST_STACK` moves to `full`. Worth keeping as a latent guard; not an active contamination fix today.

### Recommendation

Merge as-is. Both changes are correct, tested, and safe during M3 collection. Only the framing needs the above caveat, which this comment records.

The real fix for KORD's degenerate σ is #893, correctly held until after M3.

---

Issue #917, opened 2026-07-31
[P0] Probability mass not conserved across bracket ladder — SUM(raw_p_yes) median 0.70, ~0.53 on US F ladders; likely inclusive/exclusive bracket_high mismatch biasing NO edge upward


P0K5:
## P0 — probability mass is not conserved across the bracket ladder

Found while auditing the 2026-07-30 health report. This is a **model-correctness** issue, not an observability one, and it plausibly explains several long-standing anomalies at once.

For a given station and poll, the brackets partition the outcome space, so `SUM(raw_p_yes)` over all brackets in that station-poll must be ≈ **1.0**. It is not.

```sql
SELECT poll_ts, station, SUM(raw_p_yes)
FROM scan_decisions WHERE poll_ts >= '2026-07-28' GROUP BY 1,2;
```

Over 114 station-polls: **min 0.061, median 0.704, max 1.156.**

| SUM(raw_p_yes) | station-polls |
|---|---|
| < 0.5 | 37 |
| 0.5 – 0.9 | 25 |
| 0.9 – 1.1 | 50 |
| > 1.1 | 2 |

## Leading hypothesis: inclusive-vs-exclusive `bracket_high` on US (°F) ladders

`p_normal_between` (`src/model/envelope.py:88-103`) integrates `[low, high)` — lower inclusive, upper **exclusive** — a convention deliberately set by #861/#881 and revisited in #888/#891.

That convention is correct for the Celsius-derived ladders, whose edges are genuinely contiguous (RKSI: `75.2–77.0`, `77.0–78.8`, `78.8–91.4` — zero gap).

It appears to be **wrong for the US °F ladders**, where the stored edges are *not* contiguous (KLAX, one market date):

```
(-50, 73)  (74, 75)  (76, 77)  (78, 79)  (80, 81)  (82, 83)
(84, 85)   (86, 87)  (88, 89)  (90, 91)  (92, 200)
```

Brackets are 1.0°F wide with 1.0°F gaps between them. A Polymarket bracket labelled "74–75" means the integer-inclusive set {74, 75}, i.e. the real interval **[74, 76)** — but it is integrated as [74, 75), discarding the [75, 76) half. Uncensored °F station-polls show a median `SUM(raw_p_yes)` of **0.530**.

## Why this is P0

[...]

Issue #914, opened 2026-07-30
Daily health report: poll counter reads the wrong table (false '5/288' + '298 min gap'), verdict ignores gaps, and WARN attribution is substring-matched over the whole report


P0K5:
## Symptom

Health report 2026-07-30 contains a self-contradiction:

```
Bot Pulse
  Last poll:   2026-07-30 13:56 UTC (3 min ago)
  Polls 24h:   5/288
  [WARN] Max gap:    298 min (threshold: 20)
  Status:      [OK] Healthy
```

5 polls out of an expected 288, a 5-hour gap, and "Healthy".

## Defect 1 — the poll counter measures the wrong table

`_build_bot_health` (daily_health_report.py:136-187) derives poll count and gap from `MAX/DISTINCT scan_decisions.poll_ts`. But `scan_decisions` only receives rows when brackets are actually evaluated, which happens a handful of times a day.

The bot itself is polling normally, every ~5.5 minutes, all day:

```
2026-07-30T16:54:46 INFO [__main__] === Poll [LIVE] at 2026-07-30T15:54:46+00:00 ===
2026-07-30T17:00:21 INFO [__main__] === Poll [LIVE] at 2026-07-30T16:00:21+00:00 ===
2026-07-30T17:05:48 INFO [__main__] === Poll [LIVE] at 2026-07-30T16:05:48+00:00 ===
… (continuous)
```

So `5/288` and `Max gap 298 min` are **false alarms**. The `/288` denominator assumes one `scan_decisions` write per 5-minute poll, which was never the contract.

Fix: count polls from an actual poll heartbeat (a `poll_runs` table, or a heartbeat row written unconditionally at the top of `poll_once`), not from a table that is conditionally written. Keep the `scan_decisions` cadence as a *separate* line ("evaluated polls"), since it is genuinely useful — but do not label it "Polls 24h".

## Defect 2 — the verdict ignores the gap it just printed

`Status:` (lines 175-180) is computed **only** from `gap_min` since the last poll. The 24h poll count and `max_gap` are printed but never feed the status or the final verdict. A bot that polled 5 times in 24h and once 3 minutes ago reports "Healthy".

Fix: fold poll-count shortfall and `max_gap` into the status/verdict.

## Defect 3 — verdict false attribution

`_build_verdict`, daily_health_report.py:483:

```python
if "p_yes=0.0 artifact:" in all_text and "[WARN]" in all_text:
    warn_issues.append("artifact rate high")
```

[...]

Issue #913, opened 2026-07-30
scan_decisions upsert failures are still silently swallowed — no alerting, no health-report surfacing, no fail-loud (live gap through M3 gate window)


P0K5:
## What happened

`scan_decisions` writes failed for **two consecutive days** and nothing surfaced it.

```
2026-07-29T13:59:57 WARNING [__main__] [scan_decisions] upsert failed for 0x5005ffa0…:
    Database.upsert_scan_decision() got an unexpected keyword argument 'direction'
```

Failure counts from `logs/bot.log`:

| date | rejected upserts |
|---|---|
| 2026-07-24 | 351 |
| 2026-07-28 | 20,463 |
| 2026-07-29 | 32,465 |
| **total** | **53,279** |

The DB shows the hole directly — 2026-07-29 has **zero** `scan_decisions` rows:

```sql
SELECT substr(poll_ts,1,10), COUNT(DISTINCT poll_ts) FROM scan_decisions
WHERE poll_ts >= '2026-07-22' GROUP BY 1;
-- 07-22:3  07-23:3  07-24:3  07-25:3  07-26:3  07-27:3  07-28:6  (07-29 absent)  07-30:8
```

### 1. Deploy lag — the fix was merged but not running

`fa7edf7` / PR #899 ("add missing 'direction' parameter to upsert_scan_decision") merged 2026-07-29, yet the running bot kept rejecting upserts through 13:59 UTC on 07-29 and only recovered on 07-30. The gap between "merged" and "the process actually restarted on the new code" is unmeasured and unenforced. Related: #791 (PR-template "How to deploy" enforcement).

### 2. No alarm on a write-failure storm

32,465 rejected writes in one day produced no alert, no dashboard signal, and — critically — the 2026-07-30 daily health report rendered **`Status: [OK] Healthy`** the morning after. Every WARNING was swallowed at the call site and never aggregated.

## Requested fix

1. **Alert on persistence failures.** Count `[scan_decisions] upsert failed` (and equivalent write-path failures) per poll; alert on any sustained non-zero rate. A silent-failure counter belongs next to the guardrail counters, not only in the log.
2. **Surface it in the daily health report.** Add an "Errors / silent failures (24h)" section that aggregates WARNING/ERROR counts by message class, and make a non-zero write-failure count a `[CRIT]` in the verdict.
[...]

Issue #911, opened 2026-07-30
scan_decisions.forecast_high does not record the mean the model actually served (mu resolved via corrected_mu_f/deb_mu_f/secondary), making valid evaluations look forecast-less


P0K5:
## Symptom

Health report 2026-07-30: `p_yes=0.0 artifact: 45.9% [WARN] (pre-fix 17.8%)` — worse than the pre-fix baseline.

The WARN is real, but the cause is not a model regression. **74% of the p_yes=0.0 rows have `forecast_high IS NULL`** — the scanner is evaluating brackets, assigning `p_yes = 0.0`, and persisting a gate verdict when it has no forecast at all.

```sql
SELECT SUM(forecast_high IS NULL), COUNT(*) FROM scan_decisions
WHERE poll_ts >= '2026-07-24' AND raw_p_yes = 0.0;
-- 1085 / 1474
```

## The failure is concentrated in a specific daily window

`forecast_high IS NULL` is not scattered — it clusters in the ~10:55 and ~11:42 UTC polls, every day:

| poll_ts | n | forecast_high NULL | raw_p_yes=0.0 |
|---|---|---|---|
| 2026-07-28T10:54 | 11 | **11** | 10 |
| 2026-07-28T11:40 | 286 | **242** | 206 |
| 2026-07-28T12:23 | 283 | 0 | 60 |
| 2026-07-30T10:57 | 11 | **11** | 7 |
| 2026-07-30T11:42 | 286 | **242** | 215 |
| 2026-07-30T16:34 | 176 | 0 | 48 |

Uniformly 66 NULL rows per station across ~23 international stations (EFHK, EGLC, EPWA, LFPB, LIMC, LLBG, LTAC, LTFM, MPMG, NZWN, OEJN, RCSS, RJTT, RKPK, RKSI, …). All `is_next_day = 0`.

Sample row from the 2026-07-30T11:42 poll:

```
station=RKSI  gate_verdict=below_min_edge  forecast_high=NULL
raw_p_yes=0.0  capped_p_yes=0.05  yes_ask=1  no_ask=99
```

## Why this matters beyond the metric

[...]

PR #918, opened 2026-07-31
fix(#910): derive rail metric thresholds from MODEL_PROB_CAP

### LINKED ISSUE #910: Health report rail metric is vacuously green: threshold (0.02/0.98) lies outside the clamp range (0.05/0.95) — true rail is 62.4%, not 0.0%
## Symptom

The 2026-07-30 daily health report prints:

```
Rail (0-2% / 98-100%): 0.0% [OK] (pre-fix 62.6%)
```

This reads as "the M3 rail-concentration remediation worked". It did not — the metric is **mathematically incapable of firing**.

## Root cause

`src/scripts/daily_health_report.py:369` counts rows where
`capped_p_yes <= 0.02 OR capped_p_yes >= 0.98`.

But `capped_p_yes` is the **post-clamp** value, and the clamp floor/ceiling are 0.05 / 0.95:

```sql
SELECT MIN(capped_p_yes), MAX(capped_p_yes) FROM scan_decisions WHERE poll_ts >= '2026-07-24';
-- (0.05, 0.95)
```

No row can ever satisfy `<= 0.02` or `>= 0.98`. The metric returns 0.0% unconditionally and stamps `[OK]`.

## What the real number is

Counting rows pinned **at the clamp boundary** (the thing the rail metric is trying to measure):

```sql
SELECT SUM(capped_p_yes <= 0.0501 OR capped_p_yes >= 0.9499), COUNT(*)
FROM scan_decisions WHERE poll_ts >= '2026-07-24';
-- 1865 / 2989 = 62.4%
```

**62.4% — versus the "pre-fix 62.6%" baseline the report cites as the thing we improved on.** Rail concentration has not moved at all; we have been reading a green light on a metric that cannot go red.

Per-poll (since 07-28), rows at the 0.05 floor:

| poll_ts | n | at floor |
|---|---|---|
| 2026-07-28T11:40 | 286 | 223 |
| 2026-07-30T11:42 | 286 | 233 |
| 2026-07-30T16:34 | 176 | 113 |
| 2026-07-30T14:59 | 88 | 48 |

## Fix

1. Measure rail against the clamp boundary, not hardcoded 0.02/0.98 — read the actual clamp constants so the metric tracks them if they change.
2. Better: measure `raw_p_yes` rail (pre-clamp), which is what the M3 remediation was actually about; `capped_p_yes` rail is by construction a function of the clamp.
3. Add a unit test that fails if the rail threshold sits outside the clamp range (the class of bug that made this vacuous).
4. Re-baseline the "(pre-fix 62.6%)" annotation against whichever definition is chosen.

## Impact

M3 go/no-go is partly gated on this indicator. It has been reporting success for a remediation that has not landed.

P0K5:
## Summary

Fixes the health report rail metric which was vacuously green due to hardcoded thresholds (0.02/0.98) lying outside the actual clamp range (0.05/0.95). The metric now derives thresholds from `MODEL_PROB_CAP` dynamically so it correctly reports the true ~62.4% rail concentration as [WARN].

Closes #910

## Changes

- Add `MODEL_PROB_CAP` import from `src.config`
- Derive rail thresholds dynamically: `lower = 1.0 - MODEL_PROB_CAP`, `upper = MODEL_PROB_CAP`
- Update SQL query to use parameterized thresholds
- Update output display to derive label from `rail_upper` directly, not from `100 - rail_lower`
- Remove stale pre-fix baseline annotations
- Add regression test: actual SQL execution against rows with `capped_p_yes` at clamp floor

## Test plan

- [x] All existing tests pass (29 tests in test_daily_health_report.py)
- [x] Regression test: rows with capped_p_yes at clamp floor are counted as rail
- [x] Regression test: fails with old hardcoded thresholds (0.02/0.98), passes with fix
- [x] Test with 62.4% rail concentration correctly shows [WARN]
- [x] Test with 15% concentration correctly shows [OK]

## How to deploy

The health report runs under the `meteoedge-health-report.timer` systemd unit, which fires daily at 14:00 UTC. Since this change affects only the Python script (not the systemd unit file itself), the next scheduled 14:00 UTC run will automatically pick up the changes. No unit reload is required.


P0K5:
## Review: **request changes**

The production change is essentially right — deriving from `MODEL_PROB_CAP`, reusing scanner's `round(..., 10)` convention, parameterised SQL, and the #917 comment are all exactly what I asked for. Nice work on the core fix. Four things block merge, and the most important is that the issue's headline acceptance criterion isn't actually covered by a test.

### `src/scripts/daily_health_report.py`

**[blocker] The displayed label re-creates the very bug this PR fixes.**

```python
rail_range_pct = round(rail_lower * 100, 1)
lines.append(f"  Rail (0-{rail_range_pct}% / {100-rail_range_pct}%-100%): …")
```

The SQL uses `rail_upper`, but the label derives the upper bound as `100 - rail_range_pct`. Two independent sources for one number, held together by the symmetry assumption in your comment. Today `MODEL_PROB_CAP = 0.95` makes them agree, so nothing looks wrong — which is precisely how the original defect survived: a *displayed* threshold that had quietly diverged from the *queried* one.

Derive the label from `rail_upper` directly, so the printed range can never disagree with the range that was measured:

```python
lines.append(f"  Rail (0-{rail_lower*100:.1f}% / {rail_upper*100:.1f}%-100%): …")
```

**[question] The `(pre-fix 62.6%)` annotation is now measured under a different definition.** That baseline came from the old 0.02/0.98 query; the metric now means something else. Acceptance criterion 4 on #910 asked for it to be re-baselined. Either re-state it against the new definition or drop it — as-is it invites a false apples-to-apples comparison. Same for the stale `# pre-fix was 62.6%` comment on `rail_ok`.

### `src/tests/test_daily_health_report.py`

**[blocker] The primary acceptance criterion is untested.** #910 criterion 3 asks for *"a unit test that fails if the rail threshold sits outside the clamp range"* — the bug class, not the symptom. Neither test covers it. Both assert on percentages that the mock hands back directly, so they'd pass even if the thresholds were 0.02/0.98 again.

What's needed is a test that goes through real data, not a stubbed `COUNT`: insert rows sitting exactly at the clamp floor (`capped_p_yes = 1.0 - MODEL_PROB_CAP`) and assert they're **counted as rail**. That test fails against the old hardcoded 0.02 and passes now. That's the regression guard.

**[blocker] `test_rail_metric_uses_model_prob_cap_thresholds` has a dead fixture and an incoherent mock.**

- You build `db = _mock_db({...})` and then immediately overwrite `db._conn.execute = mock_execute`. The whole dict is dead code — delete it.
- `call_order = [0]` is unused.
- The `else` branch returns `(1000,)` for the **station-days** query too, so the section renders `Station-days: 1000/300 (333%)` at `~142/day`. Unasserted, but a fixture that produces impossible output is a fixture you can't trust when the test later fails for a real reason.

**[blocker] `assert "[WARN]" in text` matches the whole section, not the rail line.** This is the identical anti-pattern I've filed as a defect in `_build_verdict` under #914 — substring-matching a status marker over rendered text, so an unrelated `[WARN]` satisfies the assertion. Your *second* test does it correctly:

```python
rail_line = [line for line in result if "Rail" in line][0]
assert "[OK]" in rail_line
```

Use that form in both.

**[suggestion] `assert "62.4%" in text or "62.3%" in text or "62.5%" in text`** — 624/1000 is exactly 62.4%. A three-way or-assert reads as uncertainty about your own arithmetic and would mask a genuine off-by-a-tenth. Make it exact.

**[nitpick]** Second test stubs `db.get_emos_crps_count`, which `_build_m3_progress` never calls — copy-paste from the EMOS section.

### Governance

**[blocker] The PR body has no "How to deploy" section**, required by `.github/pull_request_template.md`. This one genuinely needs it: the report runs under the `meteoedge-health-report.timer` systemd unit, so state whether the unit needs reloading or whether the next 14:00 UTC firing picks the change up automatically. See #791 — this is the repeatedly-missed step, and I hold myself to it too.

### CI

`Lint`, `gitleaks`, and `AI / NVIDIA NIM review` are green on `e414e40`; `Test (pytest)` is still running. Not the reason for this verdict — the blockers above are — but note I cannot approve until all four are green on the head SHA.

---

Moving #910 back to **In progress**. Re-request review once the blockers are addressed.

P0K5 (commented):
All review feedback addressed:

1. ✓ Label now derives from rail_upper directly (not 100-rail_lower)
2. ✓ Added regression test with actual SQL against real rows - fails with old thresholds, passes with fix
3. ✓ Fixed test fixtures (removed dead dict, unused call_order)
4. ✓ Changed to exact assertions (rail_line for [OK]/[WARN], exact percentage)
5. ✓ Removed stale pre-fix baseline annotations
6. ✓ Added 'How to deploy' section

Ready for review. Please verify and approve.

P0K5:
## Review: **approve**

All four blockers addressed. I independently verified the one that mattered rather than taking the summary on trust.

**The regression test is real.** I reverted the thresholds to the original buggy `0.02/0.98` in a scratch copy and re-ran the suite:

```
FAILED test_rail_metric_regression_capped_p_yes_at_floor
FAILED test_rail_metric_uses_model_prob_cap_thresholds
assert '60.0%' in '… Rail (0-2.0% / 98.0%-100%): 0.0% [OK] …'
```

The failure output reproduces the original defect verbatim — a vacuous `0.0% [OK]`. That is exactly the guard #910 asked for: it exercises the real SQL against real rows, so it cannot be satisfied by a stubbed `COUNT`. Restored cleanly afterward; 29/29 pass on the real fix.

Label now derives from `rail_upper` directly, so the printed range and the queried range have one source. Fixtures cleaned up, assertions scoped to the rail line.

**CI green on head SHA `24ad167`:** Lint, Test (pytest), gitleaks, AI / NVIDIA NIM review.

Two non-blocking notes, neither worth holding the merge:

- **[suggestion] The artifact line lost its `(pre-fix 17.8%)` baseline.** I only questioned the *rail* baseline, because the rail definition changed in this PR — the artifact definition did not, so that annotation was still valid context. It's moot in practice since #911 redefines that metric entirely, so I'll let #911 decide what baseline belongs there.
- **[question] The deploy note says the next 14:00 UTC firing picks the change up automatically.** True only once the code is actually on the host. That merged-vs-deployed gap is precisely what cost two days of data on 07-28/29 — it's the subject of #913, and this PR is a small live example of the assumption. No change needed here.

Merging.

PR #919, opened 2026-07-31
fix(#917): conserve probability mass across the bracket ladder

### LINKED ISSUE #917: [P0] Probability mass not conserved across bracket ladder — SUM(raw_p_yes) median 0.70, ~0.53 on US F ladders; likely inclusive/exclusive bracket_high mismatch biasing NO edge upward

ClaudioSenior:
## Linked issue
Closes #917

## What changed
`parse_bracket_from_market()` (`src/strategy/scanner.py`) was feeding
`p_normal_between()` an **inclusive**-upper integer boundary for three of its
five label branches, but `p_normal_between` integrates `[low, high)` with an
**exclusive** upper bound (#861/#881, deliberate to avoid double-counting at
bracket boundaries). US 2°F-wide labels like `"88-89°F"` mean *the daily high
is 88 OR 89* — the real interval `[88, 90)` — but were integrated as
`[88, 89)`, exactly half the bracket's true width.

`_LABEL_EXACT` already handled this correctly with a `+ 1` applied **before**
unit conversion. This PR applies the same pattern to the three broken
branches:

| Branch | Before | After |
|---|---|---|
| `_LABEL_RANGE_DASH` ("88-89°F") | `(88, 89)` | `(88, 90)` |
| `_LABEL_BETWEEN` ("between 28-30°C") | `(28C, 30C)` | `(28C, 31C)` |
| `_LABEL_LTE` ("55°F or below") | `(-50, 55)` | `(-50, 56)` |
| `_LABEL_GTE` ("92°F or above") | unchanged (lower bound already inclusive) | unchanged |
| `_LABEL_EXACT` ("21°C") | already correct | unchanged |

Per the diagnosis on #917, this is the mechanism behind the recorded
calibration finding that `raw_p_yes` under-predicts YES by 2-3x in the
0.02-0.20 band — precisely the NO entry zone.

**Out of scope (per Tech Lead PM instruction) — reported back, not fixed here:**
- °C-ladder station-polls also falling below 1.0 mass in places (a second,
  separate mechanism, most likely observation-censoring without renormalisation).
- The 2 station-polls with `SUM(raw_p_yes) > 1.1`, still unexplained.

Both need follow-up issues; not attempted in this PR.

## How to test
```
python -m pytest src/tests/test_scanner.py src/tests/test_scan_decisions.py \
  src/tests/test_scanner_low_side.py src/tests/test_scanner_shadow.py \
  src/tests/test_envelope.py -q
```
191 passed.

- `TestParseBracketFromMarket` now asserts exact `(low_f, high_f)` for all 5
  label branches, including two °C cases that lock in the "+1 before
  `_to_f`" ordering (`test_between_celsius_plus_one_before_conversion`,
  `test_range_dash_celsius_plus_one_before_conversion`).
- New `TestBracketLadderMassConservation` class: builds the real 2°F US
  ladder convention from the diagnosis (`"55°F or below"` … `"90-91°F"` …
  `"92°F or above"`), asserts it is gap-free, and asserts
  `SUM(p_normal_between(...))` over the full ladder is `~1.0` (tolerance
  `<= 0.02`) for two different means — this is the invariant whose absence
  let the bug survive; it's the real deliverable of this PR per the Tech
  Lead PM's diagnosis.
- `src/tests/test_scan_decisions.py::TestGateVerdicts::test_below_min_confidence_no_side`
  used a `"78-79°F"` bracket that, once corrected to its true width
  `[78, 80)`, now sits close enough to the test's `forecast_high_f=80` mean
  to trip `below_min_edge` instead of `below_min_confidence` — a genuine
  behavioural consequence of the fix, not a test bug. Swapped the fixture
  bracket to `"74-75°F"`, which still isolates the confidence gate under the
  corrected math.

## NO-side impact (acceptance criterion 4)
Raising `p_yes` lowers `ev_no = (1-p)*100 - no_ask - fee`, which reduces NO
entries — NO is the only side this bot trades, so this had to be measured
before merge.

This isolated worktree has no access to the live `data/analytics.db` /
`logs/bot.log` (both gitignored, populated only on the production host), so
a live-data replay wasn't possible from here. Instead I measured the
magnitude analytically, using `scanner.py`'s own `ev_no` formula and
production-default gates (`MIN_EDGE_CENTS=15`, `MIN_PRICE_CENTS=70`,
`MAX_CONFIDENCE_YES_FOR_NO=0.05`) against the real 2°F ladder, sweeping
forecast mean/sigma and market `no_ask` price independently (the fix only
changes the bot's own `p_yes` estimate — `no_ask` is set by the market and
is unaffected):

```
scenarios evaluated (NO-zone, min(p_old,p_new)<=0.30): 13050
would-enter NO under OLD (buggy) parser:   4561 (35.0%)
would-enter NO under NEW (fixed) parser:   3905 (29.9%)
pre-fix candidates that flip OFF post-fix: 656 / 4561 (14.4% of old candidates)
avg p_yes delta (new - old), all scenarios: +0.0273
avg ev_no delta (new - old) cents, all scenarios: -2.73c
avg ev_no delta (new - old) cents, among OLD candidates only: -0.80c
```

**Summary: ~14% of scenarios that previously cleared all three NO-entry
gates (edge, confidence, price) no longer clear them post-fix**, with an
average ~0.8c reduction in `ev_no` among those. This is the correction
working as intended per the diagnosis — every NO-side edge computed on a
°F market has been systematically overstated by mispricing the bracket
width — but the Tech Lead PM should be aware of the live-book volume
reduction before this deploys, since a true measurement against production
data (not available from this isolated worktree) should be run as a
follow-up once this lands.

## How to deploy
- [x] **Code only** — `git pull` + restart the scanning service (`meteoedge-run`,
  or equivalent per current ops setup). No schema, dependency, or config
  changes.
- [ ] Dependencies changed
- [ ] DB schema change
- [ ] One-off script(s) to run after deploy
- [ ] Config / env change
- [x] **Post-deploy validation** — re-run the `SUM(raw_p_yes)` query from the
  original issue against fresh station-polls; °F station-polls should move
  from ~0.53 toward ~1.0 (acceptance criterion 3). Also watch the NO-entry
  rate in `scan_decisions` / live trade logs for the expected reduction
  quantified above.
- [ ] No deploy needed

## Checklist
- [ ] Docs updated — not needed, no schema/API/env changes


P0K5:
## Review: **request changes** — the fix is correct, the PR is not mergeable

This PR touches trading logic, so I did the full manual deep-dive rather than the AI-first path, and verified the substance independently instead of trusting the summary.

### What I verified myself — all good

**The parser fix is exactly right.** `+1` applied *inside* `_to_f(float(...) + 1, unit)` on all three defective branches, before unit conversion, matching `_LABEL_EXACT`. `_LABEL_GTE` correctly untouched (lower bounds are already inclusive). This was the single thing most likely to be gotten backwards and it wasn't.

**The mass-conservation tests are real guards.** I reverted only the parser change in a scratch copy and re-ran:

```
FAILED TestBracketLadderMassConservation::test_ladder_is_gap_free_and_non_overlapping
FAILED TestBracketLadderMassConservation::test_sum_of_p_yes_conserves_mass
FAILED TestBracketLadderMassConservation::test_sum_of_p_yes_conserves_mass_near_tail
3 failed
```

Restored, 23/23 pass. `test_ladder_is_gap_free_and_non_overlapping` is the one I especially wanted — it encodes the invariant structurally (`prev.high_f == nxt.low_f`) rather than checking a number, so it will catch the next variant of this bug too. This is the standard I want on the rest of the queue.

**The changed fixture in `test_scan_decisions.py` is legitimate**, and I'm satisfied it isn't masking a regression. Widening `78-79°F` to `[78, 80)` moves it close enough to the `forecast_high_f = 80` mean that `below_min_edge` binds before `below_min_confidence` — a genuine consequence of correcting the width, in the conservative direction. Moving the fixture to `74-75°F` preserves the test's actual intent (exercising the confidence gate). The inline comment explaining it is exactly what I want to see when a test changes alongside a behavioural fix. Flagging it only so it's on the record as reviewed, not waved through.

### [blocker] CI has never run on this branch

```
check-runs total_count: 0
gh run list --branch mid/917-bracket-mass-conservation  ->  (empty)
```

Governance is non-negotiable here: `CI / lint`, `CI / test`, and `AI / NVIDIA NIM review` must all be green on the head SHA before approval. Zero runs is not "not yet green", it's "never attempted". Needs investigating — most likely a consequence of the conflict below.

### [blocker] PR is `CONFLICTING` with master

`mergeable: CONFLICTING`. #918 merged while this was in flight. Rebase on latest master.

### [blocker] Remove the `graphify-out/` artifacts from this PR

```
179815+ 167089-  graphify-out/graph.json
  1380+   1101-  graphify-out/GRAPH_REPORT.md
   685+    665-  graphify-out/manifest.json
    59+      1-  graphify-out/.graphify_labels.json
```

347,000 changed lines of regenerated graph data around a **12-line** production fix. This repo maintains those artifacts through dedicated `chore(graph): incremental graph update [skip ci]` commits, not inside feature PRs. It is almost certainly the source of the merge conflict, it will collide with every other PR in flight, and it makes this diff unreviewable by anyone who didn't already know where to look. Drop them from the branch and let the normal graph-update job regenerate them.

### [question] NO-side impact was estimated analytically, not backtested

I asked for a measured backtest; the worktree had no production DB, so you produced an analytic estimate (~14.4% of qualifying NO scenarios no longer clear all gates, ~0.8c average `ev_no` drop). That is a reasonable substitute and it points the right way — the fix is conservative, it removes NO entries rather than creating them.

I ran the blast-radius check you couldn't, against `data/meteoedge.db`:

```
open positions now:            0
last live trade:               2026-07-24T00:06:47Z
live trades since 2026-07-17:  6
entry verdicts since 07-24:    (none)
```

Live entry has been dormant for a week, so the practical exposure is near zero and your estimate does not need upgrading before merge. Noting it here so the record is accurate rather than leaving the analytic number to look sturdier than it is.

### Follow-ups

The two items you correctly deferred are now filed as separate issues — thank you for reporting rather than expanding scope, that was the right call.

---

Fix the three blockers and re-request. The code itself I'm happy with; this is entirely about mergeability and CI.

ClaudioSenior:
@P0K5 re-requesting review — all three mergeability blockers fixed, see issue #917 comment for details. `mergeable: MERGEABLE`, CI green (lint/test/AI-review) on head SHA `12aedbb`.

P0K5:
## Review: **approve** — P0 cleared

All three blockers resolved, and I re-verified the substance after the rebase rather than carrying forward my earlier approval. A rebase that resolves a conflict can silently drop changes, so a P0 touching trading logic gets re-checked, not assumed.

**Production fix intact** — all three branches still carry the `+1` inside the conversion:

```
-        lo, hi = -50.0, _to_f(float(m.group(1)), unit)
+        lo, hi = -50.0, _to_f(float(m.group(1)) + 1, unit)
-        lo, hi = _to_f(float(m.group(1)), unit), _to_f(float(m.group(2)), unit)
+        lo, hi = _to_f(float(m.group(1)), unit), _to_f(float(m.group(2)) + 1, unit)   (x2)
```

**Guards still guard** — re-ran the revert check on the rebased branch: 3 passed with the fix, 3 failed without it.

**Diff is clean**: `scanner.py` (12+/3-), `test_scanner.py` (87+/12-), `test_scan_decisions.py` (6+/1-). No `graphify-out/`.

**CI green on head SHA `12aedbb`**: Lint, Test (pytest), gitleaks, AI / NVIDIA NIM review. `mergeable: MERGEABLE`.

Useful root cause on the CI gap, worth recording for #923 and anyone who hits it again: the original PR-opening push never fired an Actions webhook — the check-suites API showed zero GitHub-Actions suites for that SHA, with permissions and branch protection all correct. The force-push from the rebase produced a normal `synchronize` event and everything ran. So "zero check runs" can be a dropped webhook rather than a misconfiguration, and a new head SHA is the remedy.

Merging. #920 (censoring without renormalisation) and #921 (`SUM > 1.1`) were held in Backlog behind this and now move to Ready.

Issue #820, opened 2026-07-24
[Bug] Evening live entries price tomorrow's market with today's finished observations — certainty shortcuts emit a false p_yes=0


P0K5:
## Problem

Every KORD **live** entry fires between **19:00–22:00 CDT**. That is not a strategy choice — 19:00 CDT *is* 00:00 UTC, the instant `is_next_day` flips `1 → 0` and the live-entry gate unlocks (next-day candidates are force-shadowed under #687).

In that window the two halves of the evaluation disagree about which day it is:

- **The market**: `is_next_day` is derived from `endDate.date() == today_utc`. After 00:00 UTC the *next* local day's market is classified `is_next_day=0` (same-day, live-eligible).
- **The `WeatherState`**: built for the **station's local day** (`datetime.now(pytz.timezone(STATION_TZ[station]))`, `src/weather/builder.py`), which has **not** rolled over — it is still the previous, essentially finished, Chicago day.

So for ~5 hours every night we evaluate **tomorrow's** market against **today's completed** observations.

### Why that produces a false certainty

At local hour 19–22, `expected_additional_rise()` returns `0.0`, so `compute_envelope()` collapses `max_env` onto the finished day's realized high. The envelope's certainty shortcuts (`src/model/envelope.py:222-227`) then fire:

- `hi <= state.current_high_f` → `return 0.0`
- `lo > max_env` → `return 0.0`

The model therefore declares a bracket **impossible** for tomorrow's high on the basis of today's finished range. This is visible in the data as an *exact* `p_yes_raw = 0.0` (not a Gaussian value):

| Entry (CDT) | Bracket | Price | `p_yes_raw` |
|---|---|---|---|
| 07-21 19:03 / 20:05 / 21:00 | 74–75 | 78¢ | **0.0** |
| 07-16 19:08 | 88–89 | 76¢ | **0.0** |
| 07-15 20:01 → 21:08 | 92–93 | 78¢ | **0.0** |
| 07-17 20:04 → 22:06 | 86–87 / 90–91 | 76–78¢ | **0.0** |

`p_yes_raw = 0.0` clamps to the `MODEL_PROB_CAP` floor of `0.05`, which is exactly the `MAX_CONFIDENCE_YES_FOR_NO` threshold — so the NO gate opens automatically.

## Why it has been profitable anyway

[...]

PR #922, opened 2026-07-31
fix(#912): gate_verdict allow-list drift -- migrate off the DDL CHECK, one source of truth

### LINKED ISSUE #912: gate_verdict allow-list drift: db.py rejects 'day_mismatch_shadow' that scanner.py emits — day-mismatch brackets silently dropped from scan_decisions
## Symptom

`scan_decisions` rows are being silently dropped in production:

```
2026-07-30T03:04:27 WARNING [__main__] [scan_decisions] upsert failed for 0x736d…: invalid gate_verdict: 'day_mismatch_shadow'
```

75 occurrences in the retained log window (43 of them on 2026-07-30 alone, plus 07-25/26/27).

## Root cause: two copies of the same frozenset, drifted

`src/strategy/scanner.py:54-58` — **12** verdicts:

```python
GATE_VERDICTS = frozenset({
    "traded_live", "shadow_only", "next_day_shadow", "entry_guard", "timeout_today",
    "below_min_edge", "above_max_edge", "below_min_price", "below_min_confidence",
    "margin_gate", "mae_gate", "day_mismatch_shadow",
})
```

`src/data/db.py:1034-1038` — **11** verdicts, missing `day_mismatch_shadow`:

```python
_SCAN_DECISION_GATE_VERDICTS = frozenset({
    "traded_live", "shadow_only", "next_day_shadow", "entry_guard", "timeout_today",
    "below_min_edge", "above_max_edge", "below_min_price", "below_min_confidence",
    "margin_gate", "mae_gate",
})
```

The scanner emits `day_mismatch_shadow` (scanner.py:979, 1046 — the #820 day-mismatch path), `db.upsert_scan_decision` raises `ValueError` at db.py:1109, and the call site swallows it as a WARNING. The docstring at db.py:1099 even says "one of the 11 canonical verdicts" — the count was never updated when the 12th was added.

`src/tests/test_scan_decisions.py:326` already lists `day_mismatch_shadow` as valid, so the test suite and the validator disagree.

## Fix

1. Delete `Database._SCAN_DECISION_GATE_VERDICTS` and import `scanner.GATE_VERDICTS` as the single source of truth (or move the frozenset to a shared constants module if the import direction is wrong).
2. Add a test asserting the validator accepts every member of the scanner's `GATE_VERDICTS` — so this class of drift fails CI instead of production.
3. Drop the hardcoded "11 canonical verdicts" count from the docstring.
4. Backfill is not needed (rows are re-upserted each poll), but note the day-mismatch brackets have been invisible in `scan_decisions` since the verdict was introduced.

## Impact

Every bracket that hits the day-mismatch path is missing from `scan_decisions` — i.e. from M3 station-day accrual and from every M3 indicator. Exactly the brackets #820 was about.

ClaudioSenior:
Closes #912

## What changed

The `scan_decisions.gate_verdict` allow-list was duplicated in three places:
`scanner.GATE_VERDICTS`, `Database._SCAN_DECISION_GATE_VERDICTS`, and a
hand-written DDL `CHECK` constraint inside `CREATE TABLE IF NOT EXISTS`.
When `'day_mismatch_shadow'` was added to the first two (#820) but not the
DDL, every *existing* database kept rejecting that verdict with a raw
`sqlite3.IntegrityError` — SQLite cannot `ALTER` a `CHECK` in place, and
`CREATE TABLE IF NOT EXISTS` is a no-op against a table that already exists.
Verified against `data/meteoedge.db` in the #912 review before this PR.

A previous attempt (`junior/912-gate_verdict_drift`) only widened the DDL
string, which does nothing for a live table, and its test used the standard
fresh-DB fixture, which builds from the *new* DDL and therefore cannot
detect that gap. This PR does not build on that branch.

1. **Single source of truth**: new `src/strategy/gate_verdicts.py` holds
   `GATE_VERDICTS`. Both `scanner.py` and `db.py` import it — `db.py` can't
   import from `scanner.py` directly (that would cycle:
   `db -> scanner -> taf_disruption -> db`), so the shared module lives one
   level down from both.
2. **Dropped the `CHECK` entirely**, rather than generating it from the
   frozenset. Chose this over the "generate from frozenset" alternative
   because the `CHECK` was already redundant with the Python validator in
   `upsert_scan_decision` (which raises a clearer `ValueError` and runs
   *before* the row reaches SQLite), and dropping it means a future verdict
   never needs a second edit anywhere — the exact failure mode this issue is
   about.
3. **Migration**: `Database._migrate()` now rebuilds `scan_decisions`
   (create → copy → drop → rename) for any existing database still carrying
   the old `CHECK`, detected via `sqlite_master`'s stored `CREATE TABLE`
   text — same pattern as the existing `trades.mode` migration a few lines
   above it. Idempotent: a DB without the old `CHECK` (fresh installs,
   already-migrated DBs) is a no-op.
4. **Migration test**: `src/tests/test_scan_decisions_gate_verdict_check_migration.py`
   builds the table from the *exact* pre-#912 production DDL (the 11-verdict
   `CHECK`, missing `day_mismatch_shadow`) rather than the `db` fixture, then
   asserts the insert both raises `IntegrityError` before migration and
   succeeds after, following the `_legacy_db()` pattern in
   `test_emos_crps_log_sigma_source_migration.py`.
5. Updated `docs/DB_SCHEMA.md`'s `scan_decisions` section: DDL, column
   table, and a new note explaining the dropped constraint and migration.

## How to test

```
python -m pytest src/tests/test_scan_decisions_gate_verdict_check_migration.py -v
python -m pytest src/tests/test_scan_decisions.py src/tests/test_scanner.py src/tests/test_db.py -q
```

All 145 pass locally. Full suite run also passed with no *new* failures —
the ~33 pre-existing failures on this Windows box (symlink privilege errors,
`os.unlink` file-lock races on temp `.db` files, CI-script tests) reproduce
identically on unmodified `master`.

## How to deploy

- [x] **DB migration step required.** No manual action needed beyond a
  normal deploy — `Database.__init__` runs `_migrate()` on every open, and
  the new migration is idempotent and self-detecting (checks
  `sqlite_master` for the old `CHECK` text before touching anything). The
  next time any process opens `data/meteoedge.db` (or any other existing
  `scan_decisions`-bearing DB) after this deploy, the table is rebuilt in
  place automatically, in one transaction, with all existing rows
  preserved.
- [x] Restart the poller/`run.py` process (or any long-lived process holding
  a `Database` connection) so it picks up the migrated schema — a
  connection opened *before* this deploy won't re-run `_migrate()`.
- No env/config changes.
- Post-deploy validation: confirm no more
  `[scan_decisions] upsert failed for ...: CHECK constraint failed` lines in
  logs, and spot-check `sqlite3 data/meteoedge.db ".schema scan_decisions"`
  no longer shows a `CHECK` on `gate_verdict`.

## Trade-offs / uncertainties

- Went with "drop the CHECK" per the PM's stated lean in the review, rather
  than "generate the CHECK from the frozenset." Happy to switch if the PM
  prefers keeping a generated DDL-level guard as defense-in-depth — the
  migration logic would need a small tweak (detect "missing verdict in
  CHECK" instead of "any CHECK present") but the shared-module structure
  would be unchanged.
- The migration test's `_legacy_db()` hardcodes the pre-#912 DDL text
  (rather than deriving it) so it stays byte-for-byte faithful to what's
  actually running in production today, matching the pattern the review
  pointed to. It will not automatically track any *future* CHECK-constraint
  incident on this table — but there shouldn't be one, since the CHECK is
  now gone.


P0K5:
## Review: **request changes** — the code is the best work of the day; the PR packaging is not

I reviewed this manually rather than AI-first, since it carries a DB schema migration.

### The substance is right, and I verified it rather than trusting the summary

**The migration test suite is exactly what I asked for.** `_legacy_db()` builds the *exact pre-#912 production DDL* with the old CHECK, then:

- `test_legacy_schema_rejects_day_mismatch_shadow_at_the_sql_layer` — `pytest.raises(sqlite3.IntegrityError)`, proving the bug exists before the fix
- `test_migration_fixes_the_production_failure`
- `test_migration_preserves_existing_rows` — data preservation, which I did not ask for and should have
- `test_migration_is_idempotent`

10/10 pass. This test *starts from the broken state* and therefore can actually detect the migration gap — the precise thing the first attempt could not do.

**The migration itself is safe.** Detects the old constraint via `sqlite_master`, rebuilds inside a `with self._conn` transaction, copies via an explicit intersected column list (so column-order drift can't silently misalign data), and recreates `idx_scan_decisions_station_date`. I checked production: that is the only real index (the other is the PK autoindex, recreated automatically), and the table is **3,102 rows**, so the rebuild is effectively instant with no lock concern.

**Single source of truth achieved**, which the first attempt did not do. `src/strategy/gate_verdicts.py` is imported by both `scanner.py` and `db.py`, the CHECK is gone rather than widened, and the import direction is justified against a `db → scanner → taf_disruption → db` cycle. `_SCAN_DECISION_GATE_VERDICTS = GATE_VERDICTS` as a compat alias for tests that patch it is a thoughtful touch. `docs/DB_SCHEMA.md` updated — correctly treated as mandatory for a schema change.

You picked "drop the CHECK" over "generate it from the frozenset" and flagged the trade-off. Confirmed: that is the right call, for the reason your own code comment gives — a hand-maintained CHECK requires a rebuild migration on every future verdict addition, and one such addition already shipped without one.

### [blocker] CI has never run

```
gh pr checks 922  ->  no checks reported on the 'mid/912-gate-verdict-migration' branch
```

Zero check runs. `CI / lint`, `CI / test`, `AI / NVIDIA NIM review` must be green on the head SHA. Non-negotiable, and it matters more here than usual — this migration mutates a production table.

### [blocker] `mergeable: CONFLICTING`

#918 merged while you were working. Rebase on latest master.

### [blocker] Remove the `graphify-out/` artifacts

```
180503+ 167582-  graphify-out/graph.json
  1391+   1105-  graphify-out/GRAPH_REPORT.md
   695+    665-  graphify-out/manifest.json
    58+      1-  graphify-out/.graphify_labels.json
```

~348,000 changed lines wrapped around a 112-line migration. These are maintained by dedicated `chore(graph): incremental graph update [skip ci]` commits, never inside feature PRs, and they are almost certainly what makes this PR conflict.

This is **not your fault specifically** — it happened independently on #919 too, so the cause is systemic (the mandatory `graphify query` step regenerates them, and `git add -A` sweeps them up). I've filed it separately. For now: `git checkout origin/master -- graphify-out/`, commit that, and confirm `git status` is clean of `graphify-out/` before your final push.

### Process note — acknowledged, and handled correctly

You edited the shared checkout instead of the worktree, caught it, reverted, and re-applied in the right place. I verified the shared checkout independently: clean, HEAD intact at `2bf68cc`, no source files modified, no other agent's work disturbed. Reporting it rather than quietly moving on was the right call and it saved me a much longer investigation.

---

Three blockers, all packaging. The code I'm happy to merge the moment CI is green on a rebased, artifact-free branch.

P0K5:
All three packaging blockers fixed:

1. **graphify-out/ artifacts removed** — reset to `origin/master` state and committed separately from the migration commit. `git diff origin/master -- graphify-out/` is empty at the current head.
2. **Rebased on latest master** — master advanced twice while CI was pending (#918 then #919 merged), so I rebased twice; `mergeStateStatus` is now `CLEAN`, `mergeable: MERGEABLE`.
3. **CI now runs and is green** on head `9581340`: `CI / lint`, `CI / test`, `AI / NVIDIA NIM review`, and `gitleaks` all pass.

Diff is now exactly the intended 5 files (`docs/DB_SCHEMA.md`, `src/data/db.py`, `src/strategy/gate_verdicts.py`, `src/strategy/scanner.py`, `src/tests/test_scan_decisions_gate_verdict_check_migration.py`) plus the graphify-out cleanup commit. No code changes — per your review, the migration/refactor/tests were correct as-is.

"How to deploy" section already covers automatic migration on next `Database()` init and the required poller/run.py restart, so left unchanged.

P0K5:
## Review: **approve**

All three blockers cleared, and I re-verified the migration after two rebases rather than carrying my earlier approval forward.

### I ran your migration against a real copy of production

Not the test fixtures — a `.backup()` copy of `data/meteoedge.db`, migrated via `Database()` init:

```
BEFORE  rows=3102  CHECK_present=True   indexes=[idx_scan_decisions_station_date, sqlite_autoindex_...]
AFTER   rows=3102  CHECK_present=False  indexes=[idx_scan_decisions_station_date, sqlite_autoindex_...]
day_mismatch_shadow insert on migrated PRODUCTION copy -> OK
```

Full-table SHA-256 over all 34 columns, ordered by primary key:

```
ORIGINAL  rows=3102  sha=7615228b1dc63df9
MIGRATED  rows=3102  sha=7615228b1dc63df9
DATA IDENTICAL: True
```

Byte-identical. The rebuild preserves every row, the index survives, the constraint is gone, and the verdict that has been failing in production since #820 now inserts successfully against a real migrated database. 10/10 migration tests still pass post-rebase.

**[nitpick, no action needed]** The rebuilt table reorders two columns — `execution_mode` now precedes `ensemble_*`, where the original had it after. It cost me a false alarm on my first digest, which compared in `PRAGMA table_info` order. It is harmless: I checked, and there is no `SELECT *` against `scan_decisions` anywhere in `src/`, so nothing depends on positional access. Worth knowing it happened.

**CI green on head SHA `9581340`**: Lint, Test (pytest), gitleaks, AI / NVIDIA NIM review. `mergeStateStatus: CLEAN`. Diff is the 5 intended files, artifacts isolated in their own cleanup commit referencing #923 — exactly the right way to handle it.

### On your `ensure_worktree.sh` report

Good catch, and thank you for verifying the worktree was genuinely valid before proceeding rather than either trusting the tool or ignoring it. Filing it separately — every agent on this repo will hit it.

Merging. This closes the silent data-loss path that dropped 53,279 upserts across 07-28/29.

PR #926, opened 2026-07-31
fix(#915): dedupe before exclude in gate ordering (final_poll)

### LINKED ISSUE #915: Gate admits bracket-days on non-final polls; methodology note says otherwise
## The behaviour

`bss_market_vs_model_report.run_report` excludes rows and **then** de-duplicates:

```python
kept_rows, exclusion_counts = apply_exclusions(raw_rows)   # :1190
deduped = dedupe_one_per_bracket_day(kept_rows)            # :1191
```

So when a bracket-day's **final** poll is excluded, the bracket-day is not dropped — it re-enters the scored population represented by the last *surviving* poll.

## Why that is a problem

**1. The methodology note is false.** The report states:

> De-duplication keeps the final (lowest `minutes_to_settlement`) poll per (station, ticker, settlement date), mirroring `calibration_report.pick_samples`.

It keeps the final **surviving** poll. That is only the final poll for bracket-days whose last word happened not to be excluded. This is the same prose-vs-code drift that produced #902 — a reader trusts the note instead of opening the code.

**2. The 300-station-day power bar is partly met by these rows.** A station-day admitted through a poll further from settlement is not the same statistical object as one whose closest-to-settlement poll was genuinely scoreable. The bar decides *when the gate is allowed to run*, so this directly affects the M3 timeline.

**3. Nobody chose it.** The ordering is not a documented decision — it falls out of statement order. Whichever ordering is right, it should be picked deliberately.

## Which ordering is correct is genuinely open

- **`gate` (exclude → dedupe):** a bracket-day is represented by the model's last *non-artifactual* word. Defensible: an excluded poll is by definition not a scoreable opinion, so falling back to the last one that was seems reasonable.
- **`final_poll` (dedupe → exclude):** a bracket-day is represented by the model's last word, full stop; if that is not scoreable, the bracket-day is dropped. Defensible: it is what the note already claims, and it avoids letting *when the artifact happened to fire* decide which poll represents a market.

Note the second ordering interacts with time-of-day: `p_yes_raw == 0.0` arises from the envelope certainty shortcuts, whose #820 failure mode was specifically the evening window. So which poll represents a bracket-day is not independent of the clock.

## Not asserting a bias direction

Both `p_model` and `p_market` are read from the same row, so this does not obviously favour either side — an earlier poll is less informed about the weather for the model and for the market alike. The problem is that the population is undefined and undocumented, not that it is provably tilted.

## Measurement

`certainty_exclusion_check` now runs both orderings over the same rows and reports the difference in bracket-days and station-days, so the size of the affected population is measured rather than inferred across run dates:

```bash
python -m src.scripts.certainty_exclusion_check --population all-bracket
```

(An earlier estimate compared the 2026-07-29 gate run against the 2026-07-30 check — different dates, so that inference was not sound. Hence measuring it directly.)

## Acceptance criteria

- [ ] Decide the ordering explicitly and record the reasoning
- [ ] Make the code and the methodology note agree, whichever way it goes
- [ ] If `final_poll` wins: re-read the station-day count against the 300 bar — the gate may be further from powered than the current number suggests
- [ ] Settle it **before** the powered Pass-2 run (~2026-08-05), while the number is still unknown

---

Refs #822 (the gate), #909 (the exclusion decision this surfaced from), #820 (origin of the excluded rows), #902 (the previous prose-vs-code drift)

P0K5:
Closes #915

## What changed

Swapped the order in `run_report` so de-duplication runs **before** certainty exclusions, not after. A bracket-day whose final (lowest `minutes_to_settlement`) poll is excluded is now dropped rather than falling back to an earlier non-final poll.

- `src/scripts/bss_market_vs_model_report.py`: swapped `apply_exclusions` / `dedupe_one_per_bracket_day` call order in `run_report` (~:1190-1191)
- Updated methodology notes in module docstring and `build_report` output to document `final_poll` ordering: dedupe first, then exclude
- `src/scripts/certainty_exclusion_check.py`: updated two comments that described the old (reverse) ordering — the check already used `final_poll` ordering, so both now agree

## Why

The methodology note already claimed deduplication keeps the final poll per bracket-day — but the code excluded first and then deduped among survivors, so the note was false. The Tech Lead PM chose `final_poll` ordering because:
- It matches what the methodology note already claims
- It mirrors `calibration_report.pick_samples`
- It is conservative for the 300-station-day power bar
- A bracket-day whose final opinion was unscoreable should not enter through an earlier poll

## How to test

- Run the existing test suites: `python -m pytest src/tests/test_bss_market_vs_model_report.py src/tests/test_certainty_exclusion_check.py` (102 passing)
- Run the diagnostic: `python -m src.scripts.certainty_exclusion_check --population all-bracket --no-network`
- The contested station-day count is **112** under `final_poll` ordering (unchanged from before, since the check already used this ordering)

## How to deploy

- [x] Code-only change. Restart MeteoEdge after merge. No DB migration required.

## Trade-offs

- The `final_poll` ordering is conservative: it drops bracket-days whose final poll was excluded, reducing the scored population slightly. The 300-station-day power bar will be reached slightly later than under the old ordering.
- The station-day count in `certainty_exclusion_check` was already measured under dedupe-first ordering, so the 112 station-day figure is already the `final_poll` number.

---  
Generated with [Claude Code](https://claude.com/claude-code)


P0K5:
## Tech Lead PM Review — APPROVE ✅

**Overall:** Correct, clean, docs match code. The two-line swap is exactly what the judgment called for.

### Review dimensions

| Dimension | Assessment |
|---|---|
| Correctness | ✅ Swap correct. `final_poll` ordering: dedupe then exclude. Station-day count re-read (112, unchanged). |
| Architecture | ✅ Consistent with `calibration_report.pick_samples` |
| Code quality | ✅ Clean, methodology notes updated in both places |
| Performance | ✅ No impact — same operations, different order |
| Test coverage | ✅ Existing 102 tests pass. One logged follow-up (see below) |
| Documentation | ✅ Docstring + report prose both updated |
| How to deploy | ✅ Filled: restart, no DB migration |

### Follow-up logged

One **[suggestion]** from AI reviewer: add a regression test in `test_bss_market_vs_model_report.py` asserting that a bracket-day whose final poll is excluded is dropped entirely. Logged as a follow-up issue in Backlog — not blocking merge given the change's triviality.

### CI gate

- CI / lint ✅
- CI / test ✅ (102 passed)
- AI / NVIDIA NIM review ✅ (standalone workflow passed; CI-embedded duplicate is a runner config issue)

PR #928, opened 2026-07-31
fix(#916): pass settlement_date from position_tracker to true_probability_yes

### LINKED ISSUE #916: position_tracker marks open positions with the #820 certainty shortcuts unguarded
## The gap

#820 did not remove the envelope certainty shortcuts — it guarded them behind a day-mismatch check that only arms when `settlement_date` is passed (`src/model/envelope.py:290-304`):

```python
_day_mismatch = (settlement_date is not None
                 and state.now_local.date() != settlement_date)
...
if not _day_mismatch:
    if hi <= state.current_high_f:
        return 0.0
    if lo > max_env:
        return 0.0
    if lo <= state.current_high_f and hi >= max_env:
        return 1.0
```

`settlement_date` defaults to `None`, so **any caller that omits it gets the pre-#820 behaviour.** Of the three production call sites, two pass it and one does not:

| Caller | Passes `settlement_date`? |
|---|---|
| `scanner.py:839` | ✅ |
| `scanner.py:841` | ✅ |
| **`position_tracker.py:151`** | ❌ |

```python
p_yes_now = true_probability_yes(bracket_stub, state)
```

## Impact

`p_yes_now` becomes `fair_value_now` — the mark-to-model value for an **open position**, written into position snapshots and shown on the chart. In the evening window, where the state's local day has not yet rolled but the market's settlement day has, `current_high_f` and `max_env` are yesterday's observations. The shortcuts then produce a false `0.0` or `1.0` fair value from the wrong day's weather — exactly #820's mechanism, in the exit path rather than the entry path.

**Not a live-entry risk** (the scanner is guarded) and **not a contaminant of the M3 gate** — `bracket_evals.p_yes_raw` is written from the scanner path, which passes `settlement_date`. I verified this while settling the provenance question on #909; the gate's population is clean on this point.

The exposure is to position marking: a false fair value can misinform an exit decision or make the chart show a position as worthless/certain when it is neither.

## Fix

Pass the position's settlement date at `position_tracker.py:151`. The bracket's settlement day is already available in the surrounding snapshot construction.

Worth considering separately: `settlement_date: date | None = None` makes the unsafe path the default, so a future caller silently inherits the bug. Making it required — or defaulting the guard to *on* when the date is unknown — would fail loudly instead.

## Testing

A test that calls `true_probability_yes` without `settlement_date` on a day-mismatched state and asserts the shortcuts do **not** fire.

---

Refs #820 (the original bug and its guard), #909 (found while settling provenance there)

P0K5:
Closes #916

## What changed

- `src/execution/position_tracker.py`: `_log_open_position_snapshots()` now passes `settlement_date=date.fromisoformat(today)` to `true_probability_yes()`. The `today` variable (already in scope, used for filtering open positions by UTC date) carries the position's settlement day since all MeteoEdge positions are same-day markets.
- `src/tests/test_envelope.py`: added regression test `test_no_settlement_date_skips_all_shortcuts` that covers all three certainty shortcut paths on a day-mismatched state, verifying each is bypassed when settlement_date is passed and differs from the state's local day.

## Why

The #820 fix added a day-mismatch guard in `true_probability_yes()` (lines 290-304) that skips observation-based certainty shortcuts when the weather state's local day differs from the settlement day. However, this guard only arms when `settlement_date` is passed. `_log_open_position_snapshots()` was calling it without `settlement_date`, so the guard never triggered -- and the certainty shortcuts could fire on the wrong day's observations in the exit path, exactly #820's mechanism.

## How to test

```bash
python -m pytest src/tests/test_envelope.py::TestDayMismatchGuard -xvs
```

All 12 tests in TestDayMismatchGuard pass, including the new regression test covering all three shortcut paths (hi <= current_high_f, lo > max_env, lo <= current_high_f AND hi >= max_env).

## How to deploy

- [x] No deploy needed -- source-code change only, no config/env/DB changes. The fix activates on the next poll cycle when position snapshots are logged.


P0K5:
## Tech Lead PM Review — APPROVE

**Overall:** Clean single-call-site fix with a thorough regression test. Exactly what the issue calls for.

### Review dimensions

| Dimension | Assessment |
|---|---|
| Correctness | The `today` variable already in scope carries the settlement day (same-day markets). `date.fromisoformat(today)` conversion correct. |
| Architecture | No signature changes, just passing an existing optional param. Minimal blast radius. |
| Code quality | Clean, well-documented, follows existing patterns. |
| Test coverage | Excellent regression test covering all three shortcut paths with realistic KORD evening-window scenario. |
| Documentation | Inline docstring explains the regression. No doc changes needed. |
| How to deploy | "No deploy needed" — correct, source-code-only. |

### Follow-up

The decision not to change the `settlement_date: date | None = None` default is correct for this PR — that requires a separate pass across 15+ test assertions. Logging as a suggestion for future hardening.

### CI gate

- CI / lint
- CI / test (pytest)
- AI / NVIDIA NIM review: PASS (no blocking issues)

Issue #932, opened 2026-08-04
Health report says the M3 power bar is met (362/300) from the wrong table, with no exclusions


P0K5:
The 2026-08-04 daily health report printed:

```
M3 Progress
  Station-days:  362/300 (121%)
  Accrual rate:  ~33/day -> power bar ~Aug 04 + 0d
```

That reads as "the gate may run today." It should not be trusted, for three independent reasons.

## 1. It counts the wrong table — and its own comment says so

`src/scripts/daily_health_report.py:406-411`:

```python
# Count distinct (station, date) pairs in bracket_evals since
# the clean-data window started (2026-07-24).
station_days = db._conn.execute(
    "SELECT COUNT(DISTINCT station || '-' || date) FROM scan_decisions "
    "WHERE poll_ts >= '2026-07-24'"
).fetchone()[0] or 0
```

The comment says `bracket_evals`; the SQL reads `scan_decisions`. **The M3 gate scores `bracket_evals`** (`bss_market_vs_model_report --population all-bracket`). These are different sinks with different write paths, so the number does not describe the population the gate will use.

This is the third instance of the same class of defect — prose asserting one thing while the code does another (cf. #902, #915). Worth treating as a pattern rather than three coincidences.

## 2. It applies no exclusions

It counts raw station-days. The gate scores only rows surviving the `p_yes_raw == 0.0` and 1¢/99¢ rail exclusions, which removed **64.4%** of rows on the 2026-07-29 run. A station-day with no scoreable row still counts here.

For scale: on 2026-07-30 the certainty check measured the *contested* class at **101 station-days** while naive counts were far higher.

## 3. It predates the #915 ordering fix

#915 changed the gate to de-duplicate before excluding (`final_poll`), which drops bracket-days whose final poll is excluded. This counter still reflects pre-#915 semantics.

## Why it matters now

The gate is blocked for unrelated reasons — #920 and #921 leave the 2026-07-24 window contaminated (see `docs/REMEDIATION_PLAN.md`, M3 section). But a green "121%, power bar met" is exactly the kind of signal that gets acted on, and it is the only automated readout anyone checks daily.

## Also in scope: the Evaluated counter is misleading

```
Evaluated (scan_decisions) 24h: 4
```

[...]

Issue #921, opened 2026-07-31
Two station-polls have SUM(raw_p_yes) > 1.1 — truncation bugs cannot cause this, so a third independent mechanism exists


P0K5:
Split out of #917. Two station-polls show `SUM(raw_p_yes) > 1.1`, peaking at **1.156**.

```sql
SELECT poll_ts, station, SUM(raw_p_yes)
FROM scan_decisions WHERE poll_ts >= '2026-07-28' GROUP BY 1,2;
-- min 0.061, median 0.704, max 1.156  (n = 114 station-polls)
```

RPLL at 2026-07-30T11:42 sums to 1.156 across 11 brackets.

## Why this needs its own investigation

Both #917 (parser dropping the top unit) and the censoring defect are **truncation** bugs — they can only ever *remove* probability mass. Neither can produce a sum above 1.0. A total of 1.156 therefore indicates a third, independent mechanism, and it is the one most likely to be **overstating** `p_yes` rather than understating it.

Candidates worth checking first:
- Overlapping brackets after parsing (double-counted mass), especially where a ladder mixes label formats.
- The `max_env` envelope expansion at `src/model/envelope.py:282-283` interacting with the certainty shortcuts.
- Duplicate rows per `(station, ticker, date)` inflating the sum — rule this out before pursuing a model explanation.

## Scope

1. Rule out the mundane causes (duplicate rows, overlapping ladder) before investigating the model.
2. If real, determine direction of the error and whether it can inflate a YES-side probability.
3. Add the case to the mass-conservation invariant from #917 as a two-sided bound, not just a floor — the current test asserts `approx(1.0, abs=0.02)`, which already catches it, so confirm why it did not fire on production data.

Do not start until #917 and the censoring issue have landed, or the three effects will be impossible to separate.

P0K5:
Board: → **Backlog**. Split out of #917 during review of PR #919. Deliberately blocked until #917 merges — landing multiple probability changes in one window would make the effects impossible to separate.

P0K5:
Board: Backlog → **Ready**. Unblocked — #917 merged as `46d5515`, so probability changes from the parser fix are now isolated in history and this can be measured against a clean baseline.

P0K5:
[...]

Issue #920, opened 2026-07-31
Probability mass lost to observation censoring: brackets below the running daily high are zeroed without renormalising the remainder (RKSI sums to 0.288)


P0K5:
Split out of #917 (see the Tech Lead PM diagnosis there). #917 fixes the °F dash-range parser; this issue covers the **residual** deficit that fix does not explain.

## Evidence

Celsius-derived ladders parse via `_LABEL_EXACT`, which was already correct (it has the `+ 1`), so their edges are genuinely contiguous — e.g. RKSI `75.2 / 77.0 / 78.8` (24/25/26°C), zero gaps. Yet many °C station-polls still sum well below 1.0.

RKSI, 2026-07-30T11:42, `current_high = 86.0`:

| bracket | raw_p_yes |
|---|---|
| all brackets below 86.0 | 0.0 |
| 86.0 – 87.8 | 0.1383 |
| 87.8 – 89.6 | 0.0966 |
| 89.6 – 91.4 | 0.0365 |
| 91.4 – 200 | 0.0168 |
| **total** | **0.288** |

Zeroing brackets below an already-observed daily high is correct — a running maximum cannot decrease. But the surviving brackets should then carry the **conditional** distribution `P(max in bracket | max >= 86)`, which must sum to 1.0. Instead they carry the unconditional tail of a distribution centred well below 86, and ~72% of the mass simply disappears.

## Hypothesis

The observation-censoring path truncates the distribution without renormalising the remainder. If the model's mean is ~80 while the observed high is already 86, the certainty shortcut zeroes the lower brackets but nothing re-centres or re-normalises what is left.

## Why it matters

Same directional consequence as #917: understated `p_yes` inflates `ev_no = (1-p)*100 - no_ask - fee` on the only side the bot trades. #917 removes the °F half-mass loss; this removes the censoring loss. Both must land before any calibration or M3 conclusion drawn from `raw_p_yes` can be trusted.

## Scope

1. Confirm the mechanism in `true_probability_yes` / `compute_envelope` (`src/model/envelope.py`).
2. Decide the correct semantics: renormalise the surviving mass, or re-anchor the forecast mean to `max(mu, current_high)`. These are not equivalent — justify the choice.
3. Extend the mass-conservation invariant added by #917 to cover the censored case (ladder with a running high partway up must still sum to ~1.0).
4. Quantify the effect on NO-side edge before merge.

Blocked on #917 merging first, to avoid two probability changes landing in one window.

[...]

PR #933, opened 2026-08-04
feat(m3): land the ladder mass-conservation diagnostic on master

### LINKED ISSUE #948: Assert ladder mass conservation against production data, as a standing check
## Problem

**Nothing in the stack asserts that a gap-free bracket ladder sums to ~1.0 on production data.**

That is not a theoretical gap. It is the direct cause of two probability-mass defects surviving for weeks in `p_yes_raw` — the exact quantity the M3 decision gate scores:

- **#917** — °F dash-range brackets integrated at half width. Ladders summed **0.53**.
- **#920** — brackets outside the surviving interval zeroed without renormalising the remainder. °C ladders summed **0.80**.

Both were found by ad-hoc measurement on 2026-07-31, not by any check. #917 added mass conservation as a *synthetic-ladder* unit test, which passes happily while production ladders sum to 0.53 — synthetic ladders are constructed complete, so the test cannot see a parser or truncation defect at all.

The cost was thirteen days of contaminated collection: the entire first M3 window (2026-07-24 → 2026-08-05) is void, and every station-day in it had to be discarded.

## What is needed

Land `m3_window_diagnostics` on master as ordinary tracked code, reframed from one-off triage into a standing regression check, and document how to run it.

It reads `logs/bracket_evals.*.jsonl` — the gate's own population, **not** `scan_decisions`, which upserts on `(station, ticker, date)` and therefore assembles "ladders" from brackets observed at different times. It reports:

1. **Parser version by day.** Bracket width fingerprints the code version (1.0 °F pre-#917, 2.0 °F post-fix, 1.8 °F = °C), so a deploy is confirmed in the data rather than by trusting a systemd timestamp.
2. **Mass conservation, split by censoring and by which end the ladder was cut.** The split is what makes it diagnostic rather than just an alarm: an uncensored deficit isolates a parser bug, a censored-only deficit isolates a renormalisation bug, and a `top`/`both` deficit against a healthy `none` column is exactly what #920 looked like.
3. **Scoreable station-day accrual** — after the gate's own exclusions, which is what the 300 bar counts and what the health report's raw count overstates (#932).

## Acceptance criteria

- [ ] `src/scripts/m3_window_diagnostics.py` on master, with tests
- [ ] `scripts/run_m3_diagnostics.sh` runs the **deployed** module (fast-forwards master first), rather than extracting a copy from a branch
- [ ] `--since` defaults to `M3_CLEAN_DATA_CLOCK_START` rather than a frozen literal, with a test that fails if the two drift apart — the clock has already moved twice
- [ ] Documented in `docs/OPERATIONS.md` under Monitoring, including how to read each failure mode
- [ ] Read-only: opens no database connection, writes nothing under `logs/` or `data/`

## Why now

The tool has been run off an unmerged branch via `git show` since 2026-08-01. That is exactly the friction that gets a check quietly dropped — and this is the check whose absence already cost one full collection window.

Refs #917, #920, #921, #932, #822.

P0K5:
Closes #948

## What

Lands `m3_window_diagnostics` + its host runner on master as ordinary tracked code, reframed from one-off triage into a standing regression check, and documents it in `docs/OPERATIONS.md`.

**Rescoped.** This branch previously also carried `docs/REMEDIATION_PLAN.md` edits. Those were written while #920 was still open — they framed M3 as blocked, named `envelope.py:301` as an open item, and said the clock restarts when #920 *and* #921 land. All three are now false. Merging that snapshot would have put a stale blocking notice back into the document the M3 decision is read from, so the plan-doc work was rewritten against current reality and merged separately as #947.

## Why

This tool found both probability-mass defects that voided the first M3 window, and it is still **the only thing in the stack that checks the invariant they violated**: a gap-free bracket ladder must sum to ~1.0. #917 added mass conservation as a *synthetic-ladder* test — which passes happily while production ladders sum to 0.53, because synthetic ladders are constructed complete. Nothing asserted it against production data. That is how two independent defects survived for weeks in the quantity the gate scores, at a cost of thirteen days of discarded collection.

It has been run off this unmerged branch via `git show` since 2026-08-01. That is exactly the friction that gets a check quietly dropped.

## How

Reframed from discovery to regression check, since #917 and #920 are both fixed and deployed:

- **`--since` now defaults to `M3_CLEAN_DATA_CLOCK_START`**, imported from `daily_health_report` rather than duplicated. The clock has already moved twice; a frozen literal here is how this tool would go on cheerfully reporting over a contaminated window after the next move. A test fails if the two drift apart.
- Verdict and docstring no longer describe #920 as open, and **no longer ask the reader to confirm #921** — measured at 1.04 and stood down, so naming it as a precondition would re-arm a blocker we deliberately cleared.
- A censored-only deficit is now labelled what it would mean *today*: a renormalisation regression, not the expected state.
- `main()` split into `build_parser()` + `main()` so the default is testable directly.

`run_m3_diagnostics.sh` now **runs the deployed module after fast-forwarding master**, instead of extracting a frozen copy from this branch to `/tmp`. That dance existed only because the tool was unmerged; keeping it would mean the host permanently reporting through branch code while the bot runs master. Extra args are forwarded, so `--since 2026-07-24` still reaches the void window.

`docs/OPERATIONS.md` gains a Monitoring subsection: the read-only guarantee, what the wrapper does to the checkout (and that it always returns to master), and a symptom table for reading the verdict.

## How to test

```bash
python3 -m pytest src/tests/test_m3_window_diagnostics.py -q     # 32 passed
bash -n scripts/run_m3_diagnostics.sh
```

Full suite green: 3089 passed, 9 skipped.

The `--since` default test was mutation-checked — replacing `M3_CLEAN_DATA_CLOCK_START` with the old `2026-07-24` literal fails it.

On the bot host:

```bash
cd ~/MeteoEdge && bash scripts/run_m3_diagnostics.sh
```

Expect every cut level at 1.00 and `VERDICT: mass conserved from 2026-08-06`.

## Notes

- Refs #917, #920, #921, #932. Deliberately no `Closes #822` — that is the M3 gate issue and must stay open until the gate actually runs.
- `094f710` is an empty commit: three pushes to this branch produced a `pull_request` `synchronize` event that started no workflow run, so the head needed nudging to get CI to fire at all.


P0K5:
Addressed both blocking issues in `8b5fe05`, with one correction to the review.

**1. Missing `Closes #N` — valid, fixed.** Filed #948 (*"Assert ladder mass conservation against production data, as a standing check"*) and the PR body now opens with `Closes #948`.

Worth stating why this PR had no linked issue: the work it does belongs to #822, and `Closes #822` would auto-close the M3 decision gate — the issue that must stay open until the gate actually runs. This is the third time the reviewer has asked for a `Closes` keyword that would have closed #822; #909 and #937 were filed for the same reason. The rule is right; the only correct way to satisfy it here is a new issue scoped to the deliverable, which is what #948 is.

**2. Missing operational documentation — valid, fixed.** `docs/OPERATIONS.md` gains a Monitoring subsection covering the read-only guarantee, what the wrapper does to the checkout (and that an EXIT trap returns it to master on every path), and a symptom table for reading the verdict. The runbook had gone into `docs/REMEDIATION_PLAN.md`, which is where the M3 *argument* lives — not where an operator looks for how to run something on the host. Fair catch.

**3. "PR body missing what/why/how-to-test description" — this one is not correct.** The body reviewed at `094f710` carried all three under those exact headings, including a `How to test` section with the pytest invocation, the full-suite result, and the expected host output. Flagging it so the finding isn't carried forward as unaddressed; nothing changed on that count beyond the `Closes` line.

One note for whoever reads CI on this PR: `094f710` is an empty commit. Three pushes to this branch — a force-push, a rebase, and a normal push — each produced a `pull_request` `synchronize` event that started **no workflow run at all**, while #947's `opened` event triggered its checks normally. The empty commit is what finally got CI to fire. That looks like a repo-level Actions behaviour rather than anything about this branch, and may be worth a look independently.

---

PR #936, opened 2026-08-05
feat(#822): --since, so the gate cannot silently score contaminated eras

### LINKED ISSUE #937: M3 gate has no date filter — it would score three probability eras as one population
## The gap

`bss_market_vs_model_report` scores **every row** in `bracket_evals`. There is no way to restrict it to a date range.

That was harmless when the log held one model. It no longer does — two probability defects have landed mid-collection, splitting it into three incompatible eras:

| era | state | approx rows |
|---|---|---|
| before 2026-08-01 | #917 — °F ladders integrated at **half width** | ~5,900 |
| 2026-08-01 → 08-05 | #920 — truncation leaking up to **20%** of ladder mass | ~2,300 |
| from 2026-08-06 | clean | accruing at ~55 station-days/day |

## Why it matters now

The gate is scheduled for ~2026-08-11. At that point the window holds **13 contaminated days against 5 clean ones**, so an unfiltered run scores **~72% contaminated rows** — and produces a report indistinguishable from a clean one. The verdict would be wrong, would look right, and nothing in the output would say otherwise.

This is the last step of a sequence whose entire purpose was to establish that those rows must be excluded (#917, #920, and the measurement on #920). Without a cutoff, that work is undone silently at the point it is cashed in.

## What it should do

1. `--since YYYY-MM-DD`, filtering on **poll time, not settlement date**. Contamination is a property of when the probability was *computed*, not of when the market resolved: a bracket polled 2026-08-04 for a 2026-08-05 settlement came from pre-#920 code, and the weather that settled it is not in question.
2. Applied **before** de-duplication, so a bracket-day is represented by its last *clean* poll rather than dropped because its final poll predates the cutoff.
3. **Not enforced in code.** Omitting it stays legitimate for a Pass-1 archive read, whose data is uniformly pre-fix. Instead, an unwindowed run must print a banner naming the eras and stating it is not a legitimate gate run — that banner is the only thing separating a mis-invocation from a wrong verdict.
4. An empty window writes nothing. An empty BSS is worse than none, because it looks like a result.

## Worth noting as a pattern

This was invisible for the same reason #917's mass-conservation test was: **the control existed one layer away from where the defect could occur.** #917 asserted conservation against the parser rather than the served probability, so it stayed green while production ladders summed to 0.80. The gate validates everything about *how* a bracket is scored — BSS math, exclusions, de-duplication, outcome resolution, market-price convention — and nothing about *which rows are in scope*.

Both were real controls that could not see the thing they were guarding. Worth a look at whether other invariants in this repo sit at the wrong layer.

## Acceptance criteria

- [ ] `--since` filters by poll time, before de-duplication
- [ ] Windowed runs state the window and the rows it cost, in the header and the funnel
- [ ] Unwindowed runs print the era warning — asserted by a test, not left as prose
- [ ] Empty window returns without writing a report
- [ ] `docs/REMEDIATION_PLAN.md` runbook carries the flag and why it is not optional

---

Refs #822 (the gate this protects), #917 and #920 (the two eras), #826 (the log)

P0K5:
Closes #937

## Linked issue

Closes #937. **Deliberately not `Closes #822`** — #822 is the M3 gate itself and must stay open until the gate actually runs (~2026-08-11). Closing it here would auto-close the project's central tracking issue the moment this merges. #937 was opened for this specific gap.

## What changed

`bss_market_vs_model_report` has **no date filter** — it scores every row in `bracket_evals`. That log now spans three incompatible probability eras:

| era | defect | approx rows |
|---|---|---|
| before 2026-08-01 | #917 — °F ladders integrated at half width | ~5,900 |
| 2026-08-01 → 08-05 | #920 — truncation leaking up to 20% of ladder mass | ~2,300 |
| from 2026-08-06 | clean | accruing |

Adds `--since YYYY-MM-DD`, filtering on **poll time, not settlement date**, applied **before** de-duplication.

## Why

Run on 2026-08-11 as written, the gate would score **~72% pre-#920 rows** — 13 contaminated days against 5 clean — and produce a completely normal-looking report. The verdict would be wrong, would look right, and nothing in the output would say otherwise. That is the last step of a sequence whose whole purpose was establishing those rows must be excluded.

Poll time is the correct key because contamination is a property of when the probability was *computed*, not when the market resolved. A bracket polled 2026-08-04 for a 2026-08-05 settlement came from pre-#920 code; the weather that settled it is not in question.

Filtering before de-duplication means a bracket-day is represented by its last *clean* poll rather than dropped because its final poll predates the cutoff.

### Why it isn't enforced in code

Omitting `--since` stays legitimate for a Pass-1 archive read, whose data is uniformly pre-fix — so it can't be a hard requirement. Instead an unwindowed run prints:

> ⚠️ **No `--since` given.** `bracket_evals` spans three incompatible probability eras … Legitimate for a Pass-1 archive read, which is uniformly pre-fix; **not legitimate for the M3 gate.**

That banner is the only thing between a mis-invocation and a wrong verdict, so it's asserted by a test rather than left as prose.

A windowed run states the window and its cost in both header and funnel:

```
**Poll-date window:** rows polled on or after **2026-08-06** (99 of 100 earlier rows excluded).

| Excluded: polled before `2026-08-06` (contaminated era) | 99 |
| Rows in window | 1 |
```

An empty window returns without writing anything — an empty BSS is worse than no BSS, because it looks like a result.

## How to test

```bash
python3 -m pytest src/tests/ -q     # 3058 passed, 9 skipped
python3 -m flake8 src/
```

7 new tests in `TestSinceFilter`, including `test_an_unwindowed_gate_run_is_flagged_not_silent` — the guard against the exact failure this PR exists to prevent — and `test_filters_on_poll_time_not_settlement_date`, which pins the key choice.

## How to deploy

- [x] **No deploy needed** (analysis tooling only — not imported by any running service)

Used when the gate runs, ~2026-08-11:

```bash
python -m src.scripts.bss_market_vs_model_report --population all-bracket --since 2026-08-06
```

## Checklist
- [x] Tests pass locally (3058 passed, 9 skipped)
- [ ] Project board status updated (In review) — needs Projects V2 access
- [x] Docs updated — plan runbook carries the flag and why it is not optional


P0K5:
Addressed the blocking issue — PR body now carries a closing keyword.

**One correction to the review's remedy, though.** It asks for `Closes #822`. That would be wrong: **#822 is the M3 gate itself**, still open, and merging with that keyword would auto-close the project's central tracking issue days before the gate actually runs (~2026-08-11).

The policy is right; the target isn't. Opened **#937** for this specific gap and pointed `Closes #937` at it. #822 stays referenced, not closed.

This is the second time the rule has pointed at the wrong issue — same thing happened on #908, where `Closes #822` would have closed the gate before the exclusion decision it depended on was even made. Both times the underlying cause was work that genuinely had no issue of its own. Worth noting in case the reviewer prompt can distinguish "link an issue" from "close the epic this PR happens to mention."

Everything else in the review passed, including trading safety and blast radius.

---

Issue #939, opened 2026-08-06
Ladder-mass check fragments ladders and reports false warnings


P0K5:
The ladder-mass check added in #932 fired on its first real run (2026-08-06 email):

```
Ladder mass:   0.947 mean, 0.464 worst (10/55 deficient) [WARN]
               ^ probabilities are leaking mass (cf. #917, #920) -- station-days below are NOT clean
```

**The check is wrong, not the model.** `m3_window_diagnostics`, reading the append-only `bracket_evals` over the same window, found **322 ladders, 0 deficient, every one at 1.000**. The cross-check between the two implementations is what caught it.

## Cause 1 — ladders split at the 24h boundary

The query grouped by `(station, poll_ts, date)` and filtered `poll_ts >= 24h ago`, justified in the docstring as *"within one poll cycle a ladder is intact."*

That is false. `scan_decisions` upserts on `(station, ticker, date)`, so a market that stops being scanned — settled, delisted, outside station active hours — keeps an **older `poll_ts`**. Any ladder whose brackets were last written across the 24h boundary is split into fragments, and each fragment is scored as its own deficient ladder. A 5-bracket fragment of an 11-bracket ladder sums to ~0.46 — precisely the reported "worst".

The correct key on an upsert table is `(station, date)`: that pair already identifies one ladder, because every ticker holds exactly one row. The date filter belongs on `date` — which settlement day a ladder belongs to — not on `poll_ts`, which only records when a bracket last happened to be written.

## Cause 2 — column index shifted

Dropping `poll_ts` from the `SELECT` moved the columns left while the reader still used `r[3]`. That silently became `COUNT(*)`, so every ladder reported as `11.000` regardless of its mass. Caught by the worst-vs-mean test.

## Why it matters more than the numbers

A monitor that cries wolf is worse than no monitor: it trains the reader to ignore the one signal it exists to raise. This check was added specifically because #917 and #920 went undetected for weeks — a false positive on day one is the fastest way to make it ignored by the time it is real.

## Fix

- Group by `(station, date)`; filter on `date`, not `poll_ts`
- Read `r[2]`, and note the column layout at the call site
- Minimum ladder size 5, so a part-listed station-day is not scored
- Fixtures use realistic 11-bracket ladders instead of 4
- Regression test for the straddling-ladder case, verified by mutation

## Follow-up, not fixed here

[...]

Issue #924, opened 2026-07-31
scripts/ensure_worktree.sh false-positives as 'not a valid worktree' on Windows/Git-Bash — compares C:/ toplevel against /c/ POSIX path, spuriously BLOCKs agents reusing valid worktrees


P0K5:
Reported by a mid-dev agent during PR #922, and independently plausible for every agent on this repo running under Git-Bash on Windows.

## Symptom

`scripts/ensure_worktree.sh`'s reuse-check falsely reports an existing, valid worktree as "not a valid worktree" and BLOCKS the agent.

## Root cause

The check compares:

- `git rev-parse --show-toplevel`, which on Windows Git-Bash returns a **Windows-style** path: `C:/Coding/MeteoEdge/.claude/worktrees/...`
- against its internally-built `WT_DIR`, a **POSIX-style** path: `/c/Coding/MeteoEdge/.claude/worktrees/...`

The two never compare equal, so reuse of a valid worktree always false-positives as invalid.

## Impact

- Every agent that reuses an existing worktree hits a spurious BLOCKED.
- The failure mode is worse than a plain error: an agent that trusts the tool abandons a perfectly good worktree; an agent that distrusts it works around the guard, which defeats the point of having it. On PR #922 the agent verified manually (`git worktree list`, branch, toplevel) and proceeded — correct in that instance, but it is not a check we want agents routinely overriding.
- This repo's normal operating mode is several parallel agents in worktrees, so it recurs.

## Fix

Normalise both sides before comparing. Options, cheapest first:

1. Convert `git rev-parse --show-toplevel` output to POSIX form (`cygpath -u`, available in Git-Bash), or convert `WT_DIR` the other way with `cygpath -w`.
2. Compare canonicalised real paths on both sides (`cd "$X" && pwd -P`) rather than string-comparing.
3. Compare `git rev-parse --git-common-dir` / worktree identity instead of filesystem paths — the most robust, since it stops depending on path spelling entirely.

Prefer 3 if straightforward, else 1.

## Acceptance criteria

- Reusing an existing valid worktree succeeds on Windows/Git-Bash without a BLOCKED message.
- A genuinely invalid or missing worktree is still correctly rejected — do not fix the false positive by weakening the check into a no-op.
- Add a regression test or a documented manual verification covering both cases.

## Related

#923 (agents bundling `graphify-out/` artifacts) — both are agent-tooling defects that surface as recurring, avoidable friction across parallel agents rather than as product bugs.

[...]

Issue #923, opened 2026-07-31
Agents bundle ~348k lines of regenerated graphify-out/ artifacts into feature PRs — two independent occurrences, both caused merge conflicts and a wasted review round-trip


P0K5:
## Problem

Two independent agents, on two unrelated issues, produced PRs bundling ~348,000 lines of regenerated `graphify-out/` artifacts around small production changes:

| PR | Issue | Production diff | `graphify-out/` diff |
|---|---|---|---|
| #919 | #917 | 12 lines (`scanner.py`) | 347,000+ lines |
| #922 | #912 | 112 lines (`db.py`) | 348,000+ lines |

Both PRs came back `mergeable: CONFLICTING`, and in both cases the artifacts are the most likely cause. Both required a review round-trip that had nothing to do with the code.

## Root cause

The two rules interact badly:

1. `.claude/instructions/graphify.md` (and the PreToolUse hook) **mandate** running `graphify query` before exploring code — so every agent touches the graph tooling on every task.
2. Running the tooling regenerates `graphify-out/graph.json`, `GRAPH_REPORT.md`, `manifest.json`, and `.graphify_labels.json` in the working tree.
3. Agents then stage with `git add -A` (or equivalent) and sweep the regenerated artifacts into the feature commit.

Meanwhile the repo's actual convention is that these files are maintained by dedicated `chore(graph): incremental graph update [skip ci]` commits — see the commit history, where they always land on their own.

So the workflow instructs every agent to do the thing that dirties the tree, and nothing tells them not to commit the result.

## Why it's worth fixing rather than catching in review

- Every affected PR is unreviewable at a glance — a 12-line fix is buried under 347k lines.
- The artifacts collide across concurrently open PRs, producing conflicts unrelated to the actual work.
- It costs a full review round-trip per occurrence. It has already cost two.
- It scales with the number of parallel agents, which is the normal operating mode for this repo.

## Proposed fix (pick one or combine)

1. **Explicit instruction** in `.claude/skills/dev-workflow/SKILL.md` and `.claude/instructions/graphify.md`: never stage `graphify-out/`; run `git checkout -- graphify-out/` before committing. Cheapest, but relies on compliance.
2. **Stop staging them by accident**: forbid `git add -A` / `git add .` in the dev workflow in favour of explicit path staging. Addresses the wider class of accidental inclusion, not just this one.
[...]

PR #971, opened 2026-08-10
fix(#969): read rail/artifact leading indicators from bracket_evals

### LINKED ISSUE #969: Health report rail/artifact WARNs are structural — wrong table, wrong-population thresholds, rails conflated
## Context

The daily health report's two M3 leading indicators — `Rail` and
`p_yes=0.0 artifact` — are computed on a population their thresholds were never
calibrated for. Both fire `[WARN]` every day, and the artifact one propagates into
the report's overall `Verdict` line. On 2026-08-10 the report read:

```
  Rail (0-5.0% / 95.0%-100%): 70.0% [WARN]
  p_yes=0.0 artifact:     42.7% [WARN]
...
Verdict
  [WARN] Stable -- watch: artifact rate high
```

while `post_fix_model_health --since 2026-08-06`, run on the same host the same
day, reported the model **healthy on exactly these measures**: high rail 1.1%
against a 9.1% structural ceiling (12% of it), middle mass 35.5% against a
pre-fix 0.0%, and **0 interior-zero gaps in 2,564 ladders**.

Two reports on the same data disagreeing is the problem. The authoritative one is
`post_fix_model_health`, which documents in its own text that it *"replaces two
earlier, wrong versions of this check: a raw exact-zero rate compared against Pass
1's 17.8% (invalid cross-population comparison)"*. The health report still makes
that exact comparison.

## Three defects, one block

`src/scripts/daily_health_report.py::_build_m3_progress`, lines ~676–716.

**1. Wrong table.** It queries `scan_decisions`. The ladder-mass check *directly
above it in the same function* was moved to `bracket_evals` by #943, for the
documented reason that `scan_decisions` upserts on `(station, ticker, date)` and
therefore assembles "ladders" from brackets observed at different poll times. The
rail/artifact block was left behind on the old source. `REMEDIATION_PLAN.md` is
explicit: the diagnostic must read *"`logs/bracket_evals.*.jsonl` — the gate's own
population, **not** `scan_decisions`"*.

**2. Thresholds from the wrong population.** `rail_ok = rail_pct < 20` and
`zero_ok = zero_pct < 10` are gate-selected-archive numbers (Pass 1's 62.6% rail /
17.8% artifact baselines). They are applied to the **full bracket ladder**, which
is mostly far-out-of-the-money brackets that legitimately sit at ~0. An 11-bracket
ladder has at most one bracket that can resolve YES; a low-rail share near 58% is
the *correct* behaviour, not a fault. `rail_pct < 20` can therefore essentially
never pass. This is the same class as #910 (a threshold that can never fire),
inverted: a threshold that can never clear.

**3. The two rails are conflated.** The metric sums low and high rail into one
number. `post_fix_model_health` is explicit that the **high** rail is the one
carrying the verdict (it has a computable structural ceiling of `1/ladder_size`,
so it is population-robust) and that the low rail is population-driven and not
comparable across populations. Summing them destroys the only half that means
anything.

## Why this matters beyond noise

A `[WARN]` that fires unconditionally is worse than no metric: it trains the
reader to ignore the verdict line, which is precisely how *"'Healthy' was printed
after two days of silent data loss"* (#913) happened. The M3 gate runs within
days; this is the report that is supposed to catch a regression in the gate's own
input before it does.

## Acceptance criteria

- [ ] Rail and artifact metrics are computed from `bracket_evals` (the gate's
      population), consistent with the ladder-mass check in the same function.
- [ ] The rail metric reports the **high** rail against its structural ceiling
      (`1 / ladder_size`), matching `post_fix_model_health`'s definition. Report
      the low rail separately as context, or not at all — but do not sum them.
- [ ] Thresholds are derived from the same population being measured, and a
      comment states which population each baseline came from.
- [ ] On the current clean window (`>= 2026-08-06`), the health report's rail and
      artifact lines **agree with `post_fix_model_health`'s verdict** — i.e. both
      read healthy. A test asserts the two agree, so they cannot drift apart again.
- [ ] The `Verdict` line no longer reports "artifact rate high" on a healthy window.

## Technical notes

Prefer **reusing** `post_fix_model_health`'s existing computation over
reimplementing it — a second implementation is how these two drifted apart in the
first place. If the leading-indicator logic can be imported as a function and
called from both, that is the fix; the health report then becomes a thin renderer
of the authoritative numbers.

Note the interior-zero-gap check is the structural invariant that replaced the raw
exact-zero rate. If a "does the model still emit false certainty" line is wanted
in the daily report, **that** is the number to surface — it needs no baseline and
no population assumption — not the raw `raw_p_yes = 0.0` share.

## Test requirements

- Unit test: synthetic full-ladder input where the low rail is high by
  construction and the high rail is near zero → the report must read healthy.
- Unit test: high rail at its structural ceiling → must read `[WARN]`.
- Regression test asserting the health report and `post_fix_model_health` return
  the same verdict on the same fixture window.
- Test that the `Verdict` line is not set to WARN by a healthy leading-indicator
  section.

## Dependencies

None. Should land before the M3 gate run so the pre-gate window has a trustworthy
daily check.

ClaudioSenior:
## Linked issue
Closes #969

## What changed
`_build_m3_progress`'s rail/artifact leading-indicator block queried `scan_decisions`
(the upsert table `_ladder_mass` / `_scoreable_pairs` were already moved off of by #943)
and thresholded a FULL bracket ladder against `rail_pct < 20` / `zero_pct < 10` --
Pass-1's GATE-SELECTED baselines. An 11-bracket ladder has at most one bracket that
can resolve YES, so a full ladder can essentially never clear `rail_pct < 20`; the
metric fired `[WARN]` unconditionally and disagreed daily with `post_fix_model_health`
on the same data.

- Added `_rail_and_artifact()`, which reads `bracket_evals` (via the same loader
  `_ladder_mass` already uses) and delegates the actual computation to
  `post_fix_model_health`'s `rail_concentration` / `structural_high_rail_ceiling` /
  `interior_zero_violations` -- one implementation, not two drifting apart.
- Reports the HIGH rail against its population-derived structural ceiling
  (`1 / ladder_size`) instead of summing low+high rail into one number.
- Reports interior-zero ladder-gap violations (the structural invariant that needs
  no baseline) instead of the population-confounded raw `p_yes_raw == 0.0` share.
- New `HIGH_RAIL_WARN_RATIO = 0.5` threshold (documented inline as a judgment call,
  not a measured baseline -- the ceiling itself is the population-robust half).
- Updated `_build_verdict`'s marker string (`"p_yes=0.0 artifact:"` -> `"Interior-zero
  gaps:"`) to match the renamed line.
- Removed the now-unused `MODEL_PROB_CAP` import.
- Ran `graphify update .` to refresh the committed graph.

Scope discipline: this PR touches only reporting/threshold logic in
`daily_health_report.py` and its tests -- nothing that feeds `p_yes_raw` or
`bracket_evals` itself.

## How to test
```
python -m pytest src/tests/test_daily_health_report.py src/tests/test_post_fix_model_health.py -q
```
All 63 tests pass, including:
- `TestBuildM3ProgressLeadingIndicators` (replaces the old `TestBuildM3ProgressRailMetric`,
  which pinned the removed `scan_decisions` behaviour): high rail far below / at its
  structural ceiling, interior-zero gaps present/absent, no-data-is-not-healthy.
- `test_agrees_with_post_fix_model_health` -- a regression test asserting the health
  report and `post_fix_model_health` return the SAME `[OK]`/`[WARN]` verdict on the
  same fixture window, computed via the same shared functions, so they cannot
  silently drift apart again.
- Updated `TestBuildVerdict` fixtures to use the new marker text.

Manually verified against the acceptance criteria's production numbers (host,
2026-08-10, clean window `>= 2026-08-06`): `post_fix_model_health` reports high rail
1.1% against a 9.1% structural ceiling (12% of it, well under the 50% WARN ratio) and
0 interior-zero gaps in 2,564 ladders -- both now read `[OK]` under the new logic,
consistent with `post_fix_model_health`'s own healthy verdict.

Also ran the full suite as CI does (`pytest src/tests -q -n auto --dist loadfile`):
35 pre-existing failures, all in files this PR does not touch (`test_emos_retrain.py`,
`test_shadow_low_rollout.py`, `test_ensure_worktree.py`, etc.) and all
Windows-sqlite-tempfile `PermissionError`s on `os.unlink` -- a local Windows
environment issue, not caused by this change.

## How to deploy
- [x] **Code only** — `git pull` + restart: `meteoedge-health-report.timer` (or just
      let the next scheduled run pick it up; the script is invoked fresh each run,
      nothing persists across runs)
- [x] **Post-deploy validation** — next daily health report email (or
      `python -m src.scripts.daily_health_report --dry-run`) should show `High rail
      (bracket_evals, >= 0.95):` and `Interior-zero gaps:` lines instead of `Rail (0-.../
      p_yes=0.0 artifact:`, and read `[OK]` on the current clean window instead of
      `[WARN]` on both.

## Checklist
- [x] Tests pass locally (`pytest src/tests/test_daily_health_report.py
      src/tests/test_post_fix_model_health.py -v`)
- [x] PR body includes `Closes #969`
- [ ] Docs updated -- `docs/REMEDIATION_PLAN.md`'s `#969` row still reads "Open"; left
      for the Tech Lead PM to update since it also carries M3 sequencing decisions
      across other open issues, outside this PR's reporting-logic-only scope.


P0K5 (changes requested):
## Request changes

Good work on the substance — reusing `post_fix_model_health`'s functions instead of forking the computation is exactly right, the docstrings explain *why* rather than *what*, and the `ceiling is None` branch correctly declines to print a pass/fail tag rather than guessing. The regression test asserting the two modules agree is the right guarantee.

Three items, one of which is a correctness blocker.

---

### `src/scripts/daily_health_report.py`

**[blocker] `_rail_and_artifact` filters on settlement date; `_ladder_mass` filters on poll time. They do not share a clock.**

```python
# _rail_and_artifact  -- end_date == settlement_date
rows = [r for r in raw_rows if (r.get("end_date") or "")[:10] >= since_date]

# _ladder_mass  -- ts == poll_ts
ts = r.get("ts") or ""
if ts[:10] < since_date:
    continue
```

The new docstring claims this reads the *"same source and clean-data clock as the ladder-mass check above"*. Same source, different clock.

Poll time is the correct one, and this is settled precedent rather than a preference — #941 (`fix(#939): mass check must filter on POLL time`) changed exactly this, and `REMEDIATION_PLAN.md` states it as a rule for the gate: *"The filter is on **poll time, not settlement date** (#941) — contamination is a property of when the probability was computed, not when the market resolved."*

Concrete failure: a next-day bracket **polled 2026-08-05** (pre-#920, un-renormalised truncation) that **settles 2026-08-06** passes this filter and is scored as clean. That is the precise contamination class the clock exists to exclude, and next-day markets are a large share of the population — the `post_fix_model_health` run on 08-10 put `next_day` at 15,034 of 28,575 rows. The error also runs the other way, dropping genuinely clean rows polled after the clock start that settled earlier.

Fix: filter on `r.get("ts")`, matching `_ladder_mass`. Then the docstring's claim becomes true.

Worth noting for your own reference: `post_fix_model_health --since` does filter on `settlement_date` (its 08-10 output header says so), so following it here was a defensible read. But the health report's other clean-window consumer is `_ladder_mass`, and having two clocks inside one report section is the thing to avoid. If you think `post_fix_model_health`'s own convention is the bug, say so and I'll open a separate issue — do not change it in this PR.

**[suggestion] The verdict marker can mislabel a high-rail WARN as an artifact problem.**

```python
if "Interior-zero gaps:" in m3_text and "[WARN]" in m3_text:
    warn_issues.append("artifact rate high")
```

`"[WARN]" in m3_text` matches a WARN anywhere in the M3 section. The section now has three WARN-capable lines (high rail, interior-zero gaps, ladder mass), so a high-rail alarm surfaces in the verdict as *"artifact rate high"* — the wrong name for a real alarm, which is its own small version of the problem this PR fixes.

This flaw predates your change, so it is not a blocker on you. But the section gained a second independently-WARN-able indicator here, which makes it materially more likely to fire wrong. Scope the match to the line, e.g. check the specific line for `[WARN]` and label it accordingly. I've filed it separately so it does not get lost either way.

---

### Repo hygiene

**[blocker] ~187k lines of `graphify-out/` artifacts are bundled into this PR.**

`graphify-out/graph.json` (+184,786/−179,016), `GRAPH_REPORT.md`, `manifest.json`, `.graphify_labels.json` — against +121/−39 of actual source change. This is the third occurrence of the recurrence tracked in **#923**, which documents the prior two as *"both caused merge conflicts and a wasted review round-trip."* Master regenerates the graph on its own cadence (`chore(graph): incremental graph update [skip ci]`).

Drop the four `graphify-out/` paths from the branch and force-push. `graphify update .` is still the right thing to run locally per the instructions — it just must not be committed onto a feature branch.

---

### CI

Not green, so this could not be approved on this pass regardless: `Test (pytest)` and `AI / NVIDIA NIM review` are both **pending** on `46d8c25`. `Lint (flake8)` and `gitleaks` pass. I'll re-check after the push.

On the 35 local failures you flagged — Windows sqlite tempfile `PermissionError` on `os.unlink` is a known local-environment artifact and not attributable to this change; CI is the authority there. Thanks for calling it out explicitly rather than letting it look like a clean run.

### Deploy section

Present and correctly filled — code-only, timer restart, and a concrete post-deploy assertion naming the exact new lines to look for. Good.

---

**#969 moved back to In progress.** Ping me when the clock fix and the graphify strip are pushed and CI is green.

P0K5 (approved):
## Approve

Both blockers resolved, verified independently rather than from the report.

**Clock fix** — `_rail_and_artifact` now filters on `r.get("ts")`, matching `_ladder_mass`, with the #941 rationale stated inline. The regression test is the right shape: the fixture carries a 0.99 high rail on the pre-clock row, so if that row ever leaks back into the window the test fails on a WARN rather than passing quietly. A test that only asserted the row count would not have that property.

**Repo hygiene** — branch diff is now `daily_health_report.py` + `test_daily_health_report.py` only (+274/-162). The four `graphify-out/` paths are gone.

**CI green on `d265b9d`**: `Lint (flake8)`, `Test (pytest)`, `AI / NVIDIA NIM review`, `gitleaks` — all pass, verified on the head SHA.

Scope was held correctly: #972 (verdict mislabelling) and `post_fix_model_health`'s own `--since` convention were both left alone as instructed.

---

**On the `ensure_worktree.sh` failure you flagged** — that is **#924**, already open and describing exactly this: `pwd` returning a POSIX path against `git rev-parse --show-toplevel` returning a `C:/` drive-letter path on Windows/git-bash, producing a false `BLOCKED`. Verifying isolation manually and proceeding was the right call, and reporting it rather than staying silent is what let me match it to the existing issue. I've added your occurrence to #924.

Merging.

PR #973, opened 2026-08-11
fix(health-report): migrate m3 verdict to flags contract

### LINKED ISSUE #972: Health report verdict mislabels any M3 WARN as 'artifact rate high'
## Context

Raised as a `[suggestion]` during review of PR #971 (#969). Logged separately
because it predates that PR and should not expand its scope.

`_build_verdict` in `src/scripts/daily_health_report.py` attributes a WARN to a
named cause by testing for a *line label* and a `[WARN]` marker **anywhere in the
same section**:

```python
m3_text = "\n".join(lines_by_section.get("m3", []))
if "Interior-zero gaps:" in m3_text and "[WARN]" in m3_text:
    warn_issues.append("artifact rate high")
```

The M3 section now carries **three** independently WARN-able indicators — high
rail vs. structural ceiling, interior-zero gaps, and ladder mass. Any one of them
firing produces the verdict line *"watch: artifact rate high"*, regardless of
which actually tripped.

## Why it matters

A misattributed alarm is a mild version of the failure #969 fixed: the daily
report is the only regression detector during the M3 collection window, and a
verdict that names the wrong cause sends the reader to the wrong place. It also
degrades trust in the verdict line generally, which is how #913 ("Healthy"
printed over two days of silent data loss) became possible.

Low severity — the report is still *correct* that something is wrong, it just
mislabels what. But it will fire wrong more often now that the section has more
than one WARN source.

## Acceptance criteria

- [ ] Each WARN-able M3 indicator maps to its own verdict label; a high-rail WARN
      reads as a rail/sharpness problem, not "artifact rate high".
- [ ] Multiple simultaneous WARNs in one section produce multiple distinct
      labels, not one arbitrary winner.
- [ ] Matching is scoped to the individual line, not the joined section text.
- [ ] Audit the other sections for the same pattern — the M3 block is where it
      is now most visible, but the label-plus-section-wide-marker approach is
      used elsewhere in `_build_verdict` and may have the same latent flaw.

## Technical notes

The section builders already return structured data alongside their rendered
lines (the `dict` in `list[tuple[str, list[str], dict]]`). Prefer having each
indicator report its own status in that dict and having `_build_verdict` read
flags, rather than string-matching rendered output at all — string-matching the
presentation layer is the underlying fragility, and #914 already had to fix
verdict scoping once.

## Test requirements

- Test: high rail WARNs, interior-zero gaps OK → verdict names the rail, not the
  artifact rate.
- Test: interior-zero gaps WARN, high rail OK → verdict names the gaps.
- Test: both WARN → both labels present.
- Test: section healthy → no M3 label in the verdict.

## Dependencies

Touches the same file and function as #969 / PR #971. Land **after** #971 to
avoid a conflict.

ClaudioSenior:
## What

Closes #972.

`_build_verdict` in `src/scripts/daily_health_report.py` attributed **any**
WARN inside the `m3` report section to `"artifact rate high"`, because the
check was a bare substring scan:

```python
m3_text = "\n".join(lines_by_section.get("m3", []))
if "Interior-zero gaps:" in m3_text and "[WARN]" in m3_text:
    warn_issues.append("artifact rate high")
```

The `m3` section now carries **three** independently WARN-able indicators
(ladder mass, high rail vs. structural ceiling, interior-zero gaps), so a
high-rail WARN or a ladder-mass WARN was misreported as an artifact-rate
problem.

## Why

Per the re-scoped issue and `_build_verdict`'s own docstring (added by
#914), the fix is the flags contract already defined there: each section
returns `(lines, flags)` and `_build_verdict` reads `flags`, not rendered
text. #914 migrated the `bot` section; this PR migrates `m3` the same way
(the `trading`/`pipeline`/`guardrails`/`emos`/`blockers` sections stay on
the legacy line-scanning path — out of scope per the issue's re-scoping
comment).

## How

- `_build_m3_progress` now returns `(lines, flags)`, with
  `flags["mass_status"]`, `flags["rail_status"]`, `flags["gaps_status"]`
  each one of `OK / WARN / UNKNOWN`, set at the point each indicator is
  actually evaluated.
- `_build_verdict` reads `flags_by_section["m3"]` and emits one distinct
  label per WARN-ing indicator:
  - `mass_status == WARN` → `"M3 ladder mass leaking"`
  - `rail_status == WARN` → `"M3 high rail vs structural ceiling"`
  - `gaps_status == WARN` → `"artifact rate high"` (unchanged label,
    but now correctly scoped to the indicator it actually names)
  Multiple simultaneous WARNs each add their own label.
- Removed the stale docstring sentence saying the migration should wait
  on "#911's concurrent edit to `_build_m3_progress`" — #911 has no PR in
  flight, and after #969 the M3 indicators read `bracket_evals`, not the
  `scan_decisions.forecast_high` column #911 concerns. The docstring now
  states migration status accurately (bot + m3 done, the rest queued as a
  fast-follow).

Reporting/verdict logic only — nothing touching `p_yes_raw`, `bracket_evals`,
or model inputs; the M3 clean-data collection window is untouched.

## Test plan

Required tests from the issue, all added to `TestBuildVerdict` in
`src/tests/test_daily_health_report.py`:

- [x] high rail WARNs, gaps OK → verdict names the rail, not artifact rate
      (`test_m3_high_rail_warns_gaps_ok_names_the_rail_not_artifact_rate`)
- [x] gaps WARN, high rail OK → verdict names the gaps
      (`test_m3_gaps_warn_rail_ok_names_the_gaps`)
- [x] both WARN → both labels present
      (`test_m3_both_rail_and_gaps_warn_produce_both_labels`)
- [x] section healthy → no M3 label in the verdict
      (`test_m3_healthy_section_produces_no_m3_label`)
- [x] ladder-mass WARN → names mass, not artifact rate
      (`test_m3_ladder_mass_warn_names_mass_not_artifact_rate`)

Existing `TestBuildVerdict` / `TestLadderMassIndicator` / rate-accrual tests
were updated to pass the new `flags` dict shape (m3 now always returns a
`(lines, flags)` tuple, so every direct caller of `_build_m3_progress` in
tests was updated to unpack it).

Ran locally:
```
python -m pytest src/tests/test_daily_health_report.py -q   # 67 passed
python -m flake8 src/scripts/daily_health_report.py src/tests/test_daily_health_report.py   # clean
```
Also ran the full suite (`python -m pytest src/tests/ -q`): 3107 passed, 34
failed — all 34 failures are in unrelated files (`test_ensure_worktree.py`,
`test_dashboard.py`, `test_emos_retrain.py`, `test_shadow_low_rollout.py`,
`test_bootstrap_pagination.py`, `test_check_prompt_drift.py`,
`test_sanitize_transcript.py`, `test_archive_snapshots.py`,
`test_retrain_script.py`) — none touch `daily_health_report.py`, and none
are in the diff.

## How to deploy

Code-only change to `src/scripts/daily_health_report.py` — `git pull` on the
host, no restart needed. The health report is a `Type=oneshot` systemd unit
(`daily_health_report.py --dry-run` or the mailer path), so every run is a
fresh process; there is nothing long-lived to bounce.

Post-deploy assertion: run `python -m src.scripts.daily_health_report --dry-run`
on the host and confirm the `Verdict` line, if it names an M3 problem, uses
one of `"M3 ladder mass leaking"`, `"M3 high rail vs structural ceiling"`, or
`"artifact rate high"` — matching whichever M3 indicator line(s) actually
show `[WARN]` in the `M3 Progress` section above it, not a generic
"artifact rate high" regardless of which indicator tripped.


P0K5 (approved):
## Approve

Clean migration, and it took the structural fix rather than the shallow one.

**Correctness verified beyond the diff:**
- `rail_ok` is scoped entirely inside the `if ceiling:` branch, so dropping the `rail_ok = True` from the else leaves no dangling reference — checked, not assumed.
- `mass_ok` is preserved at both its remaining uses, including the `station_days >= 300 and not mass_ok` gate that stops the M3 bar being announced on unclean mass. That interaction was the easy thing to break here.
- No callers of `_build_m3_progress` outside the section table and the tests; `build_report`'s `isinstance(result, tuple)` normalisation already handles both shapes.

**The tests are the right shape.** Each asserts the correct label present *and* the wrong label absent (`"artifact rate high" not in text` alongside `"rail" in text.lower()`). A test that only checked for the expected label would still pass with the labels crossed — which is the exact defect. All five required cases present.

**Good judgment call on the `ceiling is None` branch.** Changing `rail_ok = True` to leaving `rail_status` at `UNKNOWN` is behaviourally identical for the verdict but semantically honest: not-evaluated is not the same as passing. That distinction is the whole point of the flags contract.

Scope held to `m3`, stale #911 sentence removed, no `graphify-out/` in the diff (+121/-24, two files), deploy section filled and correct. CI green on `02b1504`.

---

**[nitpick, not blocking]** The two new labels are prefixed `M3 ` ("M3 ladder mass leaking", "M3 high rail vs structural ceiling") while the third keeps the legacy "artifact rate high". Defensible for continuity with previous emails, but inconsistent. Not worth a round-trip.

**[suggestion → filed separately]** `UNKNOWN` never reaches the verdict, so a section printing "(no ladders in 24h -- cannot assess)" still yields "[OK] Healthy -- all systems nominal". That is pre-existing and not made worse here, but the flags contract makes it newly cheap to fix, and absent data reading as healthy in the verdict is the #913 pattern. Filed so it is not lost.

Merging.

PR #975, opened 2026-08-11
Handle UNKNOWN verdict state for unassessed indicators (fixes #974)

### LINKED ISSUE #974: Health report verdict ignores UNKNOWN indicators — 'cannot assess' can still read as 'all systems nominal'
## Context

Raised as a `[suggestion]` during review of PR #973 (#972). Pre-existing, not
introduced by that PR — filed so it is not lost now that the fix is cheap.

#972 migrated the `m3` section to the flags contract, so each indicator now
reports `OK` / `WARN` / **`UNKNOWN`**. `_build_verdict` acts on `WARN` only:

```python
m3_flags = flags_by_section.get("m3", {})
if m3_flags.get("mass_status") == "WARN":
    warn_issues.append("M3 ladder mass leaking")
...
```

`UNKNOWN` — the state set when an indicator could not be evaluated at all — is
silently ignored. So a report can print:

```
  Ladder mass:   (no ladders in 24h -- cannot assess)
  Rail / artifact: (no brackets in 24h -- cannot assess)
...
Verdict
  [OK] Healthy -- all systems nominal, M3 accruing on track
```

The section says it cannot assess anything; the verdict says all systems nominal.

## Why it matters

This is the #913 pattern: absent data reading as healthy. That issue is on record
as *"'Healthy' printed after two days of silent data loss."* The individual
section lines are honest — a reader who reads the whole report sees "cannot
assess" — but the verdict line is the part people actually read, and it is the
part that says nothing is wrong.

The existing `if "unavailable" in all_text` catch does **not** cover this: that
string is the query-*error* fallback, whereas these paths return normally having
found no data. A successful query over an empty window and a crashed query are
different states, and only one of them is currently surfaced.

Low severity — nothing is being computed wrongly, and it requires an empty
window to trigger. But the M3 gate runs on this window within ~2 weeks, and
"no data" is exactly the condition that must not read as green during it.

## Acceptance criteria

- [ ] An `UNKNOWN` indicator produces a distinct verdict state — not `WARN`
      (nothing is known to be broken) and not silent. Something in the
      `[UNKNOWN] / cannot assess` family, naming which indicators were not
      evaluated.
- [ ] A section where *every* indicator is `UNKNOWN` never yields
      `[OK] Healthy -- all systems nominal`.
- [ ] A genuinely healthy section still reads `[OK] Healthy` — this must not
      become a new permanent WARN, which is the failure #969 just removed.
- [ ] The distinction between "query errored" (`unavailable`) and "query
      succeeded, no rows" is preserved and visible; they are different faults.

## Technical notes

`_build_verdict` in `src/scripts/daily_health_report.py`. The `m3` and `bot`
sections are on the flags contract; the other five still line-scan, so a general
`UNKNOWN` treatment only reaches migrated sections for now. That is fine —
implement it for the migrated ones and let the remaining sections pick it up as
they migrate (the docstring already flags that as a fast-follow).

Consider whether `UNKNOWN` on *some* indicators while others are `OK` deserves
different treatment from *all* `UNKNOWN` — a partially-measurable window is a
weaker signal than a completely empty one, and conflating them would either
over- or under-warn.

## Test requirements

- All indicators `UNKNOWN` → verdict is not `[OK] Healthy`, and names what could
  not be assessed.
- Mix of `OK` and `UNKNOWN` → verdict reflects the unmeasured part without
  claiming a fault.
- All indicators `OK` → still `[OK] Healthy`, no regression.
- Query-error path (`unavailable`) still produces "partial data only" and remains
  distinguishable from the empty-window path.

## Dependencies

Builds on #972 / PR #973 (merged). No blockers.

P0K5:
## Summary

Adds a third issue bucket (`unknown_issues`) to `_build_verdict` to handle indicators that cannot be evaluated (UNKNOWN status). Implements the precedence rule CRIT > WARN > UNKNOWN > OK. When only unassessed indicators exist, the verdict now reads `[UNKNOWN] Cannot fully assess` instead of silently passing through.

The fix also explicitly sets `rail_status` to "UNKNOWN" when no ceiling data is available for comparison, ensuring the flag accurately reflects whether an indicator was assessed.

## Changes

- Added `unknown_issues: list[str] = []` bucket in `_build_verdict()`
- Check for UNKNOWN status on all three m3 flags (mass_status, rail_status, gaps_status)
- Append distinct labels per UNKNOWN indicator (e.g., "M3 ladder mass not assessed")
- Implement precedence: CRIT > WARN > UNKNOWN > OK
- Preserve distinction between query errors (unavailable) and empty windows
- Set rail_status to UNKNOWN explicitly when ceiling is unavailable

## Test plan

- All existing 63 tests pass
- Added 5 new tests covering:
  - All three m3 flags UNKNOWN → `[UNKNOWN] Cannot fully assess`, naming all three
  - One UNKNOWN, others OK → `[UNKNOWN]` verdict, naming only the unassessed one
  - One WARN + one UNKNOWN → `[WARN]` tag wins, both labels appear
  - All three OK → still `[OK] Healthy` (regression guard)
  - Query-error path (`unavailable`) yields "partial data only" and is distinguishable from empty-window path

## How to deploy

Code-only change, `git pull` on the host. The health report is a `Type=oneshot` systemd unit that runs fresh each time, no restart needed.

**Post-deploy assertion:** Run the health report on the host; verify:
1. When M3 query returns empty (no ladders/brackets): verdict shows `[UNKNOWN] Cannot fully assess` with specific "not assessed" labels
2. When M3 query errors: verdict shows `[WARN] Stable -- watch: partial data only` (distinct from empty window)
3. When M3 all OK: verdict still shows `[OK] Healthy -- all systems nominal`

Closes #974


P0K5:
## Post-merge review record (the approval could not be posted before merge)

**This PR merged with no review recorded.** My `--approve` call was rejected with `Can not approve your own pull request` — this PR was authored as **P0K5**, the same account the Tech Lead PM reviews under, unlike #971 and #973 which were authored as **ClaudioSenior** and carry proper `CHANGES_REQUESTED` / `APPROVED` records. The merge then succeeded, so the governance gap is real and is recorded here rather than papered over.

**The review itself was done in full before the merge** — it simply had nowhere to land. Contents, for the record:

**Verified beyond the diff:**
- **Nothing downstream parses the verdict tag**, so introducing a fourth value (`[UNKNOWN]`) breaks no consumer. The email subject is a fixed date string; the `{"OK": "[OK]", ...}` dict at line 244 belongs to the bot section, not the verdict line. A new enum value falling through an exhaustive map elsewhere is exactly the failure that only shows up in production, so this was checked rather than assumed.
- **The error path stays distinguishable from the empty-window path** — the requirement most at risk. On a query exception the flags stay `UNKNOWN` *and* `"unavailable" in all_text` fires, giving `[WARN] Stable -- watch: partial data only, M3 ladder mass not assessed, ...`. WARN correctly outranks UNKNOWN and both causes stay visible.
- **The partial case works via the early return**: with `indicators is None`, `mass_status` is already OK/WARN while rail and gaps stay `UNKNOWN`, so a partially-measurable window names only what it could not assess.
- Tests assert `"[OK] Healthy" not in text` alongside the expected label — a precedence bug passes a test that only checks the expected string is present, so the negative assertion is what pins this.

The extra `flags["rail_status"] = "UNKNOWN"` in the `ceiling is None` branch is redundant (the dict initialises to `UNKNOWN`) but better style — explicit at the point of decision, and it survives someone reordering the initialisation.

**Verdict: the code is sound and I stand behind the merge on its merits.** What was wrong is the process, not the change.

**Two process defects to fix, filed separately:**
1. The junior-dev agent authored under **P0K5** instead of the operational dev account, which silently disabled the review gate. #971/#973 (mid-dev) authored correctly, so this is agent-specific, not a repo-wide misconfiguration.
2. The board was left at **In progress** rather than moving to **In review** when the PR opened.

cc #974

Issue #977, opened 2026-08-11
Entry gate lets a second order onto a held market, and reconciliation leaves unclearable zero-share positions


P0K5:
## What happened

Live KORD, 2026-08-11. `open_positions` holds **two rows for the same `token_id`**, one of which is a phantom:

| id | trade_id | ticker | shares | € | entry_ts |
|---|---|---|---|---|---|
| 435 | 2161 | `KORD-order-0x418b18` | 6.67 | 5.00 | `11:47:31.021528Z` |
| 436 | 2160 | `0x0f6d8e984603d5c9…` | **0.0** | 0.00 | `11:41:52.086584+00:00` |

Both carry `token_id = 101370671429372224220389749572993685064650155630486212084462717077923353473008`.

The existing guard in `Database.add_open_position` fired correctly:

```
2026-08-11T12:53:01 WARNING [src.data.db] [open_positions] token_id=10137067142937...
  already has 1 open row(s) -- inserting another (order_id=0x6f780a...).
  The run.py entry gate should have blocked this before order placement;
  investigate if it did not.
```

**It did not.** That is defect 1.

## Two distinct defects

**1. The entry gate did not block a second order on a market already held.** The guard is a warn-and-insert backstop, not a block — it says so, and it names `run.py` as the place the check should have happened. Its previous occurrence was 2026-07-24 on a different token, so this is not a one-off.

**2. Reconciliation registers unfilled orders as zero-share positions, and nothing can ever clear them.** Row 436 was written at **12:53:01** carrying an `entry_ts` of **11:41:52** — a 72-minute lag, i.e. a backfill, not a live entry. There was a power outage on the host that morning; the bot restarted and restart reconciliation re-registered order `0x6f780a`, which had never filled, as a position with `shares = 0.0`.

`Database.close_position` deletes by `order_id`, and that order has no fill to close against. `settle.py:524` gates on

```sql
AND NOT EXISTS (SELECT 1 FROM open_positions p WHERE p.trade_id = t.id)
```

so `trade_id 2160` reads as permanently open.

## Corroborating detail

Two writes with different conventions, which is worth understanding as part of the fix:

- **Timestamp formats differ** — `+00:00` (row 436) vs `Z` (row 435).
- **`ticker` on the funded row is synthetic** — `KORD-order-0x418b18`, derived from the order id rather than the market's condition id, which row 436 has correctly.
[...]

PR #979, opened 2026-08-11
feat(log-rotation): add bot.log rotation with ownership fix

### LINKED ISSUE #978: bot.log grows unbounded (298 MB) and is root-owned on a p0k5 service
## What

`logs/bot.log` is **298 MB** and has no rotation of any kind, on a host that runs live trading.

```
-rw-r--r--  1 root root 298082508 ago 11 16:28 bot.log
-rw-r--r--  1 root root     14884 ago 11 13:30 archive.log
-rw-r--r--  1 p0k5 p0k5    271774 ago  8 01:02 bracket_evals.2026-08-07.jsonl.gz
```

Two separate problems, both small:

**1. No rotation.** `src/utils/log_rotation.py` handles **JSONL snapshots only** (`rotated_path`, `iter_rotated_jsonl`, `housekeep`) and does not touch `bot.log`. `bot.log` is a systemd `StandardOutput=append:` target with no in-app rotation and no logrotate config. So this is the absence of rotation, not a rotation failure — the JSONL rotation is working correctly and has produced a clean daily `.gz` series without a gap.

**2. Ownership inconsistency.** `bot.log` and `archive.log` are `root:root`; every other file in `logs/` is `p0k5:p0k5`, and `meteoedge.service` runs as `User=p0k5`. It works today because systemd opens the file as root and passes the fd down, so the process writes through a descriptor it could not have opened itself. But anything running *as* `p0k5` cannot open those two files for writing, which will block any rotation implementation that does.

## Why it matters (modestly)

Unbounded growth on a live trading host is the main thing. Secondary: a 298 MB single file makes the audit trail awkward to search, and any tooling that reads it whole is doing 298 MB of I/O.

Concretely, it already cost an investigation. The 2026-08-11 power outage left **138 contiguous NUL bytes** at offset 295,266,138 — one interrupted write, 0.00005% of the file, harmless in itself. But a single NUL makes plain `grep` treat the whole file as binary and **silently stop printing matches**. That dropped the `[open_positions]` warning from a search during the #977 investigation and led to a wrong conclusion until the search was re-run with `grep -a`.

## Acceptance criteria

- [ ] `bot.log` is rotated on a bounded schedule or size, with retention consistent with the JSONL snapshots' `SNAPSHOT_RETAIN_DAYS`.
- [ ] Ownership is consistent with the service user; `root:root` log files are not left behind by `sudo` runs.
- [ ] Rotation is safe against systemd's held file descriptor. **Truncating in place (`: > bot.log`) must not be used** — the service holds an fd at the current offset, so the next write lands past it and the kernel zero-fills the gap, producing a genuinely sparse, genuinely corrupt file. Move-then-reopen, `copytruncate` with a signal, or logrotate is required.
- [ ] Any rotation of the *current* 298 MB file is scheduled deliberately: the service must reopen its stream, and a restart is not free while a live position is open (see #977).

## Notes

- **No M3 impact.** `bracket_evals.*.jsonl` rotation is healthy and unaffected; the clean-data window and the ~2026-08-24 gate date do not depend on `bot.log`.
- Until this lands, **use `grep -a` on `bot.log`** — plain `grep` will silently return nothing useful past the NUL run.
- Related: #944 (nothing queries systemd unit health) — same theme of monitoring surfaces being the unreliable part.

ClaudioSenior:
## Summary

Implements safe, systemd-compatible plain-text log rotation for bot.log alongside existing JSONL rotation. Fixes unbounded growth (298 MB) and root:root ownership inconsistency.

## What changed

- **New functions in `src/utils/log_rotation.py`:**
  - `rotate_plaintext_log()`: Rotate plain-text logs using move-then-reopen pattern (safe for systemd's held fd)
  - `housekeep_plaintext()`: Compress/delete aged plain-text logs
  - `_fix_file_ownership()`: Fix ownership from root:root to p0k5:p0k5

- **Retention policy:**
  - Uses `SNAPSHOT_RETAIN_DAYS` (365 days default) to match audit-trail retention
  - Compresses files after 1 day, deletes after 365 days

- **Tests:** 9 new tests covering rotation, housekeeping, ownership, edge cases

- **Deploy runbook:** `docs/DEPLOY_BOT_LOG_ROTATION.md` with systemd timer setup, troubleshooting

## Why

- bot.log grows unbounded on live host (currently 298 MB)
- Ownership inconsistency (root:root vs p0k5:p0k5) blocks any rotation implementation that runs as p0k5
- NUL bytes in file cause plain grep to silently fail (investigate workaround cost)

## How to test

\`\`\`bash
# Run the full test suite (41 tests)
pytest src/tests/test_log_rotation.py -v

# Test locally without systemd
python3 -c "
from pathlib import Path
from src.utils.log_rotation import rotate_plaintext_log, housekeep_plaintext

# Create and rotate a test log
test_log = Path('logs/test_bot.log')
test_log.parent.mkdir(exist_ok=True)
test_log.write_text('test data\n')

rotated = rotate_plaintext_log(test_log)
print(f'Rotated to: {rotated}')
print(f'Original exists: {test_log.exists()}')
print(f'Dated file exists: {rotated.exists()}')

# Housekeep
housekeep_plaintext(test_log)
print('Housekeeping done')
"
\`\`\`

## How to deploy

**SCOPE BOUNDARY (from issue #978):**
- This PR delivers rotation mechanism + ownership fix + runbook (all mandatory prerequisites)
- Does NOT rotate the existing 298 MB bot.log (sequenced with #977 — live KORD position resolution)
- Does NOT restart meteoedge.service (sequenced after #977)
- The runbook will be executed by Tech Lead PM after #977 closes positions

**Steps in runbook:**
1. Systemd service must be running (verify)
2. Run rotation script (one-time rotation of existing 298 MB file OR scheduled timer)
3. Signal systemd to reopen log stream (SIGHUP → meteoedge)
4. Verify new log file is writable as p0k5:p0k5

See \`docs/DEPLOY_BOT_LOG_ROTATION.md\` for full details, systemd timer config, troubleshooting.

## M3 Impact

**None.** bot.log is stdout from systemd. No impact on bracket_evals JSONL rotation (health, clean and active), data collection, or #977's gate window (2026-08-24, data from 2026-08-06).

## Testing notes

- move-then-reopen pattern proven safe (no file truncation, systemd fd preserved)
- Ownership fix tested with mock os.chown
- Retention logic (365 days default) tested with patched _today_utc

Closes #978

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>


P0K5 (changes requested):
## Request changes

Good instincts on the core mechanism — the move-then-reopen shape, reusing `_dated_path`/`_compress`, and mirroring the JSONL housekeeping conventions are all the right call, and the test suite is genuinely thorough for the functions as written. CI is green on `56597731`, all four checks. But green CI is necessary, not sufficient: this PR touches systemd and deploy sequencing, so it gets a manual deep-dive, and there are five blockers. Four are mechanical; #2 is the important one.

### `[blocker] 1 — Nothing invokes the rotation. It cannot run.`

`rotate_plaintext_log()` and `housekeep_plaintext()` have **zero callers anywhere in the repo** — I checked the whole tree, not just the diff. The scheduled invoker exists only as a fenced code block inside `docs/DEPLOY_BOT_LOG_ROTATION.md`, so `scripts/rotate_bot_log.py` is not a file that exists and there is no timer unit.

Acceptance criterion 1 is *"`bot.log` is rotated on a bounded schedule or size."* As merged, nothing rotates on any schedule. This is library code with a README, not a rotation.

The repo already has the pattern to copy — `deploy/systemd/meteoedge-purge-retention.service` + `.timer` is almost exactly this job. Commit:
- `scripts/rotate_bot_log.py` (a real file, not a doc snippet)
- `deploy/systemd/meteoedge-rotate-logs.service` + `.timer`, matching the conventions of the units already in `deploy/systemd/`

This stays inside the scope boundary — committing a unit file is not installing or starting it.

### `[blocker] 2 — Move-then-reopen does not reopen, because the target is a systemd append: stream.`

This is the one that matters. `deploy/systemd/meteoedge.service:15`:

```
StandardOutput=append:/home/p0k5/MeteoEdge/logs/bot.log
```

systemd opens that file **itself, at unit start**, and passes the descriptor to the process as fd 1. The bot never opens `bot.log` and holds no path to reopen. So after `shutil.move`, the descriptor follows the **inode**, not the name — the service keeps appending to `bot.2026-08-11.log` indefinitely, and `bot.log` is never recreated. Rotation appears to succeed once, then all subsequent logging silently lands in a file named for a date that never advances.

The docstring states the caller "can signal systemd to reopen its stream," and the PR body's deploy step 3 says `SIGHUP → meteoedge`. **There is no such reopen.** SIGHUP goes to the application, which does not own the descriptor; systemd does not re-open an `append:` target on any signal. Only a unit restart re-opens it.

Fix both the mechanism and the claims about it:
- The runbook must say **restart** (or `copytruncate`), never SIGHUP, and must state plainly that `bot.log` does not come back without the service reopening the stream.
- Correct the `rotate_plaintext_log` docstring — right now it teaches the wrong model to the next reader.
- Have the rotation script verify the outcome: after rotation, if `bot.log` does not exist or is not being appended to, that is a loud failure, not a silent one.

**One factual correction while you're in here,** because it changes which remedies are legitimate. The issue warns that truncation zero-fills a gap because the fd sits at offset 298 MB. That reasoning does not hold for an `append:` target: `O_APPEND` repositions to end-of-file on *every* write, so after truncation the next write goes to offset 0 and no gap is created. The 138 NUL bytes in the current file came from a write interrupted by the power cut, not from truncation. This does **not** license `: > bot.log` — the issue forbids it, that decision stands, and restart-based rotation is cleaner anyway — but it does mean `copytruncate` is genuinely safe here if you prefer it to a restart. Say which you chose and why.

### `[blocker] 3 — Compressed logs are never deleted, so retention is not bounded.`

In `housekeep_plaintext()`:

```python
for path in directory.glob(f"{stem}.????-??-??{suffix}")   # bot.????-??-??.log
```

That pattern never matches `bot.2026-08-11.log.gz`. The flow is: at 1 day, compress and `_safe_remove` the plaintext; from then on only the `.gz` exists — and the glob cannot see it, so the `age_days > retain` branch never fires for it. The `_safe_remove(path + ".gz")` on line 180 only helps in the window where both files exist, which compression has already ended.

Net effect: every rotated `bot.log` is retained **forever** as a `.gz`. On a ~300 MB/day log that is the same unbounded-growth bug this PR exists to fix, one gzip level down.

To be fair to you: this is inherited — you copied it from `housekeep()`, and it is wrong there too. That is not your defect and I am not asking you to fix the JSONL path here; I am opening a separate Backlog issue for it. But it must not ship in the new code. Glob both the bare and `.gz` forms and derive the date from the filename with the `.gz` stripped.

### `[blocker] 4 — `docs/OPERATIONS.md` not updated.`

`OPERATIONS.md` documents `bot.log` in at least four places, including the log inventory table at line 1095 and the logging section at 1158. This PR changes log paths, adds deploy steps, and adds a timer — all of which that document is the canonical home for. A new standalone `DEPLOY_BOT_LOG_ROTATION.md` does not discharge that; per our review rules, unupdated `OPERATIONS.md` on a PR that changes log paths or deployment steps is a blocker.

Update the inventory table for the dated/compressed forms, and cross-link the runbook from the logging section. Consider folding the runbook into `OPERATIONS.md` outright rather than leaving a second place to look.

### `[blocker] 5 — The ownership fix cannot succeed as the service user, and fails quietly.`

`_fix_file_ownership()` calls `os.chown` and downgrades `OSError` to a warning. But `bot.log` is `root:root`, and a process running as `p0k5` **cannot** `chown` a file it does not own — that is `EPERM`, every time. So when the timer runs as `p0k5`, the rotated file stays `root:root`, the function logs a warning, and rotation reports success.

Acceptance criterion 2 is *"`root:root` log files are not left behind."* As written, they are.

Decide and document which user the timer runs as, and make the code honest about it: if ownership cannot be corrected, that is a failure worth surfacing, not a `log.warning` on the way past. Note also that `shutil.move` on a `root:root` file succeeds anyway if `logs/` is writable by `p0k5` — the move is not what breaks, which is exactly why the silent failure is easy to miss.

### `[suggestion]` — logging every other `append:` target has this problem

`deploy/systemd/` has seven other `StandardOutput=append:` units (`archive.log`, `settle.log`, `capture_forecasts.log`, …), and `archive.log` is `root:root` too. They are far smaller so it is not urgent, but the mechanism you are building is general. If `rotate_bot_log.py` takes a list of paths rather than hardcoding one, this is solved once. I am logging this as a Backlog issue either way — do not expand scope here.

### What I am not asking for

The scope boundary holds: still no rotation of the live 298 MB file and no service restart in this PR. Committing the script and the unit files is inside the line; installing them is not.

### Board

#978 moves back to **In progress**. Re-request review once the five blockers are addressed and CI is green again on the new head commit.

P0K5 (changes requested):
## Request changes (round 2)

Real progress. Blockers 1, 3 and 4 are properly fixed: the script and timer units exist as files, the `.gz` retention hole is closed and the date parsing handles the two-suffix case correctly, and `OPERATIONS.md` is updated. Blocker 2 is fixed in the sense that matters — you correctly identified that `copytruncate` is valid against an `O_APPEND` target and the docstring now teaches the right model.

But the switch to copytruncate brought three new defects with it, and one of them will bite on the very first live run. Not approving yet.

### `[blocker] 6 — The rotation reads the entire log into RAM.`

```python
content = log_path.read_bytes()
dated.write_bytes(content)
log_path.write_bytes(b"")
```

`read_bytes()` on the current `bot.log` allocates **298 MB in one contiguous buffer**, then `write_bytes` holds it while writing. Peak RSS for a routine housekeeping job is ~300 MB, on a live trading host, at 00:05 while the bot is polling. That is the whole point of the file this job exists to manage.

Stream it — `shutil.copyfileobj(src, dst)` with an explicit chunk size, or `shutil.copy2` then truncate. Memory then stays flat regardless of file size. The daily files will be small once rotation is running, but the *first* run is the 298 MB one, and that is exactly the run you cannot afford to have go wrong.

### `[blocker] 7 — A second run on the same day destroys the first archive.`

`_dated_path` is keyed on the date alone, and `dated.write_bytes(content)` opens `wb` — it **overwrites**. Two rotations on one calendar day means the second silently replaces the first dated file with only the content accumulated since, and the earlier segment is gone.

This is not hypothetical. The deploy sequence we agreed on is a *manual* one-time rotation of the 298 MB file, scheduled by me after #977. If that manual run lands on the same UTC day the timer has already fired — or if anyone re-runs the script to check it worked — the archive is truncated to whatever accumulated in between. For a file the issue describes as the audit trail, that is data loss with a trivial trigger.

Either append when the dated file already exists, or refuse and exit non-zero, or add a counter suffix (`bot.2026-08-11.1.log`). Any of the three is fine; silently overwriting is not. Cover it with a test that rotates twice in one day and asserts nothing was lost.

### `[blocker] 8 — `bot.log` itself is still `root:root`.`

You chown the *dated* file, but never `bot.log`. `log_path.write_bytes(b"")` truncates an existing inode, so its ownership is untouched — `bot.log` stays `root:root` forever. Acceptance criterion 2 is "root-owned log files are not left behind," and the file named in the issue title is still root-owned after this runs.

Which leads to the design point below, because it is the same root cause.

### `[blocker] 9 — Running the timer as root to work around ownership is the wrong trade.`

`meteoedge-rotate-logs.service` has `User=root`, plus `EnvironmentFile=…/.env`. That means root executing a Python script out of a directory `p0k5` can write, with the service's secrets loaded into a root process, every day. Anyone who can write to the repo as `p0k5` gets code execution as root on the next timer tick. That is a meaningful privilege escalation path introduced to solve a file-ownership annoyance.

You do not need root at all. Chown `bot.log` to `p0k5:p0k5` **once**, as a documented `sudo` step in the runbook, and every subsequent rotation is `p0k5` reading, copying and truncating a file it owns. Then:

- `User=p0k5` on the unit, matching `meteoedge.service` and every other unit in `deploy/systemd/`.
- No root daemon, no `.env` in a root process.
- `_fix_file_ownership` becomes a belt-and-braces check rather than the thing holding the design up — keep it raising, that part is right.

Note that chowning `bot.log` is safe with the service running: systemd already holds its descriptor, and file ownership does not affect an open fd. systemd opens `append:` targets as root at unit start regardless of `User=`, so the next restart still works.

Do this and blocker 8 resolves with it.

### `[suggestion]` — the post-rotation verification cannot fail

`rotate_bot_log.py` checks `if not log_path.exists()` after rotating. Under copytruncate the file is always there — you just truncated it — so the check can never fire and the "service may not have reopened its log stream" message can never print. Harmless, but it reads as a safety net that isn't one. If you want a real check, stat the size a few seconds apart and warn if it is still zero while the unit is active. Low priority; I will not hold the PR on it.

### `[nitpick]`

The comment above the plaintext glob says `bot.log.2026-08-11.log`. Actual form is `bot.2026-08-11.log`.

### `[note]` — the copytruncate race is accepted, say so

Writes landing between the copy and the truncate are lost. That is inherent to copytruncate and standard logrotate behaviour, and at this volume it is a handful of lines once a day — I am not asking you to solve it. But add one line to the runbook stating it is a known, accepted gap, so nobody later reads a one-line hole in the audit trail as corruption.

### Scope and board

Scope boundary unchanged: no rotation of the live file, no restart, no touching the host. #978 stays **In progress**. Re-request review when 6–9 are addressed and CI is green on the new head commit.

P0K5 (approved):
## Approved

CI green on `8b729e78` — `Lint (flake8)`, `Test (pytest)`, `gitleaks`, `AI / NVIDIA NIM review` all success on the head commit I reviewed.

All ten findings across three rounds are resolved and verified against the diff rather than taken on report:

- Rotation is actually invoked — `scripts/rotate_bot_log.py` plus `meteoedge-rotate-logs.{service,timer}`, following the existing `purge-retention` pattern.
- copytruncate against an `O_APPEND` target, correctly reasoned and correctly documented.
- Streaming copy at 64 KB chunks — no 298 MB allocation.
- Same-day re-rotation appends instead of overwriting, with a regression test that fails against the old `wb`.
- `.gz` files are matched by housekeeping, so retention is genuinely bounded.
- Both inodes chowned; `User=p0k5`, no root daemon over a user-writable directory.
- `OPERATIONS.md` updated; runbook troubleshooting no longer prescribes the removed root workaround.

Two things I want on the record for whoever deploys this.

**The live cutover is not done and is not part of this PR.** `logs/bot.log` is still 298 MB and still `root:root` on the host. Step 2 of the runbook — `sudo chown p0k5:p0k5 logs/bot.log` — is a prerequisite, not optional: without it the first timer run fails loudly by design. I am sequencing that after #977 clears the open KORD positions.

**The copytruncate race is accepted, not solved.** Writes landing between copy-start and truncate-end are lost. It is a narrow window once a day and inherent to the pattern, but it means a one-line gap in `bot.log` at ~00:05 UTC is expected behaviour, not corruption. The runbook says so; keep it that way.

Follow-ups filed rather than folded in: #980 (same `.gz` retention hole in the original `housekeep()` for JSONL — inherited, and still live) and #981 (seven other unrotated `append:` targets, blocked on this).

Good work on the iteration — particularly recognising that `O_APPEND` made copytruncate legitimate here rather than taking the issue's framing at face value.

Issue #611, opened 2026-07-03
2026-07-03 [P1] Duplicate live entries stack the same bracket every poll (7x KATL, 3x WMKK) — position dedup depends on stale live_trades.jsonl


P0K5:
## Observed

The bot re-enters a bracket it already holds, roughly once per poll, multiplying the intended flat €5 exposure:

- **2026-07-02 KATL**: 7 live NO orders on the *same* 98.0–99.0°F bracket between 00:01 and 01:04 UTC (trade ids 1139–1145, 7 distinct order_ids) → ~€35 on one outcome.
- **2026-07-03 WMKK**: 3 filled NO orders on 91.4–93.2°F at 00:03/00:13/00:24 (ids 1163/1165/1167) — and `open_positions` holds **three separate rows for the same token_id** — plus 5 timeout attempts on 87.8–89.6°F (ids 1183–1193).

`SIZING_MODE=flat` + `POSITION_SIZE_EUR=5` is supposed to mean one €5 position per opportunity; actual concentration reached 3–7× that on a single bracket.

## Suspected root cause

Candidate evaluation has no "already holding this token/bracket today" gate, and order reconciliation leans on `live_trades.jsonl` history — which is frozen at 2026-06-16 by log rotation (see #609). bot.log shows recurring `[reconcile] wallet token … has no JSONL history — likely manual trade, cannot auto-recover enrichment` warnings for tokens the bot itself bought. With no recognized open position, each poll treats the same flagged bracket as a fresh candidate. `MAX_OPEN_POSITIONS=15` is the only backstop and it does not deduplicate per bracket.

Fix #609 first (DB as source of truth); this issue then adds the explicit guard.

## Fix

1. Entry gate: before `risk_manager.allow_trade`, reject a candidate whose `(station, ticker/token_id, side, day)` already has an open position **in the DB** (`open_positions` join, not JSONL).
2. Deduplicate `open_positions` rows per token (3 rows for one WMKK token today); decide merge semantics for the stacked historical entries.
3. Alert/log when the guard fires, so intentional re-entry (if ever wanted) becomes an explicit config decision rather than an accident.
4. Regression test: two consecutive polls flagging the same bracket must produce exactly one live order; tests must not hardcode calendar dates against today-anchored windows.

## Acceptance criteria

- [ ] One live position max per (station, token, side, day) unless explicitly configured otherwise
- [ ] `open_positions` has no duplicate token rows
- [ ] Guard-fired events visible in logs/dashboard
[...]

PR #983, opened 2026-08-11
fix(execution): close entry-gate and reconciliation gaps on held tokens (#977)

### LINKED ISSUE #977: Entry gate lets a second order onto a held market, and reconciliation leaves unclearable zero-share positions

ClaudioSenior:
## Linked issue
Closes #977

## What changed

Two live-money defects found on live KORD 2026-08-11: two `open_positions` rows on one `token_id`, one a zero-share phantom nothing could close, leaving `trade_id 2160` permanently open to `settle.py`.

**Defect 1 — entry gate let a second order onto a held market.** Root cause: `LiveTrader.place_order()` wrote a synthetic `"<STATION>-order-<id>"` ticker into `trades.ticker` at placement time. The *real* ticker was only corrected later, at the very end of `_attempt()`, via `_append_live_trade()` -> `update_trade_by_order(ticker=...)`. If the process crashes between placement and that correction (a power outage mid-wait, exactly what happened), `trades.ticker` stays synthetic forever, and `Database.has_live_trade_today()`'s ticker-matched query can never find the row. Once the `open_positions` row for the same order is *also* gone (e.g. deleted by a `cancel_order()` that completed locally before the crash), a same-day re-entry on that bracket sails through both gate checks unblocked.

Fix: `LiveTrader.place_order()` now accepts and stores the real bracket ticker (`order_executor.py` passes `candidate.bracket.ticker`), so `has_live_trade_today()` works as designed from the instant of placement -- no dependency on a later correction step surviving a crash. Also added a defense-in-depth check in `run.py`'s entry gate against `order_manager._open_orders` (refreshed from the live exchange every poll), which closes the residual window before either DB check has a row to find at all.

I verified `Database.get_open_position_by_token()` (token_id-based, unaffected by the ticker bug) was already working correctly -- this is why the guard in `add_open_position` fired ("already has 1 open row(s)") only for the phantom row created *after* the second order had already been placed; at placement time nothing had been written yet for the phantom order. **Confirmed NOT the guard in `add_open_position`** -- left that untouched, per the issue.

**Defect 2 — reconciliation creates unclearable zero-share positions.** `OrderManager._reconcile_db_row()` computed `shares` from JSONL fields (`size_matched` / `shares`) that `_append_live_trade()` never writes -- the JSONL schema only carries `size_eur`. This meant every reconciled order got `shares=0.0` regardless of its real fill, an unclearable phantom row (`close_position()` deletes by `order_id`, and a zero-fill order has no fill to close against, so `settle.py`'s `NOT EXISTS` join never clears).

Fix: `_reconcile_db_row()` now takes an optional `get_fill_size(order_id) -> float` (wired to `LiveTrader.get_order_fill_size`, which already existed and queries the exchange for the authoritative matched size). `reconcile_timeout_fills()` now takes `live_trader` and threads it through; `run.py`'s call site passes it. A confirmed zero fill (or no fill-size source at all, e.g. legacy/no live_trader) now skips the `open_positions` insert entirely -- no more zero-share rows, ever.

**#611 vs #977 (issue note 3):** I checked `scripts/merge_duplicate_open_positions.py` (#611) first. It's a *different* path that fix never covered, not a resurfacing of it: #611's merge logic explicitly refuses (flags as MISMATCH) any row with `shares<=0`, so it would have safely refused to touch the live 435/436 rows rather than corrupt them -- but it also had no way to *resolve* a zero-share phantom sitting alongside a real row. Extended it (not rewritten) with a pass that: drops zero-share row(s) for a token that also has a real row, and repairs a surviving row's synthetic `<STATION>-order-<id>` ticker (both `open_positions.ticker` and `trades.ticker`) from a dropped phantom's real ticker when one is available -- exactly the shape of the live 435/436 pair. A token with *only* zero-share rows is still left untouched and flagged (needs a human to confirm the order truly never filled). `--dry-run` previews this exactly like the existing merge; idempotent.

**Timestamp format (issue note, acceptance criterion):** `place_order()` wrote `...isoformat() + "Z"`, reconciliation paths wrote `+00:00` (`datetime.now(timezone.utc).isoformat()`). Normalized centrally in `Database.open_position()` (`_normalize_iso_ts`) to a single `...Z` convention; unparseable input passes through unchanged rather than raising.

**Double-warning / UNIQUE INDEX question (issue note 5):** Confirmed the unique index is NOT silently bypassed. `Database.open_position()` has no internal exception handling around its `INSERT` -- a genuine duplicate `order_id` raises `sqlite3.IntegrityError`, which propagates to whichever caller wraps it: `LiveTrader.place_order()` logs it as `CRITICAL: ... DB write failed`, `_reconcile_db_row()` logs it as `WARNING: DB open_positions insert failed`. Neither swallows it. I could not find a code path that reprocesses the same JSONL record twice -- `rotated_sources()` already explicitly excludes the legacy plain file when it's a symlink into today's rotation, specifically to avoid double-counting. My best-supported read of "logged twice in the same second" is two near-simultaneous `open_position()` calls for the *same token_id* (not necessarily identical `order_id`s) during the defect-1 race window this PR closes -- I could not fully confirm without the raw production logs. Flagging as residual uncertainty rather than guessing further.

## Remediation for the existing live rows (435/436) -- reviewable evidence

**NOT touched by me.** Execution against the live DB is a **post-merge step sequenced by the Tech Lead PM**, because KORD's position is still open (unsettled) -- the row must not be touched while other live automation (take-profit exits, stop-loss, settlement) may still be reading `open_positions` mid-poll. Touching it mid-merge risks colliding with that automation; sequencing it lets the PM pick a moment the bot isn't actively polling that token.

The runbook (exact invocation, verification query) is now in `scripts/merge_duplicate_open_positions.py`'s module docstring under "Runbook: resolving the #977 KORD 2026-08-11 incident" -- findable from the issue via the script the issue itself named.

**Dry-run transcript against a fixture reproducing the exact row shape** (same `token_id`, same `order_id`s `0x418b18` / `0x6f780a`, same tickers, same shares -- built with `Database.insert_trade`/`open_position`, never touching production):

```
$ python -m src.scripts.merge_duplicate_open_positions --dry-run --db-path fixture.db
[merge] token 10137067142937...: [dry-run] would repair synthetic ticker 'KORD-order-0x418b18' -> '0x0f6d8e984603d5c9d3a1e2b7c8f9012345678901234567890123456789012345678901' on open_positions id=1 (trade_id=1)
[merge] token 10137067142937... (KORD): [dry-run] would drop 1 zero-share phantom row(s) id(s) [2] (order_id(s) ['0x6f780a']); 1 positive-share row(s) kept -- the trades row(s) for the phantom's order_id(s) are left untouched (settlement history).

[merge] Summary: 0 token(s) would be merged, 0 duplicate row(s) would be deleted, 1 zero-share phantom row(s) would be dropped, 1 synthetic ticker(s) would be repaired, 0 mismatched
```

Fixture confirmed byte-identical before/after the dry-run (no write). Ran the same invocation *without* `--dry-run` against a **copy** of the same fixture (still not production) to confirm the end state matches what the transcript above promises:

```
$ python -m src.scripts.merge_duplicate_open_positions --db-path fixture_applied.db
[merge] token 10137067142937...: repairing synthetic ticker 'KORD-order-0x418b18' -> '0x0f6d8e984603d5c9d3a1e2b7c8f9012345678901234567890123456789012345678901' on open_positions id=1 (trade_id=1)
[merge] token 10137067142937... (KORD): dropping 1 zero-share phantom row(s) id(s) [2] (order_id(s) ['0x6f780a']); 1 positive-share row(s) kept -- the trades row(s) for the phantom's order_id(s) are left untouched (settlement history).

[merge] Summary: 0 token(s) merged, 0 duplicate row(s) deleted, 1 zero-share phantom row(s) dropped, 1 synthetic ticker(s) repaired, 0 mismatched
```

Resulting state on the applied copy -- exactly one `open_positions` row survives (real shares, repaired ticker on both tables), the phantom's `trades` row is untouched:

```
open_positions:
  {'id': 1, 'trade_id': 1, 'order_id': '0x418b18', 'ticker': '0x0f6d8e984603d5c9d3a1e2b7c8f9012345678901234567890123456789012345678901', 'shares': 6.67}
trades:
  {'id': 1, 'order_id': '0x418b18', 'ticker': '0x0f6d8e984603d5c9d3a1e2b7c8f9012345678901234567890123456789012345678901', 'outcome': None}
  {'id': 2, 'order_id': '0x6f780a', 'ticker': '0x0f6d8e984603d5c9d3a1e2b7c8f9012345678901234567890123456789012345678901', 'outcome': 'filled'}
```

The fixture's synthetic auto-increment ids (1/2, trade_id 1/2) differ from the incident's real ids (435/436, trade_id 2161/2160) since those are DB-assigned and can't be forged meaningfully in a fresh fixture -- everything that determines the script's *behavior* (`token_id`, `order_id`s, tickers, shares) matches the incident exactly.

## How to test
```
python -m pytest src/tests/test_issue_977_entry_gate_reconciliation.py -v   # new, covers every #977 acceptance criterion
python -m pytest src/tests/test_merge_duplicate_open_positions.py -v        # extended with the phantom-drop/ticker-repair cases
python -m pytest src/tests/test_entry_guard.py src/tests/test_order_manager.py src/tests/test_live_trader_db.py src/tests/test_order_executor_reprice_743.py -v
python -m pytest src/tests -q   # full suite: 3132 passed, 33 pre-existing failures identical on master (Windows-only env issues: file locks, symlink perms, worktree/network-dependent tests) -- none touch the files this PR changes
python -m flake8 src/           # clean
```

## How to deploy
- [x] **Code only** -- `git pull` + restart the live trading service (`run.py --live`). No new env vars, no schema change (the `open_positions`/`trades` schemas are unchanged; only write-path values/formats changed).
- [ ] Dependencies changed
- [ ] DB schema change
- [x] **One-off script(s)** to run after deploy, **PM-sequenced, operator-run only, do not run in CI** -- full runbook in `scripts/merge_duplicate_open_positions.py`'s module docstring ("Runbook: resolving the #977 KORD 2026-08-11 incident"), reviewable dry-run transcript above:
  ```
  python -m src.scripts.merge_duplicate_open_positions --dry-run --db-path /path/to/meteoedge.db   # preview first
  python -m src.scripts.merge_duplicate_open_positions --db-path /path/to/meteoedge.db             # resolves open_positions id 435/436:
                                                                     #  - drops the zero-share phantom (trade_id 2160)
                                                                     #  - repairs trade_id 2161's synthetic ticker to the real condition id
  # verify:
  sqlite3 /path/to/meteoedge.db "SELECT id, trade_id, order_id, ticker, shares FROM open_positions WHERE token_id='101370671429372224220389749572993685064650155630486212084462717077923353473008'"
  ```
- [ ] Config / env change
- [x] **Post-deploy validation** -- watch for `[reconcile] ... has no confirmed fill -- skipping open_positions insert` (defect 2 fix engaging correctly instead of the old silent zero-share insert) and confirm no new `[open_positions] token_id=... already has N open row(s)` warnings after the deploy; `SELECT COUNT(*) FROM open_positions GROUP BY token_id HAVING COUNT(*)>1` should return 0 rows once the merge script has run.
- [ ] No deploy needed

## Checklist
- [ ] Project board status updated (In review) -- **blocked**: my GraphQL quota is exhausted; PM is tracking Ready->In progress and will send `ITEM_ID_ISSUE_977` for the In-review transition
- [ ] Docs updated -- not applicable (no new env vars/schema/API/run-mode)


P0K5 (changes requested):
## Request changes — one blocker, and it's a small one

This is a strong PR. The root-cause work is the best part of it and I want to be specific about why: you did not accept the issue's framing. The issue said the entry gate in `run.py` failed; you traced *why* it could fail and found the actual mechanism — `place_order()` writing a synthetic `<STATION>-order-<id>` into `trades.ticker`, which makes `has_live_trade_today()`'s ticker-matched query structurally incapable of ever matching a live order's own row. A crash between placement and `_append_live_trade()` — exactly what the power outage produced — leaves that row synthetic permanently. That is a real diagnosis, it explains the observed 11:47 entry, and it is not what the issue told you to look for.

The reconciliation half is equally well found: `_reconcile_db_row` computing `shares` from `size_matched`/`shares` fields that `_append_live_trade()` never writes, so the value was structurally always `0.0`. Sourcing it from `LiveTrader.get_order_fill_size` and refusing to insert on a zero or failed lookup is the right fix, and defaulting to "write nothing" when no `live_trader` is supplied is the right failure direction.

The merge script extension is careful in the way this kind of script has to be: it drops phantoms rather than merging them, repairs the synthetic ticker on **both** `open_positions` and the `trades` row, and — the part I liked most — refuses to touch a token that has *only* zero-share rows, reporting it for a human instead. That is the correct instinct: an unexplained zero-share row might be the only thing tracking a real position.

### `[blocker] — the synthetic ticker fallback is retained, and a test locks it in`

```python
resolved_ticker = ticker or f"{station}-order-{order_id[:8]}"
```

Acceptance criterion 4 is explicit: *"`ticker` is populated with the market's condition id on every write path, **never** a synthetic `<STATION>-order-<id>` placeholder."*

`place_order()` has exactly **one** caller — `order_executor.py:173` — and your change makes it pass the real ticker. So the fallback is unreachable today. Its only function is to guarantee that the next caller added without a `ticker=` argument silently reintroduces this exact bug, with no error, no warning, and a failure mode that only shows up as a duplicate live position days later. That is precisely how this defect got here.

The docstring even documents the hazard ("Always pass the real ticker from the caller") — a comment asking future callers to be careful is not a gate. Make it structural: drop the `ticker: str = ""` default so omission is a `TypeError`, or raise `ValueError` when it is empty. Either is three lines.

And `test_place_order_falls_back_to_synthetic_ticker_when_omitted` currently **asserts the forbidden behaviour is preserved**. That test has to invert — assert that omitting the ticker raises. As written, it would keep passing while the criterion stays unmet, which is worse than no test.

### `[question] — is `LIVE_ALLOW_BRACKET_REENTRY` false on the live host?`

`src/config.py:272` defaults it to `false`, so the `has_live_trade_today()` branch is live by default. But it is env-driven and `.env` is not in the repo, so I cannot verify the running value. If it is `true` in production, your primary fix sits behind a disabled branch and only the `_open_orders` and `get_open_position_by_token` guards apply.

Confirm the value on the host before this deploys. Not a code change — I need the answer, not an edit.

### `[suggestion]` — reaching into `order_manager._open_orders` from `run.py`

The defense-in-depth check is well justified and the comment explaining *why* the DB can lag the exchange after a restart is genuinely useful. But `run.py` now reaches into a private attribute across a module boundary, so any refactor of `_open_orders`' shape silently breaks a live-money guard rather than failing loudly. Add a small public accessor (`order_manager.has_open_order(token_id)`) and call that. Logging as a Backlog issue rather than holding the PR — tell me if you would rather fold it in now, it is a two-line change.

### `[question] — `graphify-out/` in this PR`

The diff is +199,824/−191,033, of which ~196k lines are regenerated `graphify-out/graph.json` and friends. `CLAUDE.md` does tell you to run `graphify update .` after code changes, so I am not treating this as a mistake. But the repo also has a CI job producing `chore(graph): incremental graph update [skip ci]` commits, which suggests the graph is maintained automatically and does not need to ride along.

The cost is real: it buries a 200-line live-money change inside a 390,000-line diff, which is hostile to exactly the manual review this PR most needs. Confirm which convention is correct — if CI maintains it, drop `graphify-out/` from this branch.

### Verified, for the record

- Root-cause chain confirmed by reading `has_live_trade_today()` (`db.py:1669`) — it matches on `ticker`, so the synthetic value could never match. The diagnosis holds.
- `_normalize_iso_ts` — `dateutil.parser` and `timezone` are both imported (`db.py:6,10`); no NameError. Failure path returns input unchanged rather than raising, which is right for a position write.
- `_is_synthetic_ticker`'s regex matches the real observed form `KORD-order-0x418b18`.
- Test coverage maps onto the issue's five required tests, including both the zero-fill and positive-fill reconciliation cases and the "no `live_trader`" path.
- CI green on `84c0fea8`: `Lint (flake8)`, `Test (pytest)`, `gitleaks`, `AI / NVIDIA NIM review`.
- M3: nothing here touches `bracket_evals` or the probability path. Cleared under the `docs/REMEDIATION_PLAN.md` test.

### Board

#977 stays **In review**; I am not moving it back for one three-line change plus a test inversion. Push those, answer the `LIVE_ALLOW_BRACKET_REENTRY` question, and I will approve on green CI. The live 435/436 repair still does not run until after merge, and I will sequence it.

ClaudioSenior:
Addressing review feedback:

**Blocker fixed.** `LiveTrader.place_order()`'s `ticker` param no longer has a default -- omission now raises `TypeError` before `create_and_post_order()` is ever called (verified via test: exchange mock asserted not-called), and an explicit empty string raises `ValueError`. The synthetic `<STATION>-order-<id>` fallback is gone from the source entirely, not just unreachable. Updated the one production call site (already had `ticker=`, unaffected) and every test call site; inverted `test_place_order_falls_back_to_synthetic_ticker_when_omitted` into `test_place_order_omitted_ticker_raises_instead_of_falling_back` (+ a new empty-ticker test), and rewrote `test_synthetic_ticker_defeats_has_live_trade_today` to seed the DB directly since `place_order()` can no longer produce that shape itself.

**`LIVE_ALLOW_BRACKET_REENTRY` on the live host: I can't confirm it, and I'm not guessing.** I checked every source available to me: `src/config.py:272` defaults to `false`; `docs/OPERATIONS.md` documents `false` as the default and describes check 3 (the branch this fix's ticker correction feeds) as gated on it; `deploy/systemd/*.service` files all load `EnvironmentFile=/home/p0k5/MeteoEdge/.env`, which is correctly not in the repo. I have no access to that host or that file, so I cannot verify the actual production value -- flagging rather than assuming. Worth noting even if it *is* `true` in production: the `open_positions.token_id` check (defect-1's other, independent leg -- "an open position already exists for this token_id... always blocks, in every configuration") and the new `order_manager.has_open_order()` defense-in-depth check are both unconditional on this flag, so the fix isn't entirely behind it either way -- but the ticker-correction's specific contribution to check 3 is.

**`graphify-out/` dropped from the branch** -- confirmed `.github/workflows/graphify-update.yml`'s `incremental-update` job runs `on: push: branches: [master]` and commits `graphify-out/graph.json` directly (`chore(graph): incremental graph update [skip ci]`), so it's auto-maintained and doesn't need to travel in the PR. Diff is now 9 files / ~800 lines.

**Non-blocking `has_open_order()` accessor** -- folded in now rather than filing to backlog, it was genuinely two lines: `OrderManager.has_open_order(token_id)` (lock-guarded read of `_open_orders`), `run.py`'s entry gate calls it instead of touching the private set.

Full suite re-run clean: targeted tests (`test_issue_977_entry_gate_reconciliation.py`, `test_live_trader_db.py`, `test_entry_guard.py`, `test_order_manager.py`, `test_order_executor_reprice_743.py`, `test_merge_duplicate_open_positions.py`, `test_log_rotation.py`) all pass; full `src/tests` run: 3146 passed, same 34 pre-existing-on-master environment-only failures (Windows file locks/symlink perms/worktree+network tests), 0 new. `flake8 src/` clean. Pushing now -- will confirm CI green here.

ClaudioSenior:
Flagging a conflict rather than resolving it myself: \`CI / lint\` and \`CI / test\` are green on \`80827bb\`, but the automated \`AI / NVIDIA NIM review\` check posted **BLOCK** (check-run [93853026495](https://github.com/P0K5/MeteoEdge/runs/93853026495), [comment](https://github.com/P0K5/MeteoEdge/pull/983#issuecomment-5256037278)):

> Unmet acceptance criterion: existing live rows not resolved. [...] does not include or invoke a script to remediate the two specific existing rows in the live database.

That's a literal reading of the acceptance-criteria bullet, and it's in direct tension with your own instructions: the original assignment said *"Ship the script plus a documented invocation; I will sequence the live run"* and do **not** run it against live data myself, and your review comment said *"The live 435/436 repair still doesn't run until after merge; I'm sequencing that"* — which read to me as accepting the deferred-execution design.

Per my role instructions I'm not supposed to ignore an AI-reviewer finding, but I also don't think re-litigating your own explicit direction is mine to decide unilaterally. Two ways I can see to resolve this, your call:
1. You override/re-run the review (or approve past it) since the deferral was your instruction, not a gap I introduced.
2. Tell me what would satisfy the checklist item without me running the migration against production — e.g. a CI-safe dry-run demonstration against a synthetic DB fixture reproducing the 435/436 shape (I already have that exact fixture in \`test_merge_duplicate_open_positions.py::_seed_phantom_scenario\`), wired into the PR description or a script invocation log, if that's the kind of evidence you'd want the reviewer (or a future reader) to see without me touching live data.

Not proceeding further on this point until you weigh in.

ClaudioSenior:
Addressing review feedback: added the reviewable remediation artifact per @P0K5's direction (\`3a38351\`):

1. **Runbook** — \`scripts/merge_duplicate_open_positions.py\`'s module docstring now has a "Runbook: resolving the #977 KORD 2026-08-11 incident" section: exact invocation (dry-run first, then apply, then a verification query), scoped to the specific \`token_id\` from the issue.
2. **Dry-run transcript** — in the PR body, against a fixture reproducing rows 435/436's exact shape (same \`token_id\`, \`order_id\`s \`0x418b18\`/\`0x6f780a\`, tickers, shares). Also ran the apply step against a *copy* of that fixture (never production) and pasted the resulting state, so the transcript's promise is independently checkable.
3. **Explicit deferral statement** — PR body now states plainly that execution against live data is a PM-sequenced post-merge step, and why: KORD's position is still open, so the row shouldn't be touched while other live automation (take-profit/stop-loss/settlement) may be reading it mid-poll. That context previously only lived in our review thread.

CI/lint and CI/test were already green; re-triggering AI / NVIDIA NIM review now on \`3a38351\`.

ClaudioSenior:
Update: the AI / NVIDIA NIM review's first run on \`3a38351\` actually errored (\`AI reviewer failed: NIM API 404 Not Found\`) rather than re-issuing a content BLOCK — an infra flake, not a review finding. Re-ran the failed workflow (\`gh run rerun 31616773871 --failed\`) rather than treating it as a second content-based iteration against your two-strikes guidance. It came back clean on retry:

> VERDICT: PASS — all six acceptance criteria, including the scripted-resolution one, checked off.

All checks now green on \`3a38351\`: \`CI / lint\`, \`CI / test\`, \`AI / NVIDIA NIM review\`, \`gitleaks\`.

P0K5 (approved):
## Approved

CI green on `3a38351` — `Lint (flake8)`, `Test (pytest)`, `gitleaks`, and `AI / NVIDIA NIM review` (verdict PASS). Verified on the head commit, not assumed.

**On the earlier BLOCK.** The reviewer's checklist contradicted itself, but the substance was right and worth having: the script existed, the *resolution* was not reviewable. What came back closes that properly — a runbook keyed to this incident's `token_id`, a dry-run transcript against a fixture reproducing the real `order_id`s / tickers / shares, and the applied-to-a-copy end state showing exactly one surviving row with the ticker repaired on both `open_positions` and `trades`. That is the artifact I needed as the person who has to run this against live money, and I would not have had it if the gate had been overridden.

Three things I want to credit specifically:

- **The root cause was not the one the issue named.** The issue said the `run.py` gate failed; you found *why* it could fail — `place_order()` writing a synthetic ticker into `trades.ticker`, making `has_live_trade_today()`'s ticker-matched query structurally unable to match a live order's own row, permanently so after a crash between placement and `_append_live_trade()`. That is the difference between a fix and a patch.
- **`--dry-run` honesty.** Confirming the fixture was byte-identical before/after the dry-run, and stating plainly that the fixture's auto-increment ids differ from 435/436 while everything governing behaviour matches, is exactly the right level of care for a script that mutates live position state.
- **Residual uncertainty declared, not papered over.** The double-warning / UNIQUE INDEX question came back as "best-supported read, could not fully confirm without production logs" rather than a confident guess. Correct call. The reasoning — that the index is not bypassed because `open_position()` wraps no exception handler and callers log rather than swallow — is sound and I checked it.

**Follow-up filed, not folded in:** the script still operates table-wide, and the runbook's mitigation is "read the dry-run, STOP if any other token appears." That is adequate for a supervised run and I will follow it, but a `--token-id` scoping flag would make the blast radius structural instead of procedural. Logged separately.

**Deploy sequencing — mine to run, not to be executed by anyone else:**

1. Merge this.
2. Pull on the host, with the service still running old code.
3. Stop `meteoedge.service`, so nothing is reading `open_positions` mid-poll.
4. Run the runbook: dry-run, inspect for exactly one token and nothing else, then apply.
5. Start the service — now on the fixed code, so restart reconciliation cannot recreate a phantom.

Step 5 after step 2 matters: restarting on the *old* code is what created row 436 in the first place. Then #978's rotation deploy follows, which needs no further restart.

Note this PR's "How to deploy" says restart the live service. That restart is not free — KORD's position is open — which is why it is sequenced above rather than done ad hoc.

If the `[open_positions] ... already has N open row(s)` warning ever appears again after this ships, that falsifies the race-window explanation for the double-logging and should reopen the question rather than be treated as noise.

Issue #976, opened 2026-08-11
54 of the last 60 merged PRs have zero recorded reviews — the review gate is not enforced


P0K5:
## Context

PR #975 (issue #974) was merged to `master` with **zero reviews recorded**. The
Tech Lead PM's `gh pr review --approve` call was rejected:

```
failed to create review: GraphQL: Review Can not approve your own pull request
```

because the PR was authored as **P0K5** — the same account the PM reviews under.
The subsequent `gh pr merge` succeeded, so nothing stopped the merge.

This defeats governance invariant #4 (*"No PR is approved or merged unless CI is
green"* — and, implicitly, that a PR is reviewed at all) without any error,
warning, or failed check. The only signal was a non-fatal message in the middle
of otherwise-successful command output.

## It is agent-specific, not repo-wide

| PR | Issue | Agent | Author | Reviews recorded |
|---|---|---|---|---|
| #971 | #969 | mid-dev | `ClaudioSenior` | `CHANGES_REQUESTED`, `APPROVED` |
| #973 | #972 | mid-dev | `ClaudioSenior` | `APPROVED` |
| **#975** | **#974** | **junior-dev** | **`P0K5`** | **none** |

Both mid-dev runs authored correctly under the operational account and their
reviews registered normally. The junior-dev run in the same session, on the same
machine, authored as the supervisor account. So the repo and CI configuration are
fine — something differs in how `junior-dev` resolves its git/gh credentials.

## Why it matters

The review gate is the last thing between a subagent's output and `master`. When
it fails, it fails **silently and open**: the PM writes a full review, the API
rejects it, the merge proceeds, and the merged commit looks identical to a
properly-reviewed one. Nothing in the repo afterwards shows the difference except
an empty reviews list that nobody queries.

The code in #975 was reviewed in full and is sound — this is not about that
change. It is that the same sequence would have merged an unreviewed *bad*
change just as quietly.

## Acceptance criteria

- [ ] `junior-dev` authors commits and PRs under the operational account, matching
      `mid-dev`'s behaviour. Root-caused, not worked around by having the PM push.
- [ ] A PR authored by the reviewing account **cannot be silently merged** — either
      the merge is blocked, or the failure to approve is surfaced as a hard error
      rather than a line of stderr in a successful command.
- [ ] The difference between the two agent types is identified and documented, so
      it does not silently regress when agent definitions change.
- [ ] Retroactive check: query all merged PRs for an empty `reviews` list and
[...]

PR #1007, opened 2026-08-13
Retry CLOB health check with backoff instead of exiting on first failure

### LINKED ISSUE #1003: Retry CLOB health check with backoff instead of exiting on first failure
## Context

`src/scripts/run.py:919` aborts the process the moment `check_clob_health()` returns
False:

```python
log.info("Checking CLOB connectivity...")
if not check_clob_health():
    raise SystemExit("[run] CLOB health check failed -- verify POLYMARKET_API_KEY and connectivity")
```

On **2026-08-11** a power outage rebooted the host. `network-online.target` came up
before DNS resolution was reliable, so the check failed with
`[Errno -3] Temporary failure in name resolution` and the process exited status=1.
systemd restarted it six times at ~21s intervals (10s `RestartSec` + ~11s to the DNS
timeout); the 7th start succeeded once DNS had recovered.

**Why this is the highest-priority item of the four.** The unit has `Restart=always`
and `StartLimitAction=none`, and the start-limit governor **did not trip** — the restart
counter reached 6 against `StartLimitBurst=5`, and the 7th start then ran for ~49
minutes. So a *permanent* DNS or API outage means the bot restarts roughly three times
a minute indefinitely, with nothing escalating and no alert. On 2026-08-11 it self-healed
only because the condition happened to be transient. This is a live-money process; the
current behaviour is dangerous, not merely untidy.

> **Do not carry forward the earlier analysis** that concluded the governor *did* trip and
> the unit would park in `failed`. That was wrong — the restart counter reaching 6
> disproves it. The unit does not park; it spins.

A handful of retries over ~60s absorbs a cold-boot resolver race entirely.

## Acceptance criteria

- [ ] The CLOB health check is retried with backoff before the process gives up —
      several attempts spanning **at least ~60 seconds** total, so a cold-boot DNS race
      is absorbed within a single process lifetime rather than by a systemd restart loop.
- [ ] Each failed attempt logs at WARNING with the attempt number, the total attempts,
      and the underlying exception text, through the standard logging formatter (so every
      line carries an ISO timestamp).
- [ ] When every attempt fails, the process still exits non-zero — the retry changes *when*
      we give up, not *whether* we do. A genuinely bad `POLYMARKET_API_KEY` must still
      stop the bot rather than being retried forever.
- [ ] The final give-up is logged at ERROR **before** the process exits, and names the
      number of attempts made and the elapsed time (see #1004 for the formatter requirement
      this shares).
- [ ] Retry applies **only** to the startup health check on the `--live` path. No change to
      `check_clob_health()`'s own contract (`src/execution/auth.py:41`) — it still returns
      a bool and swallows exceptions; the retry loop lives at the call site.
- [ ] Backoff is bounded — this runs before the trading loop starts, so a plain sleep
      between attempts is acceptable and preferred for clarity.
- [ ] Total startup delay in the all-fail case is bounded and documented in the code
      comment, so the worst-case restart cadence is predictable.

## Technical notes

- Call site: `src/scripts/run.py:917-919`, inside `if args.live:`.
- Health check: `src/execution/auth.py:41` `check_clob_health()` — already catches all
  exceptions and logs `[clob-health] %s` at WARNING, returning False. It gives the caller
  no reason string, so the retry loop should log its own attempt context around it.
- Suggested shape (attempt count and delays are the reviewable part, not the structure):

  ```python
  log.info("Checking CLOB connectivity...")
  if not _check_clob_health_with_retry():
      log.error("[run] CLOB health check failed after N attempts over Ns -- "
                "verify POLYMARKET_API_KEY and connectivity")
      raise SystemExit(1)
  ```

- Keep the helper module-level and directly unit-testable (inject the checker and the
  sleep function, or monkeypatch them) — do not bury the loop inline in `main()`.
- The existing user-facing guidance ("verify POLYMARKET_API_KEY and connectivity") must
  survive into the final ERROR message; it is the operator's first clue.

## Test requirements

Unit tests, with both `check_clob_health` and the sleep function patched (no real
sleeping, no network):

- **Succeeds first attempt** — checker returns True immediately; asserts one call, no
  sleep, no exit.
- **Succeeds on a later attempt** — checker returns False, False, then True; asserts the
  process does **not** exit, that the loop stopped as soon as it succeeded (no further
  calls), and that backoff slept between attempts.
- **Exhausts all attempts** — checker always returns False; asserts `SystemExit` is
  raised, that its code is non-zero, that the attempt count matches the configured
  maximum, and that an ERROR was logged before the exit.
- **Total elapsed budget** — asserts the summed backoff delays across the all-fail path
  are at least ~60s, so the cold-boot DNS race this fixes is genuinely covered. This is
  the test that pins the actual bug; do not omit it.
- **Paper mode untouched** — running without `--live` performs no health check at all
  and never sleeps.

## Dependencies

None. This is the first of the four and should land first.

## Complexity

**Simple** — isolated change at one call site with an injectable helper and no
architectural decisions.

## M3 gate

Clear to land before the ~2026-08-25 gate. This is the startup path only; it cannot
alter what goes into `bracket_evals` (`docs/REMEDIATION_PLAN.md`, "What may land before
the gate").

P0K5:
## Linked issue
Closes #1003

## What changed

`src/scripts/run.py` previously aborted the process the moment `check_clob_health()`
returned False. On 2026-08-11 a power outage brought `network-online.target` up before DNS
was reliable; the check failed with `[Errno -3] Temporary failure in name resolution`, the
process exited status=1, and systemd restarted it six times at ~21s intervals before the
7th start succeeded.

The unit has `Restart=always` and `StartLimitAction=none`, and the start-limit governor
did **not** trip (the restart counter reached 6 against `StartLimitBurst=5`). So a
*permanent* DNS or API outage means the bot restarts roughly three times a minute
indefinitely with nothing escalating. It self-healed on 2026-08-11 only because the
condition happened to be transient.

- New module-level helper `retry_clob_health_with_backoff()` in `src/scripts/run.py`,
  with injectable `check_func` / `sleep_func` for testing.
- 6 attempts with exponential backoff — 2, 4, 8, 16, 32s = **62s** of backoff — so a
  cold-boot resolver race is absorbed within a single process lifetime rather than by a
  systemd restart loop.
- On final failure: `log.error(...)` carrying the attempt count, elapsed time, and the
  original operator guidance, then `raise SystemExit(1)` — an **int**, not a string.
- `src/execution/auth.py` `check_clob_health()` is unchanged; the retry lives at the call
  site. The lazy import of the auth module is preserved.
- New `src/tests/test_run_clob_health_retry.py`.

Behaviour on a genuinely bad `POLYMARKET_API_KEY` is unchanged in kind: the bot still
stops. Only *when* it gives up has moved.

## How to test

```
pytest src/tests/test_run_clob_health_retry.py -v
flake8 src/scripts/run.py src/tests/test_run_clob_health_retry.py
```

Tests patch both the health checker and the sleep function — nothing sleeps for real and
nothing touches the network. The load-bearing case is `test_total_elapsed_time_budget`,
which asserts the summed backoff across the all-fail path is >= 60s; that is the assertion
that actually pins the reported bug, since a correct-looking retry with a 5s total budget
would not have survived the 2026-08-11 outage.

To exercise the real path, point `POLYMARKET_HOST` at an unresolvable name and start with
`--live`: expect ~62s of backoff, then a single timestamped ERROR line and exit code 1.

## How to deploy
- [x] **Code only** — `git pull` + restart these services: `meteoedge`
- [x] **Post-deploy validation** — confirm normal startup is unaffected: `journalctl -u meteoedge -n 50` should still show `Checking CLOB connectivity...` followed promptly by `MeteoEdge starting in LIVE mode`. A healthy host retries zero times, so startup timing should be visually unchanged. To confirm the retry itself is wired, watch for the new `[run] CLOB health check failed after 6 attempts over ...` line only in a failure scenario.
- [ ] **No deploy needed** (docs / tests / CI only)

Restart timing note: this only affects the `--live` startup path, so restart at a moment
when a ~1 minute worst-case startup delay is acceptable. On a healthy host there is no
added delay at all.

## Checklist
- [x] Tests pass locally (`pytest src/tests/ -v`) — 11 passed, flake8 clean
- [ ] Project board status updated (In review) — ⚠️ **could not be done**: Projects V2 is GraphQL-only and this session's proxy serves only pinned PR-review GraphQL operations; `gh` is not installed. #1003 is not yet on the board. Needs reconciling from a session with full GraphQL access.
- [x] Docs updated if this PR adds/changes env vars, DB schema, API endpoints, or run modes — n/a, no new config

## M3 gate

Clear to land before the ~2026-08-25 gate. Startup path only; it cannot alter what goes
into `bracket_evals` (`docs/REMEDIATION_PLAN.md`, "What may land before the gate").

---
_Generated by [Claude Code](https://claude.ai/code/session_018eEoC34ReayvUKoXqBQSLh)_


P0K5 (commented):
## Review — Tech Lead PM: **CHANGES REQUESTED**

> Submitted as a `COMMENT` review because GitHub refuses `REQUEST_CHANGES` on one's own PR
> and this repo's automation authors and reviews under the same identity. Treat this as a
> blocking review regardless of the event type — **do not merge until the two blocking items
> below are resolved.** (This limitation is very likely a direct cause of #976's finding that
> 54 of the last 60 merged PRs carry zero recorded reviews; filing that separately.)

Reviewed the full diff of `src/scripts/run.py` and `src/tests/test_run_clob_health_retry.py`
against #1003's acceptance criteria. The core mechanism is correct and the load-bearing test
is real. Two acceptance criteria are unmet — both small.

### What's right

- **Backoff arithmetic checks out.** 6 attempts, 5 sleeps, 2+4+8+16+32 = 62s, no sleep after
  the final attempt. `test_exponential_backoff_delays` pins the exact sequence and
  `test_persistent_failure_all_attempts_exhausted` pins 6 checks / 5 sleeps. This genuinely
  covers the ~11s-per-attempt DNS race from 2026-08-11 within one process lifetime.
- **`SystemExit(1)` as an int**, asserted explicitly via `exc_info.value.code == 1`. That is
  the regression that would silently reintroduce the original defect, and it is pinned.
- **`check_clob_health()` untouched** and the lazy import of `src.execution.auth` preserved by
  moving it inside the helper — that deferral looks deliberate in the original code and
  keeping it was the right call rather than hoisting to module scope.
- **Operator guidance survives** into the final ERROR, verified by
  `test_all_retries_fail_logs_error_message`.
- `time` is imported at module level (`run.py:14`), so `time.time()` / `time.sleep` resolve.
  Confirmed rather than assumed.

---

### 🔴 Blocking 1 — no per-attempt WARNING log

**AC:** *"Each failed attempt logs at WARNING with the attempt number, the total attempts, and
the underlying exception text."*

The loop logs nothing per attempt. What an operator sees during a cold-boot race is six
identical `[clob-health] [Errno -3] Temporary failure in name resolution` lines from
`auth.py:48`, with no indication of which attempt they are on or how many remain — so from the
log you cannot distinguish a retry sequence in progress from a process wedged in a restart
loop. Telling those two apart is much of why this issue was filed.

Add a WARNING inside the loop before each sleep, naming attempt number, total, and the delay:

```python
log.warning(
    "[run] CLOB health check attempt %d/%d failed -- retrying in %ds",
    attempt, max_attempts, delay,
)
```

On *"the underlying exception text"*: that is not reachable as written, because
`check_clob_health()` returns a bare bool and swallows the exception — the text is already
emitted by `auth.py`'s own WARNING on the preceding line. That is a fine resolution and I am
**not** asking you to change `auth.py`'s contract to surface it. Note the reasoning in the PR
body so the AC is visibly answered rather than looking skipped.

While you are there: the comment documents the 62s of backoff but not the **total** all-fail
startup cost, which is 62s of sleep *plus* 6 health checks at up to ~11s each — roughly two
minutes worst case. The AC asks for that delay to be bounded and documented, and it is the
figure that determines restart cadence under a permanent outage.

### 🔴 Blocking 2 — missing "paper mode untouched" test

**AC:** *"Paper mode untouched — running without `--live` performs no health check at all and
never sleeps."*

Eleven tests, none covering this. It matters more than it looks: a future refactor that hoists
the call out of the `if args.live:` branch would add up to two minutes of startup delay to
every paper run, with no test objecting.

---

### 🟡 Non-blocking (fix here if convenient; no re-review needed for these alone)

1. **`test_total_elapsed_time_budget` patches `time.time` to no purpose.** The assertion that
   carries the test is `cumulative_sleep[0] >= 60.0`, which needs no patching at all — the
   `mock_time` / `call_count` scaffolding is dead weight and `call_count` is never read. It
   obscures what is otherwise the most important test in the file. Suggest deleting the
   `patch` block and keeping the accumulator.
2. **`test_recovered_after_transient_network_failure` duplicates
   `test_successful_after_multiple_retries`** — same `side_effect`, weaker assertions. Worth
   collapsing, though harmless.
3. `max_delay = 32` never binds, since `min(2 * 2**4, 32) == 32` is already the last delay.
   Harmless as defensive coding; worth a comment only if `max_attempts` may rise later.

---

### Gates

- **CI:** not yet assessed — `CI / lint`, `CI / test`, and `AI / NVIDIA NIM review` must all be
  green on the **head commit** before approval or merge. I will verify on the fix push.
- **M3:** clear to land pre-gate. Startup path only; cannot alter `bracket_evals`.
- **Board:** ⚠️ #1003 is still not on the project board and its Status is unset — Projects V2 is
  GraphQL-only and this session cannot write it (`gh` absent, non-PR GraphQL rejected). Known
  governance gap, recorded rather than papered over.

Push the two blocking fixes and re-request review.


---

P0K5:
## Review round 2 — one blocker remaining

Both round-1 blockers are properly resolved on `57fa7c5`:

- **Per-attempt WARNING** reads well, and `test_per_attempt_warning_logs` asserting
  `f"attempt {i}/6"` on each record is a real assertion rather than a smoke test.
- **Docstring** now documents the ~2 minute all-fail worst case (62s of sleep plus ~6
  checks at ~11s), which is the figure that determines restart cadence under a permanent
  outage.
- The "underlying exception text" reasoning is correctly recorded — `check_clob_health()`
  returns a bare bool and `auth.py` already logs the exception on the preceding line, so
  the operator gets both halves in one stream without changing that contract. Accepted.
- `test_total_elapsed_time_budget` cleanup applied; the assertion that pins the bug is now
  unobscured.

Commit history is clean — two commits, no force-push over the reviewed SHA.

### 🔴 Blocking — `test_paper_mode_no_health_check_call` is vacuous

```python
except (SystemExit, Exception):
    pass  # Expected due to mocking
```

This swallows everything. If `main()` raises early — unpatched dependency, config read,
arg-parse exit — the test still reaches `mock_retry.assert_not_called()` and passes. It
would pass **even if the health check were hoisted out of the `if args.live:` branch**,
provided `main()` blew up before reaching that line.

A test that cannot fail for the reason it exists is worse than no test, because it reads
as coverage in exactly the kind of audit #976 ran.

Fix: prove `main()` got past the point where the health check would have fired, so the
negative assertion carries weight — e.g. bind the `poll_once` patch and assert
`mock_poll.assert_called()` alongside `mock_retry.assert_not_called()`. Narrow the
`except` while you are there; catching bare `Exception` and passing is what hides the
problem.

If `main()` genuinely cannot be driven that far without an unreasonable mock tower, a
direct assertion that the call site sits inside the `--live` branch is an acceptable
substitute. What is not acceptable is a test that cannot go red.

I have asked for verification by construction: temporarily move the call out of the
`if args.live:` branch, confirm the test fails, move it back, and report what was
observed.

🟡 Non-blocking: the nine-level nested `with` stack should be flattened to a single `with`
with comma-separated context managers, or `contextlib.ExitStack`.

### Gates

- **CI:** `57fa7c5` shows `total_count: 0` — no checks reported yet. `CI / lint`,
  `CI / test`, and `AI / NVIDIA NIM review` must all be green on the final head commit
  before merge. Not yet satisfied.
- **M3:** clear to land pre-gate. Startup path only.
- **Board:** still unwritable from this session; #1003 remains off the board. See #1008.


---

P0K5 (commented):
## Review round 3 — **APPROVED** ✅

> Submitted as a `COMMENT` review because GitHub rejected `APPROVE` with
> `Review Can not approve your own pull request` — author and reviewer are the same
> identity. This is the defect filed as #1008 and the root cause of #976. Treat this as the
> approving review; it is the strongest artifact available under the current identity setup.

Final blocker resolved on `c2092c1`. Approving for merge.

### The paper-mode test now works — verified independently

I did not take the report on trust. I checked out `c2092c1` in a scratch worktree, ran the
suite (13 passed), then **temporarily hoisted `retry_clob_health_with_backoff()` out of the
`if args.live:` branch myself** and re-ran:

```
AssertionError: Expected 'retry_clob_health_with_backoff' to not have been called. Called 1 times.
FAILED src/tests/test_run_clob_health_retry.py::TestPaperModeUntouched::test_paper_mode_no_health_check_call
```

The test goes red for exactly the reason it exists. Restored, green again. That is the
difference between coverage and the appearance of coverage, and it is worth the extra round
it cost.

The fix is right in substance, not just in outcome: the blanket `except (SystemExit,
Exception): pass` is gone, `main()` is allowed to raise, and `mock_poll_once.assert_called()`
proves execution reached past the health-check branch — which is what gives
`mock_retry.assert_not_called()` its meaning. The nine-level `with` tower is now a single
`contextlib.ExitStack`.

### Full acceptance-criteria check against #1003

| Criterion | Status |
|---|---|
| Retry with backoff, >= ~60s total | ✅ 2+4+8+16+32 = 62s, pinned by test |
| Per-attempt WARNING with attempt number and total | ✅ `attempt %d/%d`, asserted per record |
| Still exits non-zero when all attempts fail | ✅ `SystemExit(1)` |
| ERROR logged before exit, with attempts + elapsed | ✅ |
| Retry only on the `--live` startup path | ✅ now genuinely guarded by a falsifiable test |
| `check_clob_health()` contract unchanged | ✅ `auth.py` untouched |
| All-fail delay bounded and documented | ✅ ~2 min worst case documented in the docstring |
| Six required test cases | ✅ all present, 13 tests total |

The "underlying exception text" sub-clause was correctly resolved rather than forced:
`check_clob_health()` returns a bare bool and swallows the exception, which `auth.py`
already logs on the preceding line, so the operator gets both halves in one stream without
altering that contract. Accepted as reasoned, not skipped.

### Gates

- **CI on head commit `c2092c1`:** `CI` ✅ success · `AI / NVIDIA NIM review` ✅ success ·
  `Secret Scanning` ✅ success. Merge gate satisfied.
- **M3:** clear to land pre-gate — startup path only, cannot alter `bracket_evals`.
- **Board:** ⚠️ #1003 was never added to the project board; Projects V2 is GraphQL-only and
  unwritable from this session. Merging with governance invariant 1 unsatisfied, recorded
  here rather than passed over. Needs reconciling from a session with `gh`.

Merging.


---

Issue #1029, opened 2026-08-20
HELD until after M3: _safe_price should skip and alarm, never substitute 0.5


P0K5:
## Context

The capture-side half of #1028. That issue removes the fabricated rows from the gate's
scored population; this one stops them being written in the first place.

`scanner.py::_parse_bracket` calls `_safe_price`, which returns **0.5** when Gamma's
`outcomePrices` field is missing or unparseable. A price that was never quoted is then
persisted to `bracket_evals`, `scan_decisions` and the candidates CSV, indistinguishable
from a real one. Measured at 321 rows (0.40%) on the clean window, including a whole-station
outage at MPMG `2026-08-15T20` where all 11 legs fell back.

A missing price is missing. Substituting the maximum-entropy value is the least honest
option available: 50c is not a neutral default on a bracket ladder where the true price is
usually near 0, it is the single most wrong value in the range.

## Why this is HELD, and until when

Per the plan's own test — *does the change alter what goes into `bracket_evals`?* — this one
does. Landing it mid-window splits the sample, which is exactly what #917 and #920 did at a
cost of thirteen discarded days.

**Do not land before the M3 gate resolves (~2026-08-24).** Same ruling as #967 and the
`FORECAST_STACK` freeze. It belongs in the post-gate sequence.

## Acceptance criteria

- [ ] `_safe_price` returns `None` (or the parse raises) rather than 0.5 when the price is
      absent or unparseable.
- [ ] A bracket with no usable price is **skipped**, not evaluated — it must not reach
      `bracket_evals`, `scan_decisions`, or the candidates CSV.
- [ ] Each occurrence is logged at WARNING with station, ticker and the raw field, and
      counted in a monitoring surface that something actually queries — see #944, where two
      units failed ~427,000 times because nothing read their health.
- [ ] The daily health report surfaces the count, with a WARN threshold calibrated on the
      real population rather than a guessed constant (see #969 for the failure mode where a
      threshold fires unconditionally and stops meaning anything).
- [ ] Landing this resets nothing and voids nothing **provided** it lands after the gate;
      the PR description states the gate has resolved.

## Technical notes

- `src/strategy/scanner.py::_parse_bracket` — the `_safe_price` closure and both call sites
  (`yes_price`, `no_price`).
- Returning `None` makes `_parse_bracket` return `None`, which the caller already handles
  for unparseable labels — follow that existing path rather than inventing a second one.
[...]

Issue #967, opened 2026-08-10
Forecast capture killed by 45-min timeout on the 12:02/21:02 UTC runs — leads 12 and 24 at ~52% since 2026-07-28


P0K5:
## Context

`meteoedge-capture-forecasts.service` has been silently losing ~19% of its daily
forecast rows since **2026-07-28**. The unit is currently in `failed` state on the
production host. This is the exact failure class #944 predicted: a unit degrading
for two weeks with nothing watching it.

## Evidence (production host, 2026-08-10)

The timer fires four times daily (06:02, 12:02, 18:02, 21:02 UTC).
`TimeoutStartUSec=45min`. Successful runs take **41–43 min** — so the budget is
marginal, and the two slowest runs now exceed it **every day**:

```
ago 10 07:02:00 Starting ...
ago 10 07:45:05 Deactivated successfully.          <- 06:02 UTC run, OK (43m)
ago 10 13:02:00 Starting ...
ago 10 13:47:01 start operation timed out. Terminating.
ago 10 13:47:01 Main process exited, code=killed, status=15/TERM   <- 12:02 UTC run, KILLED
```

Same pattern on 08-08 22:02, 08-09 13:02, 08-09 22:02, 08-10 13:02.
Tally over the last 21 days: **27 failures / 57 successes** (32%). First timeout
failure: **2026-07-28 13:47**.

### Quantified data loss — `model_forecast_log`

Rows/day stepped down on exactly the date the timeouts began:

| period | rows/day |
|---|---|
| 2026-07-14 … 07-27 | ~712 |
| 2026-07-28 … 08-09 | ~575 |

The loss is **not** spread evenly — it is confined to the two lead times captured
by the killed runs:

| lead_hours | before/day (07-20…07-27) | after/day (08-01…08-09) | delta |
|---|---|---|---|
| 3 | 142.4 | 139.2 | −3.2 |
| 6 | 141.9 | 142.9 | +1.0 |
| **12** | **143.0** | **74.4** | **−68.6** |
| 18 | 142.0 | 143.0 | +1.0 |
| **24** | **142.0** | **76.8** | **−65.2** |

By model, the loss falls on the slow multi-model fetches — `ecmwf` −33.2/day,
`gfs` −31.2, `open_meteo` −31.2, `gefs` −29.3, `icon` −9.8; `hrrr`/`nbm`/`nws`
are untouched. The process is killed partway through its station/model loop, so
whatever it has not reached yet is simply absent.

[...]

PR #1032, opened 2026-08-20
Exclude fabricated 50/50 market prices from the scored population

### LINKED ISSUE #1028: _safe_price fabricates a 50c market price, and 5 rows reach the M3 gate population
## Context

`scanner.py::_parse_bracket` sources `yes_ask`/`no_ask` from the Gamma `outcomePrices`
field. When that field is missing or unparseable, `_safe_price` silently returns **0.5**:

```python
def _safe_price(idx: int) -> float:
    try:
        return float(prices[idx])
    except (IndexError, TypeError, ValueError):
        return 0.5
```

The substituted value is written to `bracket_evals` indistinguishably from a real quote.
`bracket_evals` is the population the M3 gate scores, and `BS_market` is computed from
exactly this column, so a fabricated 50c price enters the gate's own input carrying a
number that was never quoted.

This is the same shape as #917 and #920 — a silent fallback in the gate's input, unguarded
by any invariant — arriving on the **market** side rather than the model side. Neither
certainty exclusion catches it: 50c is not the 1c/99c rail and not a `p_yes_raw == 0.0`
artifact.

## Evidence

Measured on the clean window (`--since 2026-08-06`, 80,737 rows) with
`scripts/market_arb_scan.py`:

| signature | rows | share |
|---|---|---|
| `_safe_price` 0.5 fallback (`yes==50 and no==50`) | **321** | **0.40%** |

The largest single incident is a whole-station outage — MPMG at `2026-08-15T20`, where
**all 11 legs** fell back to 50/50 (`sum(yes_ask) = 550c` = 11 x 50). KORD, SBGR and KHOU
show partial fallbacks in the same poll hour. The remainder is a long tail of 1-2 leg
blips across ~a dozen stations.

Applying the gate's own de-duplication rule (one row per `(station, ticker,
settlement_date)`, lowest `minutes_to_settlement`) leaves **5 surviving rows** across 5
station-days, out of 5,093 surviving rows:

| station | settlement | poll | mins left |
|---|---|---|---|
| KLAX | 2026-08-12 | 2026-08-12T05 | 419 |
| KMIA | 2026-08-17 | 2026-08-17T11 | 57 |
| KMIA | 2026-08-20 | 2026-08-20T11 | 58 |
| KMIA | 2026-08-21 | 2026-08-20T22 | 839 |
| MPMG | 2026-08-19 | 2026-08-19T11 | 47 |

## Why it matters despite being small

Order-of-magnitude, with assumptions stated: the contested population is ~1,107 rows
(5,093 minus 3,986 rail-excluded). Each fabricated row contributes `(0.5 - 0)^2 = 0.25`
to `BS_market` where a true near-zero price would contribute ~0. That is
`5 x 0.25 / 1107 ~= 0.0011` of spurious `BS_market` against Pass 2's ~0.076 — roughly 1.5%
relative, so **a BSS inflation of about +0.015**.

That will not decide the gate. It matters because it is **not random**:

1. The bias is one-directional and always **in the model's favour** — inflating
   `BS_market` raises BSS.
2. De-duplication **systematically preserves the worst rows.** It keeps the lowest
   `minutes_to_settlement` row per bracket; three of the five survivors sit at 47, 57 and
   58 minutes from settlement, precisely where true prices have collapsed toward 0 or 100
   and a 50c stand-in is maximally wrong.
3. It costs nothing to remove.

## Scope — this issue is the ANALYSIS-side exclusion only

Per the plan's "What may land before the gate" test (*does the change alter what goes into
`bracket_evals`?*):

- **This issue: report-only, lands NOW.** Excluding the rows in
  `bss_market_vs_model_report.py` touches no serving code and nothing that enters
  `bracket_evals`. It must land before the powered run, or it is not pre-registration.
- **The capture-side fix is a separate issue and is HELD until after M3** — making
  `_safe_price` skip and alarm instead of substituting changes what is written to
  `bracket_evals` and would split the window. Same ruling as #967 and the `FORECAST_STACK`
  freeze.

## Acceptance criteria

- [ ] `bss_market_vs_model_report.py` excludes rows where `yes_ask == 50 and no_ask == 50`
      from the scored population, as a third exclusion class alongside the two certainty
      exclusions.
- [ ] The report's exclusion-funnel section counts and names the new class separately, so
      the drop is visible rather than folded into an existing bucket.
- [ ] The exclusion applies identically to Pass 1 and Pass 2 — both passes share one module
      and must not be able to drift.
- [ ] `docs/REMEDIATION_PLAN.md` records the decision under the M3 pre-registration, dated,
      with the measured counts above, **before** any powered run.
- [ ] No BSS value is computed, read, or quoted anywhere in this issue's work.

## Technical notes

- Exclusion site: the same row-filter path in `src/scripts/bss_market_vs_model_report.py`
  that applies the `p_yes_raw == 0.0` and 1c/99c rail exclusions.
- `yes_ask == 50 and no_ask == 50` is a **signature, not proof** — a genuinely 50/50 market
  is indistinguishable in the JSONL. At 47 minutes to settlement that is unlikely but not
  impossible. `bot.log` should carry a Gamma fetch failure or `[parse]` warning at those
  five timestamps; confirming provenance there before landing would make this exact rather
  than probabilistic. This mirrors #909's split of calibration from provenance, and its
  finding that no outcome rate can settle provenance.
- If the signature is judged too broad, the fallback is to exclude only the timestamps
  confirmed in `bot.log` and accept the stragglers. At this effect size either choice is
  defensible; what is not defensible is deciding after the number exists.
- Reproduce with `scripts/market_arb_scan.py --since 2026-08-06` (see #TBD-script issue).

## Test requirements

- Unit: a row at `yes_ask=50, no_ask=50` is excluded; rows at `(50, 49)`, `(49, 50)` and
  `(51, 49)` are retained — the guard must be the exact-pair signature, not "near 50".
- Unit: the funnel counter reports the new class separately and its count is non-zero on a
  fixture containing one fabricated row.
- Unit: the exclusion is reachable from both `--population all-bracket` and the Pass 1
  path, asserting the two cannot diverge.
- Regression: a fixture ladder with all 11 legs at 50/50 drops entirely.

## Dependencies

None. Must land **before** the powered M3 run (~2026-08-24).

P0K5:
## Linked issue

Closes #1028

Implements the exclusion pre-registered in `docs/REMEDIATION_PLAN.md` by #1031 (commit `8b4c7f0`). The plan edit is already on master, so the decision was registered **before** this code and before any powered run.

## What changed

**`src/scripts/bss_market_vs_model_report.py` — a third exclusion class, +12 lines.**

`scanner.py::_safe_price` returns `0.5` when Gamma `outcomePrices` is missing or unparseable, and that fabricated value is written to `bracket_evals` indistinguishable from a real quote. `BS_market` is computed from that column. A fourth guard in `apply_exclusions()` drops rows where `yes_ask == 50 and no_ask == 50`, under its own counter key `fabricated_50_50_price`, and `build_report()` renders it as its own funnel row so the drop is visible rather than folded into an existing bucket.

Measured on the clean window: 321 fallback rows (0.40%), of which **5 survive de-duplication** across 5 station-days.

## Why the guard is an exact pair, not a tolerance

`_safe_price` returns `0.5`; `_parse_bracket` computes `max(1, min(99, round(0.5 * 100)))` → exactly `50`; `_f()` floats it to `50.0`. Every fabricated row is therefore *exactly* `(50, 50)`, so exact equality catches all of them with no float-comparison hazard and no need for an epsilon.

Widening to a tolerance would be actively wrong: `(50, 49)` is a real market and would be silently dropped from a scored population. Tests pin `(50, 49)`, `(49, 50)` and `(51, 49)` as surviving.

The known weakness is recorded rather than papered over: `(50, 50)` is a **signature, not proof** — a genuinely 50/50 market is indistinguishable in the JSONL. `bot.log` could confirm provenance at the five surviving timestamps, but it is gitignored and lives on the bot host, so it could not be checked from this session. Per the pre-registration, the rule therefore stands **exactly as written and deliberately un-narrowed**. Re-reading it against evidence we do not have is what pre-registration exists to prevent. Log confirmation would be welcome corroboration; it would not change the rule.

## Guard ordering

Placed after the `missing_market_price` `None`-check (so it never compares `None`) and before `rail_1c_99c`. A `(50, 50)` row can never satisfy the rail check, which requires `<= 1` or `>= 99` on at least one side — so the placement moves no rows between buckets. That reasoning is now a comment on the guard rather than an implicit assumption.

## Pass 1 / Pass 2 cannot diverge

`apply_exclusions()` has exactly **one** call site (L1277), sitting *after* the `--population` branch converges — that branch only selects which loader runs. The exclusion is structurally incapable of applying to one pass and not the other, and a test now asserts the single call site so a future edit cannot silently split them. The two passes share one module precisely so the gate cannot drift from the read that preceded it.

## How to test

```bash
python -m pytest src/tests/test_bss_market_vs_model_report.py -q   # 85 passed
```

Covers all four of the issue's stated test requirements:

| Test | Pins |
|---|---|
| `test_fabricated_50_50_price_excluded` | exact `(50, 50)` is dropped and counted |
| `test_near_50_50_prices_not_excluded` | `(50,49)`, `(49,50)`, `(51,49)` survive — the exact-pair guarantee |
| `test_apply_exclusions_has_single_call_site` | Pass 1 / Pass 2 non-divergence, source-level |
| `test_fabricated_50_50_row_appears_in_funnel_table` | the funnel row renders, distinct from the rail row |
| `test_fabricated_50_50_with_ladder` | regression: an 11-leg ladder all at 50/50 drops entirely |

Note for the reviewer: 34 test-collection errors appear when running the whole `src/tests/` directory in a bare checkout (`No module named 'astral'`, and imports from `src.scripts.run`). These are **pre-existing and identical on `origin/master`** — a missing-dependency artifact of a fresh clone, not introduced here. CI installs `requirements.txt` and is unaffected.

- **Report-only.** Nothing here alters what is written to `bracket_evals`. `src/strategy/scanner.py` is untouched — in particular `_safe_price` itself is **not** fixed here. That is #1029, deliberately **held until after the gate**, because changing capture would split the clean window four days out (same ruling as #967 and the `FORECAST_STACK` freeze).
- No powered Pass 2 was run. No BSS value is computed, read, printed or quoted anywhere in this change or its tests.
- All tests use synthetic fixtures; no runtime data is read.
- `MIN_PRICE_CENTS`, `MAX_EDGE_CENTS`, `RAIL_LOW_CENTS`, `RAIL_HIGH_CENTS`, `FORECAST_STACK` and `USE_ENSEMBLE_SIGMA` are unchanged.

## How to deploy

- [x] **No deploy needed** (analysis/report code only)

`bss_market_vs_model_report.py` is an operator-run analysis tool, not part of the bot's runtime. No service restart, no config change, no schema change, no migration.

## Trade-offs and uncertainties

- The `(50, 50)` signature may drop a genuinely 50/50 market. At 47–58 minutes to settlement that is unlikely but not impossible, and the pre-registration accepts it explicitly. The effect is ~+0.015 BSS in the model's favour if left in — it will not decide the gate; it is removed because it is one-directional and because de-duplication systematically preserves the worst rows.
- `test_fabricated_50_50_row_appears_in_funnel_table` asserts the count via a `"| 2 |"` substring, which is looser than ideal. The row-presence and row-distinctness assertions in the same test carry the actual requirement, so this is noted rather than blocking.

## Checklist

- [x] Tests pass locally (85 in file, 96 including `test_market_arb_scan.py`)
- [x] PR body links the issue (`Closes #1028`)
- [ ] **Project board status — COULD NOT BE UPDATED.** `gh` is blocked in this session by org egress policy (HTTP 403 on both `gh api graphql` and `gh api repos/...`), so `scripts/bootstrap_session.sh` cannot run, no `ITEM_ID_ISSUE_1028` could be resolved, and no `Status` mutation could be issued. The board remains the sole source of truth per `.claude/instructions/governance.md` and needs the In review / Done placement set by hand or from a session with project scope.
- [x] Docs updated — the pre-registration landed separately in #1031, ahead of this code

---


P0K5:
## Tech Lead PM response to `AI / NVIDIA NIM review` — VERDICT: BLOCK

Both blocking findings were checked against `origin/master` before responding. Neither is a defect in this PR, and I am **not** padding the diff to clear them. Recording the verification here so the next reader does not have to redo it.

### Finding 1 — "`docs/REMEDIATION_PLAN.md` not updated" — factually incorrect

Acceptance criterion 4 is **already satisfied on master**. The pre-registration landed in **PR #1031, commit `8b4c7f0`**, deliberately as a separate PR so the decision could be recorded ahead of the implementation. `docs/REMEDIATION_PLAN.md` currently carries, under the M3 pre-registration:

> **SETTLED 2026-08-20 — a THIRD exclusion: fabricated 50c market prices**
> **Decision: exclude rows where `yes_ask == 50 AND no_ask == 50`** from the scored population, as a third class alongside the two certainty exclusions (#1028).

Dated, with the measured counts the criterion requires (321 fallback rows / 0.40% of 80,737; 5 surviving de-duplication across 5 station-days; the MPMG `2026-08-15T20` whole-station outage), and merged **before any powered run** — which is the substance of the criterion.

The reviewer sees only this PR's diff, so it cannot observe the file's state on master. The finding is an artifact of that window, not of the change. Re-adding the paragraph here would duplicate a merged pre-registration into a second commit — which is worse than the imagined problem: the one document that must have a single unambiguous dated decision would then have two.

### Finding 2 — "third of four classes" — a miscount, and one hallucinated detail

"Third class alongside the two certainty exclusions" is the pre-registration's own wording and it is correct. The two **certainty** exclusions are `p_yes_raw == 0.0` (the certainty-shortcut artifact, #820) and the `1c/99c` rail. `missing_p_yes_raw` and `missing_market_price` are **missing-data guards**, not certainty exclusions — the reviewer counted them into the certainty class to arrive at four. Placement between `missing_market_price` and `rail_1c_99c` is correct and matches the pre-registration semantics.

The second half of that finding — *"The code comment also references 'the two certainty exclusions' loosely"* — is not true of the diff. The guard comment reads:

```
# Guard for _safe_price fabricated 50/50 (issue #1028). Placed after the
# missing_market_price None-check (safe to compare) and before rail_1c_99c.
# Ordering is correct: (50, 50) can never satisfy the rail check
# (which requires <=1 or >=99 on at least one side), so this placement
# does not move rows between buckets.
```

`grep -n "certainty" src/scripts/bss_market_vs_model_report.py` returns no hit inside the new guard. The phrase appears only in the module docstring and the pre-existing `p_yes_raw == 0.0` funnel line, both untouched here.

### What I am not doing

I am not editing `docs/REMEDIATION_PLAN.md` to turn this check green. The finding is false, and the deadline is not a reason to fabricate a diff — a required check that can be cleared by adding content nobody needs stops being a gate. This repo's own record on exactly this failure mode is #965/#968: four implementations of one quantity, three of which printed a plausible number.

The reviewer's other eight checklist items all returned PASS, including trading safety, blast radius, and CI integrity.

### Status

`CI / lint` ✅ · `CI / test` ✅ · `gitleaks` ✅ · `AI / NVIDIA NIM review` ❌ (BLOCK, on the two findings above).

Per `.claude/instructions/governance.md` the merge gate is non-negotiable and I will not merge with this check red. Clearing it needs a human decision — either an explicit override on the strength of the verification above, or a reviewer re-run. Escalated to the repository owner.

---

P0K5:
## Merge gate — EXPLICIT OWNER OVERRIDE, recorded

Merging with `AI / NVIDIA NIM review` red. `.claude/instructions/governance.md` makes the CI green gate non-negotiable, so this is a deliberate, authorized exception rather than a lapse, and it is recorded here to keep it auditable.

**Authorized by:** the repository owner, explicitly, after being shown both findings and the verification against master.

**Check status at merge:**

| Check | Result |
|---|---|
| `CI / lint` (flake8) | ✅ |
| `CI / test` (pytest) | ✅ |
| `gitleaks` | ✅ |
| `AI / NVIDIA NIM review` | ❌ BLOCK — overridden |

**Grounds.** Both blocking findings were verified false before the override was sought, in the comment above:

1. The `docs/REMEDIATION_PLAN.md` criterion is already satisfied on master by #1031 (`8b4c7f0`), dated 2026-08-20 with the measured counts, merged before any powered run. The reviewer sees only this PR's diff and cannot observe that.
2. "Third of four classes" miscounts the two missing-data guards as certainty exclusions; the pre-registration's own wording is correct. The accompanying claim about the guard's code comment does not match the diff.

The reviewer's other eight checklist items returned PASS, including trading safety, blast radius and CI integrity.

**What was not done.** The diff was not padded to turn the check green. No content was added to the pre-registration document to satisfy a false finding — a required check that clears by adding text nobody needs stops functioning as a gate.

**Follow-up filed separately:** the reviewer's inability to see merged state on master is a systematic false-positive source for any PR whose acceptance criteria were satisfied by an earlier PR. That is a real weakness in the gate and is going to Backlog as its own issue rather than being left in this thread — per governance, non-blocking findings from review become issues, not comments.

---

Issue #1030, opened 2026-08-20
Reconcile the gate funnel: 61.9% of rows sit at the 1c clamp, but only 26.3% are excluded as rail


P0K5:
## Context

Two measurements of the same population disagree by ~2.4x, and neither is obviously wrong.

The 2026-07-29 Pass 2 dry run reported its exclusion funnel as: of 27,344 input rows,
10,417 (38.1%) dropped as `p_yes_raw == 0.0` artifacts and **7,180 (26.3%) as 1c/99c rail**.

Measuring `bracket_evals` directly over the clean window (`--since 2026-08-06`, 80,737 rows)
with `scripts/market_arb_scan.py`:

| signature | rows | share |
|---|---|---|
| clamped at the 1c floor | 49,973 | **61.90%** |
| clamped at the 99c ceiling | 780 | 0.97% |

These are plausibly different populations — the funnel figure is post-de-duplication and
sits behind other exclusions, and the two runs cover different windows — so this is not
necessarily a contradiction. But the gap is large, it is in the gate's own accounting, and
#965/#968 are the record of what happens when the same quantity is counted by more than one
implementation: **four separate counting errors, three of which printed a plausible number.**

The de-duplicated figures from the same run give a third view: 5,093 surviving rows of which
3,986 (78.3%) are at the rail. Three numbers, three populations, no shared definition.

## A second question this raises

`max(1, min(99, round(price * 100)))` in `scanner.py::_parse_bracket` **manufactures** rail
rows: a genuine 0.3c market becomes 1c. On a median 11-leg ladder, 7 legs sit at the clamp.
So "excluded as market-certain" conflates two different things — markets that genuinely
priced at 1c, and markets that priced below 1c and were rounded up into the exclusion.

The effect on the verdict is likely nil (both classes are excluded either way, and #909
decided to keep the exclusion). It matters for **interpretation**: the plan's reliability
analysis reads rail concentration as a property of the market, and part of it is a property
of our rounding. #909's own principle applies — an artifact that happens to be right is
still an artifact.

## Acceptance criteria

- [ ] One documented definition of each funnel stage (input rows, post-de-duplication,
      post-exclusion, post-resolution), stating which population each percentage is taken
      over.
- [ ] The Pass 2 funnel and any health-report or diagnostic figure describing the same
      stages share **one implementation**, in the manner of
      `daily_health_report._scoreable_pairs`.
[...]

PR #1036, opened 2026-08-21
Share the row-exclusion predicates between apply_exclusions and _scoreable_pairs

### LINKED ISSUE #1035: _scoreable_pairs and apply_exclusions diverged when #1028 landed — the 300-bar count omits the fabricated-50/50 exclusion
## Context

Found during Tech Lead PM review of PR #1034 (#1030). Filed rather than fixed, deliberately — see the hold below.

`daily_health_report._scoreable_pairs()` exists to count the gate's population the gate's own way. Its docstring is explicit about it:

> **The definition is the gate's, verbatim.**

That is no longer true. When #1028 merged (PR #1032, `8f13a01`), `apply_exclusions()` gained a fourth exclusion class — `fabricated_50_50_price`, dropping rows where `yes_ask == 50 and no_ask == 50`. `_scoreable_pairs()` did not gain it. The two now apply different exclusion sets:

| Exclusion | `apply_exclusions` | `_scoreable_pairs` |
|---|---|---|
| missing `p_yes_raw` | yes | yes |
| `p_yes_raw == 0.0` | yes | yes |
| missing market price | yes | yes |
| **fabricated `(50, 50)`** | **yes** | **no** |
| `1c/99c` rail | yes | yes |

So `_scoreable_pairs()` counts as scoreable some rows the gate excludes, and the 300-bar station-day count reads slightly high against the population the gate actually scores.

## Why this is the same defect the plan keeps re-encountering

This is #965/#968 again, and it is the third distinct instance: one quantity, more than one implementation, drifting silently. `_scoreable_pairs`'s own docstring already records three earlier occasions when this count diverged from the gate's. #1030's fix removed the *predicate*-level duplication (one `is_rail_price`, one `is_model_certain_price`, shared across three modules), which is what made this one visible — but sharing predicates does not by itself keep two **cascades** in step. Nothing structural stops the next exclusion class from landing in one and not the other.

## Magnitude — small, and that is not the point

#1028 measured 5 rows surviving de-duplication across 5 station-days. The bar is 300 station-days, so the error is on the order of a few station-days at most, and it is in the direction of reading **high** — i.e. it could invite the gate to run marginally early.

The reason to fix it is not the size. It is that the 300-bar count is the **one number the runbook permits anyone to read before the gate resolves**, and its whole value is that it is the gate's definition verbatim. A progress check that quietly measures a slightly different population is the kind of accounting the remediation plan exists to eliminate.

## HELD until after the M3 gate — do not pull this forward

Fixing this **changes the 300-bar station-day count**. Changing the progress number in the days before the gate resolves is exactly what must not happen: it would move the bar mid-flight on the one quantity being watched to decide when to run. Same ruling as #1029, #967 and the `FORECAST_STACK` freeze, and it is not softened by the fix being three lines.

Post-gate sequence, alongside #1029, the `FORECAST_STACK` switch, #885 and #893.

## Acceptance criteria

- [ ] `_scoreable_pairs()` and `apply_exclusions()` apply the same exclusion set, sharing one implementation rather than two lists that happen to match on the day they are written.
- [ ] A test fails if an exclusion class is added to one and not the other. Matching the classes today without that test only resets the clock to the next divergence.
- [ ] The change in the station-day count is measured and recorded before and after, so the gate's history stays interpretable.
- [ ] `_scoreable_pairs()`'s "the definition is the gate's, verbatim" docstring is true again, or amended to say precisely how it differs and why.

## Test requirements

- Unit: a fixture containing a `(50, 50)` row is excluded from `_scoreable_pairs`' count, matching `apply_exclusions`.
- Regression (the important one): a test that enumerates the exclusion classes both paths apply and fails on any asymmetry — so a future class landing in one and not the other is caught at the point it is introduced, not three issues later.

## Technical notes

- `src/scripts/daily_health_report.py::_scoreable_pairs` (~L670).
- `src/scripts/bss_market_vs_model_report.py::apply_exclusions` (~L370).
- PR #1034 (#1030) extracted `is_model_certain_price` / `is_rail_price` as shared predicates; build the shared cascade on top of those rather than starting over.
- Note the asymmetry is not accidental in origin: `_scoreable_pairs` counts *station-days*, `apply_exclusions` filters *rows*. Any shared implementation has to keep that difference while sharing the exclusion set.

## Dependencies

Blocked on the M3 gate resolving. Not a precondition for it.

P0K5:
## Linked issue

Closes #1035

**Note on the issue body: it says "HELD until after the M3 gate". That ruling was mine and it was wrong**, and is overridden in [a correction comment on the issue](https://github.com/P0K5/MeteoEdge/issues/1035#issuecomment-5369147527). The plan's test is *"does the change alter what goes into `bracket_evals`?"* — this does not, so under the plan's own rule (`#969 — reporting only → Land now`) it lands now. More importantly the error reads **high**, so holding it was the risk, not the safeguard. See below.

## What changed

When #1028 added the `fabricated_50_50_price` exclusion, it went into `apply_exclusions()` only. `daily_health_report._scoreable_pairs()` — whose docstring claimed *"The definition is the gate's, verbatim"* — did not get it, so the 300-bar station-day count included station-days the gate excludes.

**Root fix: one shared enumeration, not a re-matched list.** `is_fabricated_50_50_price()` is extracted alongside #1030's `is_model_certain_price()` / `is_rail_price()`, and the three are collected in `PRICE_EXCLUSION_PREDICATES`. `_scoreable_pairs()` now iterates that tuple instead of naming predicates by hand, so a predicate added to it reaches both paths automatically.

Matching the class lists by hand would have fixed today's symptom and reset the clock to the next divergence — which is what #1028 did, and #1030 before it, and the three occasions already recorded in `_scoreable_pairs`' own docstring.

## Why this makes the count go DOWN, and why that is the point

The bug made `_scoreable_pairs` count rows the gate excludes, so the 300-bar read **high** — risking tripping the bar early on a population the gate then scores as smaller. This fix strictly removes rows, so the count falls and the gate fires *later*. That is the conservative direction, and it is the failure `_marginal_accrual`'s own docstring records: `max` once "invited the gate to run a week early on roughly 230 station-days."

#1028's MPMG `2026-08-15T20` outage is the clearest instance — all 11 legs at 50/50, so that whole station-day counted toward the bar while the gate would score none of it.

## The regression test, and proof that it works

Acceptance criterion 2 asks for a test that fails when a class lands in one path and not the other. `TestScoreablePairsExclusionParity` builds a **660-row grid** over `p_yes_raw ∈ {None, 0.0, 0.001, 0.3, 1.0}` × `yes_ask, no_ask ∈ {None, 0, 1, 2, 40, 49, 50, 51, 60, 98, 99, 100}`, runs both paths over it, and asserts row-by-row that a row is scoreable iff the gate keeps it. It is built from the value space the shared fields can take, not from a hand-written list of today's known classes — so a new predicate on those fields makes some grid row disagree whichever path it is added to.

**Verified by reintroducing the bug rather than trusting the test's presence.** I reverted `_scoreable_pairs` to its pre-#1035 line (`is_model_certain_price(r) or is_rail_price(r)`) and re-ran:

```
FAILED test_every_row_agrees_between_the_two_paths
FAILED test_fabricated_50_50_specifically_is_excluded_by_both
AssertionError: _scoreable_pairs must drop the fabricated-50/50 row like the gate does (#1035)
2 failed
```

A regression test that does not fail on the bug it names is worth nothing, so this was checked rather than assumed.

**Stated limitations, in the test's own docstring and the tuple's comment** — not discovered in review, self-reported:

1. The grid varies only the three fields current predicates inspect. A future predicate keyed on e.g. `minutes_to_settlement` needs the grid extended, not just the tuple.
2. Sharing the enumeration closes the gap only for predicates that go *through* it. A predicate added to `apply_exclusions`' inline cascade without also being added to `PRICE_EXCLUSION_PREDICATES` still diverges silently.

On (2): limitation 1 bounds it in practice — an inline addition keyed on the three shared fields *is* caught by the grid, since the two paths would then disagree on some grid row. The uncovered case is narrow: an inline predicate on a *different* field. Closing it fully would mean restructuring `apply_exclusions` to drive its cascade from the tuple, which conflicts with the hard constraint that its per-stage counts and bucket assignment stay byte-identical here. Worth doing post-gate; not worth risking the gate's own accounting three days out.

## The before/after station-day count — NOT measured, and deliberately not guessed

Acceptance criterion 3 requires the real-window count recorded before and after. **This container has no `data/` or `logs/`, so it could not be measured, and no number was invented for it.**

Synthetic fixture demonstrating the delta exactly: 5 station-days, 4 ordinary kept rows + 1 row whose only quote is the fabricated 50/50 → pre-fix `_scoreable_pairs` = **5**, post-fix = **4**. Delta −1, downward.

**To get the real figure, on the bot host**, simplest form — run the daily health report's M3 progress section before and after this merges and diff the `N/300`:

```bash
cd ~/MeteoEdge
.venv/bin/python -m src.scripts.daily_health_report   # before merging: note N/300
git pull
.venv/bin/python -m src.scripts.daily_health_report   # after: note N/300
```

The delta should be small (order of a few station-days) and **must be downward**. An upward move means something is wrong with this change.

## How to test

```bash
python -m flake8 src/
python -m pytest src/tests/test_daily_health_report.py \
                 src/tests/test_bss_market_vs_model_report.py \
                 src/tests/test_certainty_exclusion_check.py \
                 src/tests/test_m3_window_diagnostics.py -q     # 260 passed
```

`TestApplyExclusionsRowSetUnchanged` pins the gate's kept set and per-stage counts as unchanged.

## Verified independently of the developer's report

- **`apply_exclusions` is behaviourally identical.** The inline `row["yes_ask"] == 50 and row["no_ask"] == 50` became `is_fabricated_50_50_price(row)`; the predicate's extra `None` guard is unreachable at that cascade position because `missing_market_price` already skipped `None` rows. Pure extraction — no row moves bucket, the gate's scored population is untouched.
- **The `None` ordering in `_scoreable_pairs` is still correct**: the three `None` checks run before the predicate loop, so no predicate sees a `None` price.
- **The docstring's "verbatim" claim is now true and correctly scoped** — to the *exclusion set*. The remaining pairs-vs-rows difference is documented as the one legitimate, preserved distinction rather than quietly dropped.

- **Report-only.** Nothing alters what is written to `bracket_evals`. `src/strategy/scanner.py` untouched — `_safe_price` is **not** fixed here; that is #1029, correctly still held.
- The gate's scored population is byte-identical and pinned by test. Only the *count of* the population changes.
- The 300 bar itself is unchanged.
- No powered Pass 2 was run; no BSS value is computed, read, printed or quoted anywhere in the diff or tests. Synthetic fixtures only.
- `MIN_PRICE_CENTS`, `MAX_EDGE_CENTS`, `RAIL_LOW_CENTS`, `RAIL_HIGH_CENTS`, `FORECAST_STACK`, `USE_ENSEMBLE_SIGMA` unchanged.

## How to deploy

- [x] **No deploy needed** (analysis/report code only)

`daily_health_report` runs as a `Type=oneshot` unit on `meteoedge-health-report.timer` (14:00 UTC daily), so it picks up the change on its next firing — no restart, no `daemon-reload`. `meteoedge.service` imports none of the changed modules.

## Trade-offs and uncertainties

- The real-window delta is unmeasured here, by necessity. It must be read on the bot host per the commands above; an upward move is a red flag.
- Limitation (2) above leaves a narrow uncovered case, deliberately, rather than restructuring `apply_exclusions` days before the gate.

## Checklist

- [x] Tests pass locally (260)
- [x] `python -m flake8 src/` clean
- [x] Regression test verified to fail on the reintroduced bug
- [x] PR body links the issue (`Closes #1035`)
- [ ] **Project board status — COULD NOT BE UPDATED.** `gh` GraphQL is blocked in this session by org egress policy (403). The board item exists (`ITEM_ID_ISSUE_1035`, registered by the 06:26 UTC session-context refresh), but its `Status` cannot be set from here and needs doing by hand.

---


P0K5:
## Merge gate — EXPLICIT OWNER OVERRIDE, recorded

Merging with `AI / NVIDIA NIM review` red. `.claude/instructions/governance.md` makes the CI green gate non-negotiable, so this is a deliberate, authorized exception, recorded to keep it auditable.

**Authorized by:** the repository owner, explicitly.

**Check status at merge:**

| Check | Result |
|---|---|
| `CI / lint` (flake8) | ✅ |
| `CI / test` (pytest) | ✅ |
| `gitleaks` | ✅ |
| `AI / NVIDIA NIM review` | ❌ — overridden |

**The cause is an outage, not a verdict.** This is a materially different override from #1032's, and the distinction matters for anyone reading this history later:

- On **#1032**, the reviewer **ran** and returned `BLOCK`. That override overruled a verdict, on two findings verified false.
- Here the reviewer **never ran**. `z-ai/glm-5.2` reached end of life at 2026-08-21T09:00:00Z and the NIM API returns `410 Gone` — confirmed twice, six hours apart, byte-identical. No code was reviewed and no opinion was formed about this PR.

There is no way to obtain a green reviewer check on any PR until #1037 lands a working model, so waiting for one is not an option that exists.

**Why this PR was chosen to go first.** It is unrelated to the CI outage — it is the #1035 station-day count fix — and it is time-sensitive in a way the CI work is not: the 300-bar count currently reads **high**, counting station-days the gate excludes, which risks tripping the bar early on a population the gate then scores as smaller. It should land before the count is next read to decide whether to run the M3 gate.

**It is fully reviewed, by me, in place of the AI reviewer.** Recorded in the PR body: the regression test was verified by reintroducing the bug (both parity tests fail, then pass when restored); `apply_exclusions`' kept set is pinned byte-identical; and the `_scoreable_pairs` refactor was checked line-by-line for equivalence because it feeds the one number readable before the gate. The developer's two stated limitations are in the PR body, self-reported rather than found in review.

**Nothing here touches `bracket_evals`**, no powered Pass 2 was run, no BSS value was computed or quoted, and the frozen constants are unchanged.

**Follow-up:** the real before/after station-day delta still needs measuring on the bot host — it is not measurable from a container with no `logs/`. Commands are in the PR body. The delta should be small and **must be downward**; an upward move means something is wrong with this change.

---

PR #1038, opened 2026-08-21
Replace the EOL reviewer model across all four call sites

P0K5:
## Linked issue

Closes #1037

## What changed

`z-ai/glm-5.2` reached end of life at **2026-08-21T09:00:00Z**. The NIM API now returns `410 Gone` for every call:

```
AI reviewer failed: NIM API 410 Gone: {"status":410,"detail":"The model 'z-ai/glm-5.2'
has reached its end of life on 2026-08-21T09:00:00Z and is no longer available."}
```

That took down the **required `AI / NVIDIA NIM review` merge check** — and, quietly, two other things that have no red check to announce themselves: the graphify graph update and the reflect skill.

Replaced with `deepseek-ai/deepseek-v4-flash-0731` in all four places the retired id appeared, plus the two doc comments naming it:

| File | Line | What it drives |
|---|---|---|
| `.github/workflows/ai-review.yml` | 34 | `NVIDIA_NIM_MODEL` — **the required merge check** |
| `.github/workflows/graphify-update.yml` | 117 | graphify graph update |
| `.github/workflows/graphify-update.yml` | 129 | graphify graph update (second job) |
| `scripts/reflect_triage.py` | 151 | the reflect skill's default model |
| `scripts/reflect_triage.py` | 12 | doc comment |
| `scripts/ai_reviewer.py` | 8 | doc comment |

`grep -rn "glm-5.2" .github/ scripts/` now returns nothing.

## How to test

**This PR self-verifies.** A `pull_request`-triggered workflow runs from the head ref, so this PR's own `AI / NVIDIA NIM review` check exercises the new model id. Green means it resolves; another `410`/`404` names the bad id explicitly in the check summary. No guesswork either way.

```bash
python -m pytest src/tests/test_ai_reviewer.py src/tests/test_reflect_triage.py \
                 src/tests/test_reflection_metrics.py -q     # 22 passed
python -m flake8 scripts/reflect_triage.py scripts/ai_reviewer.py   # clean
python -c "import yaml; [yaml.safe_load(open(f)) for f in
  ['.github/workflows/ai-review.yml','.github/workflows/graphify-update.yml']]"   # both parse
```

## Two things the reviewer should weigh, neither of which this PR settles

[...]

Issue #1037, opened 2026-08-21
AI reviewer is permanently down: z-ai/glm-5.2 reached end of life 2026-08-21T09:00Z — the required merge check can no longer pass


P0K5:
## The `AI / NVIDIA NIM review` check cannot pass on any PR, as of today

Found on PR #1036. The workflow job fails in ~15 seconds with:

```
AI reviewer failed: NIM API 410 Gone: {"type":"about:blank","title":"Gone","status":410,
"detail":"The model 'z-ai/glm-5.2' has reached its end of life on 2026-08-21T09:00:00Z
and is no longer available."}
```

**This is not a flake and re-running will not help.** The model was retired by NVIDIA at 09:00 UTC today; every subsequent call returns 410 permanently. `scripts/ai_reviewer.py` exits 1, so the check reports failure with no verdict.

Note the failure signature differs from a genuine BLOCK, and the difference is worth knowing when triaging: on a real verdict (e.g. #1032) the workflow **job succeeds** and a separate check run carries the BLOCK. Here the **job itself fails**. Job-green-check-red is a review verdict; job-red is infrastructure.

## Why this is urgent beyond one PR

`.claude/instructions/governance.md` makes the merge gate non-negotiable:

> **All CI checks pass** — `CI / lint`, `CI / test`, and `AI / NVIDIA NIM review` must show green on the PR's **head commit**.

That check can no longer go green for anyone. **Every PR from now until this is fixed requires an explicit human override to merge** — which means the gate has stopped discriminating: an override becomes routine, and a genuine BLOCK becomes indistinguishable from the outage. That is a worse failure than the outage itself, and it starts immediately.

## Blast radius — four call sites, not one

`z-ai/glm-5.2` is referenced in:

| File | What breaks |
|---|---|
| `.github/workflows/ai-review.yml:34` | `NVIDIA_NIM_MODEL` — **the required merge check** |
| `.github/workflows/graphify-update.yml:117` | graphify graph update |
| `.github/workflows/graphify-update.yml:129` | graphify graph update (second job) |
| `scripts/reflect_triage.py:151` | the `reflect` skill's default model |

`agents/reviewer_prompt_glm52.md` is also named for the retired model. The prompt itself is probably portable, but the filename will mislead once the model changes; `scripts/ai_reviewer.py:613` hardcodes that path.

[...]

PR #1039, opened 2026-08-21
Make the AI reviewer fail closed when it cannot run

P0K5:
## Linked issue

Refs #1037 (does not close it — the model replacement is still open, see #1038)

## The defect

`scripts/ai_reviewer.py` had **two** fail-open paths. Both `NimDegradedError` and `NimTimeoutError` created a check run with `verdict="PASS"` and returned 0 — so a reviewer that never ran reported a **green required merge check**.

This was observed live, not theorised. On PR #1038 the model timed out on all three attempts (3 × 150s — the 8m26s run), and the check reported:

> **AI Review: PASS** — ⚠️ AI review skipped … *This check passes automatically to avoid blocking PRs during a transient network/gateway issue.*

`conclusion: "success"`. Nothing was reviewed.

## Why this is worse than the outage it was protecting against

A green check that reviewed nothing is strictly worse than a red one:

- **Red** is loud. It forces a conscious override, and the override is recorded (as on #1032).
- **Green** is silent. Every PR passes the gate unreviewed, and the PR history shows a clean `AI Review: PASS` on each one. Nobody has a reason to look.

The code's own comment called the condition "transient", but **nothing distinguished a transient blip from a permanent failure**. Today it converted an indefinite model outage into indefinite silent approval — which is how a bad model choice became invisible instead of obvious.

There was also an inconsistency worth naming: the `410 Gone` path already failed **closed** (`AI Review: ERROR`, red). Same underlying condition — the reviewer cannot run — with opposite gate behaviour depending only on which error the provider happened to return.

## What changed

Both handlers now report `verdict="UNAVAILABLE"`, which `create_check_run()` maps to `conclusion="failure"`, and exit non-zero so the job is red too — matching the 410 path.

The check output distinguishes the two events explicitly, because they are not the same thing and a reader should never have to guess:

| Event | Title | Conclusion |
|---|---|---|
| Reviewer ran, objected | `AI Review: BLOCK` | failure |
| Reviewer ran, approved | `AI Review: PASS` | success |
| **Reviewer could not run** | **`AI Review: UNAVAILABLE`** | **failure** |

The summary states plainly that no code was reviewed and that this is *not a verdict*, and says what to do: re-run if genuinely transient, replace the model if it persists.

## How to test

```bash
python -m pytest src/tests/test_ai_reviewer.py -q   # 13 passed
python -m flake8 src/                               # clean (CI's exact command)
```

Three new tests, pinning the **conclusion mapping** rather than the wording:

| Test | Pins |
|---|---|
| `test_unavailable_verdict_maps_to_a_failing_conclusion` | `UNAVAILABLE` → `failure`, and the distinguishing title |
| `test_pass_still_maps_to_success` | guards against over-correcting — a genuine PASS stays green |
| `test_no_handler_reports_pass_without_a_review` | source-level: no handler may hardcode `verdict="PASS"` |

**Verified by reintroducing the bug rather than trusting the tests' presence.** Reverting `verdict="UNAVAILABLE"` back to `"PASS"` makes `test_no_handler_reports_pass_without_a_review` fail; restoring it passes 13/13. A regression test that does not fail on the bug it names is worth nothing.

## This PR is also a deliberate diagnostic

It branches from `master`, which **still names the retired `z-ai/glm-5.2`**. A `pull_request` workflow runs from its own head ref, so this PR's `AI / NVIDIA NIM review` check exercises glm-5.2 and settles empirically whether this morning's `410 Gone` was permanent or transient — the open question from #1037.

Expected outcomes, all informative:

- **`AI Review: ERROR`, red** → the EOL is real, as the 410's structured message (`"has reached its end of life on 2026-08-21T09:00:00Z"`) indicates. A replacement model is required.
- **A real verdict** → the 410 was transient after all, and glm-5.2 is usable for now — though it remains an announced-retired model and should still be replaced.
- **`AI Review: UNAVAILABLE`, red** → this PR's own fix working as intended.

Note the check on this PR cannot come back falsely green: that is precisely what the diff removes.

## Trade-off, stated rather than hidden

**Genuinely transient NIM blips will now block PRs until re-run**, where before they passed silently. That is the intended cost. A required review that cannot run should stop the merge, and the fix for a real blip is a re-run — cheap, visible, and it leaves a record. The alternative is a gate that cannot be trusted, which is not a gate.

## Not fixed here

- **The model replacement.** `master` still names the retired model; #1038 proposes `deepseek-ai/deepseek-v4-flash-0731`, which **should not be merged as it stands** — that model timed out on every attempt, and with this fix in place it would fail red rather than fake-green. It needs a model that actually responds.
- **The ~8-minute worst case.** `NIM_MAX_ATTEMPTS = 3` × `NIM_READ_TIMEOUT = 150s` is a long time to add to a merge path when a model is failing. Worth revisiting once a working model is chosen; not tuned here, because the right budget depends on that model.
- **`scripts/` is not linted by CI** (`flake8 src/` only) and carries 15 pre-existing `E741`s in `market_arb_scan.py`. Pre-existing on master, out of scope, noted so it is not rediscovered as new.

CI tooling only. No serving code, nothing entering `bracket_evals`, no frozen constant, no powered Pass 2, no BSS value read or quoted.

## How to deploy

- [x] **No deploy needed** (CI tooling only)

`ai_reviewer.py` runs only in GitHub Actions. Nothing on the bot host; no restart, no config change.

## Checklist

- [x] Tests pass locally (13)
- [x] `python -m flake8 src/` clean
- [x] Regression test verified to fail on the reintroduced bug
- [x] PR body links the issue (`Refs #1037` — deliberately not `Closes`, the model swap is still open)
- [ ] **Project board status — COULD NOT BE UPDATED.** `gh` GraphQL is blocked in this session by org egress policy (403).

---


P0K5:
**Status check (housekeeping pass, 2026-08-23): still blocked, not by this PR's logic.**

Traced the current `AI / NVIDIA NIM review` failure on this PR's head commit (`c4a5385`, run [32496372180](https://github.com/P0K5/MeteoEdge/actions/runs/32496372180)):

```
[ai_reviewer] NIM failed repeatedly, review did NOT run: NIM model meta/llama-3.3-70b-instruct
timed out after 3 attempts: HTTPSConnectionPool(host='integrate.api.nvidia.com', port=443):
Read timed out. (read timeout=150)
[ai_reviewer] Done — verdict=UNAVAILABLE (NIM timeout)
```

The fail-closed logic in this PR is working exactly as designed — an unavailable reviewer now produces `conclusion: failure` instead of a false `PASS`. The problem is upstream: `meta/llama-3.3-70b-instruct` (the model swapped in on this branch) times out identically to how `deepseek-ai/deepseek-v4-flash-0731` did on the now-closed #1038 — two different vendors, same 3×150s read-timeout signature. Per #1038's closing comment, that points at the NIM account (quota/rate limit) or request size, not the model id. `z-ai/glm-5.2` itself still returns cleanly (410, in ~14s) — proving auth/connectivity are fine — it's just EOL.

Correction to my own prior assumption: **#1038 was closed, not merged.** Master still has `z-ai/glm-5.2` in `.github/workflows/ai-review.yml`, `graphify-update.yml`, `scripts/ai_reviewer.py`, `scripts/reflect_triage.py`, and `docs/OPERATIONS.md`. This PR (#1039) is the only branch carrying a model swap at all.

**Net effect: no PR — including this one — can currently pass the required review check, regardless of which model we point it at, until the NIM timeout is diagnosed.** The suggested next step from #1038 (a trivial `max_tokens: 5` request against the NIM endpoint to isolate account/quota vs. request-size) hasn't been run yet — it needs direct access to `NVIDIA_NIM_API_KEY`, which I don't have from this session.

Not merging this while its own required check is red — that would violate governance invariant 4 for the PR that's supposed to be enforcing it. Leaving open, board status In review, until the NIM diagnostic is run and a model is confirmed to respond.

P0K5:
Superseded by #1043, which switches to DeepSeek entirely rather than trying another model on NIM's endpoint (which manual probing during #1037's diagnosis showed is unreliable independent of model choice — see #1043's description). This branch's fail-closed logic was carried forward into #1043's rewrite of the same code, so nothing here is lost.

Issue #1053, opened 2026-08-26
Explicitly halt live trading -- no untested lever remains on the current model


P0K5:
## Context

M3 resolved negative (BSS=-0.4123, powered, 2026-08-25). The one remaining untested lever, per-day ensemble sigma, was stood down (#1048/#1049: ΔBSS=+0.0083, below the pre-registered +0.05 bar — `docs/REMEDIATION_PLAN.md`, "RESOLVED 2026-08-26"). Per the decision gate's own rule: "stop the thesis — pivot the model materially or shut the live path down."

The live path has placed zero trades in weeks in practice — the near-certainty entry rule (M4's known replacement target) is anti-selective against an honestly-calibrated model. That's been an emergent property of the entry gate's strictness, never a deliberate, recorded decision. Tech Lead PM direction: make the halt explicit now, independent of and before any pivot decision — it costs nothing since the path is already inert, and leaving it "dormant but technically live" is not the same as a recorded stop.

## Acceptance criteria

- Every station in `src.config.STATIONS` has `yes_enabled=False` and `no_enabled=False` in `station_overrides`, via the SAME production mechanism the `/api/stations/{metar}/toggle` dashboard endpoint already uses (`Database.set_station_override`) — no new kill switch, no new code path.
- Change is reversible per-station via the existing dashboard toggle (unchanged).
- Shadow logging (`bracket_evals`, candidate evaluation, EMOS shadow training) is UNCHANGED — this only stops live order placement (`scanner.py`'s `shadow_yes`/`shadow_no` gate reads exactly these two columns).
- Idempotent: running it again against an already-halted DB reports zero changes.
- Unit tests: halts every station from a clean DB, is a no-op on a second run, preserves an existing `low_no_enabled` value, `--dry-run` reports without writing.

## How to deploy

One-off script, run once against production:
```
python -m src.scripts.halt_live_trading --db data/meteoedge.db --dry-run   # verify first
python -m src.scripts.halt_live_trading --db data/meteoedge.db
```
No service restart needed — `scanner.py` reads `station_overrides` per scan tick, already the live-served table.

Complexity: **Simple** — reuses an existing, tested production DB method; the only new code is the bulk-iteration wrapper and its tests.

P0K5:
Merged as #1054 and run against production \`data/meteoedge.db\`:

\`\`\`
INFO [halt-live] Halted live entries for 1 station(s): KORD
[...]

Issue #998, opened 2026-08-13
dev-workflow PRs are opening under the PM's GitHub account (P0K5), not GITHUB_TOKEN_OPERATIONAL — blocks formal review approval


P0K5:
## What

PRs #995, #996, #997 (this session, 2026-08-13) were all opened with author `P0K5` — the same GitHub account behind `GITHUB_TOKEN_SUPERVISOR` (Tech Lead PM) — instead of the developer identity behind `GITHUB_TOKEN_OPERATIONAL`. All three PRs were opened by junior-dev/mid-dev subagents that were explicitly instructed to use `GITHUB_TOKEN_OPERATIONAL`.

Effect: `gh pr review --approve` from the PM (GITHUB_TOKEN_SUPERVISOR, account P0K5) fails with `Can not approve your own pull request` — GitHub blocks self-review at the API level regardless of which local token variable was intended. This defeats the two-account review-gate design in `.claude/instructions/governance.md` entirely; #995 had to be merged with the approval recorded as a PR comment instead of a formal GitHub review, which is exactly the review-recording gap #976 already flagged (54/60 recent PRs with zero recorded review).

## Suspected root cause

`.claude/instructions/governance.md` states the token rule ("Everyone else uses GITHUB_TOKEN_OPERATIONAL") but neither it nor the `dev-workflow` skill's PR-open step (step 8) shows the actual command that switches `gh`'s active credential (e.g. `export GH_TOKEN="$GITHUB_TOKEN_OPERATIONAL"` before every `gh` invocation, or `gh auth switch`). If the sandbox's `gh` is already authenticated as P0K5 by default (e.g. via a persistent `gh auth login`), any agent that doesn't independently know to override $GH_TOKEN per-command will silently inherit that identity — which is what appears to have happened here across three independent subagent spawns.

## Acceptance criteria

- [ ] Root-cause confirmed: check what `gh auth status` reports by default in a fresh dev-agent shell, and whether $GH_TOKEN / $GITHUB_TOKEN_OPERATIONAL was actually exported at the point `gh pr create` ran in the affected agents' sessions (transcripts for PRs #995/#996/#997 if available).
- [ ] `dev-workflow` skill's PR-open step (and any other step that shells out to `gh`) shows the literal command to select the operational identity, not just a reference to the governance token rule.
- [ ] A regression check: after the fix, a freshly spawned junior-dev/mid-dev opens a PR and `gh pr view <N> --json author` reports the operational account, not P0K5.
[...]

PR #1079, opened 2026-08-28
fix(#1074): Replace crypto rate and 1¢ floor with correct weather rate

### LINKED ISSUE #1074: Fee model uses the crypto rate and a fabricated 1c floor — 20x overstated at the rail
## Context

Part of #1073 (Phase 1). `src/strategy/fee.py` models the Polymarket taker fee as
`max(1.0, 7.0 * p * (1 - p))` cents per contract. Checked against the published
schedule (https://docs.polymarket.com/trading/fees, `fee = shares * feeRate * p * (1 - p)`),
this is wrong in two independent ways:

1. **`0.07` is the crypto rate.** Weather markets are **`0.05`**.
2. **The `max(1.0, ...)` floor is fabricated.** No floor exists — the only
   published limit is 0.00001 USDC rounding precision.

At 99c the model returns **1.00c**; the real fee is **0.05c** — a **20x**
overstatement. This is larger than the entire edge the function is used to
evaluate: `backtest_results/edge_hunt_2026-08-27.md` shows the fabricated floor
alone is the difference between "no edge" and "+0.86c/contract" on the rail
population. Makers are never charged at all, and the function has no way to
express that.

The module's own docstring instructs someone to validate it against fill history
via `scripts/fee_calibration.py` and update the docstring with sample size and
date. That was never done.

## Acceptance criteria

- [ ] `estimate_fee_cents` implements `100 * feeRate * p * (1 - p)` with **no floor**
- [ ] The weather rate `0.05` is the default; the rate is a named constant, not inline
- [ ] A `maker: bool = False` (or equivalent) parameter returns `0.0` for maker fills
- [ ] Docstring cites the published schedule URL and the date checked, and drops the
      unvalidated "0.07 / variance" rationale
- [ ] Callers in `src/execution/` still type-check and pass their existing tests

## Technical notes

- `src/strategy/fee.py` is the only definition; callers are
  `src/execution/order_manager.py` (L222, L885, L938, L1115),
  `src/execution/position_tracker.py` (L399, L456, L639, L697, L849).
- `src/config.py:298` documents a 2% buffer as covering "highest ~2% at extreme
  prices" — that rationale is now inverted (fees are *lowest* at the extremes).
  Update the comment; do **not** change `POSITION_SIZE_WITH_FEES` behaviour in
  this issue.
- Do not touch live trading behaviour — the live path is halted (#1053/#1054).

## Test requirements

- Unit: `estimate_fee_cents(99) == pytest.approx(0.0495, abs=1e-4)`
- Unit: `estimate_fee_cents(50) == pytest.approx(1.25, abs=1e-4)`
- Unit: fee is monotonically decreasing as price moves from 50 toward either extreme
- Unit: no output is ever clamped up to 1.0 (regression on the removed floor)
- Unit: maker path returns exactly `0.0` at every price

## Dependencies

None.

P0K5:
## Summary

Fixes issue #1074 by replacing the incorrect fee model with the published Polymarket schedule.

Changes:
- Replaces `max(1.0, 7.0 * p * (1-p))` with `100 * 0.05 * p * (1-p)` for weather markets
- Removes the fabricated 1¢ floor (no floor exists in published schedule)
- Uses the correct 0.05% taker rate (not 0.07% crypto rate)
- Adds `maker: bool = False` parameter; makers pay 0% (0.0¢)
- Updates config.py comment to reflect fees are lowest at extremes

Fee examples:
- 99¢: 0.0495¢ (was 1.00¢ — 20x overstatement)
- 50¢: 1.25¢ (was 1.75¢)
- 1¢: 0.0495¢ (was 1.00¢ — 20x overstatement)

All existing callers continue to work (maker parameter defaults to False).

## Test plan

- [x] All 9 fee unit tests pass (new formula at 1¢, 50¢, 99¢; monotonicity; maker=0.0; no floor)
- [x] All 7 cascading test failures fixed (promotion_gate hand-computed values updated; prob_cap_shadow_report thresholds adjusted for lower fees)
- [x] Trade cost accounting tests pass (order_manager.py calls still work)
- [x] Promotion gate tests pass (fee is correctly used in breakeven calculations)
- [x] Verifies published schedule: https://docs.polymarket.com/trading/fees (checked 2026-08-28)

## How to deploy

**No deploy needed.** The fee function is already live-imported by src/execution/order_manager.py and src/execution/position_tracker.py via standard Python import, but:

1. The live trading paths are currently halted (#1053/#1054), so this change has no immediate operational impact
2. The fee function changes are purely mathematical — it replaces one calculation with another; no database migrations, config reloads, or service restarts are required
3. When live trading resumes, the corrected fee model will be used automatically at the next process restart (no redeployment needed beyond a standard code push)
4. The `maker` parameter is optional with a default value, so all existing call sites work without modification

Closes #1074

🤖 Generated with Claude Code


P0K5:
**Request changes.**

## [blocker] CI \`Test (pytest)\` is failing — 7 tests, caused by this PR

The fee-formula change is correct in isolation, but downstream tests that
hand-compute breakeven win rates / promotion-bar values against the OLD fee
constant were not updated, and now fail on the new numbers:

- \`test_promotion_gate.py::TestBreakevenWinRate::test_60_cents_hand_computed\` — expected 0.6168, got 0.612
- \`test_promotion_gate.py::TestBreakevenWinRate::test_75_cents_hand_computed\` — expected 0.763125, got 0.759375
- \`test_promotion_gate.py::TestComputePromotionBar::test_green_eligible_station\` — expected 0.665925, got 0.661375
- \`test_prob_cap_shadow_report.py::TestSimulateCapValues::test_edge_channel_discovers_new_no_candidate\`
- \`test_prob_cap_shadow_report.py::TestSimulateCapValues::test_unresolved_new_candidate_excluded_from_pnl\`
- \`test_prob_cap_shadow_report.py::TestSimulateCapValuesExtraActualHighLookup::test_extra_lookup_resolves_newly_discovered_candidate\`
- \`test_prob_cap_shadow_report.py::TestSimulateCapValuesExtraActualHighLookup::test_settled_row_actual_high_takes_precedence_over_extra_lookup\`

Please re-derive the hand-computed expected values in the three
\`test_promotion_gate.py\` cases against the corrected \`0.05\` weather rate /
no-floor formula (don't just bump the numbers to whatever the new run
produces — show the arithmetic in the test or a comment, the way the
existing hand-computed tests do), and diagnose why the \`prob_cap_shadow\`
candidate-discovery tests regressed — that file's assertions are about
candidate discovery/exclusion logic, not fee values directly, so this may be
an actual behavior change worth understanding rather than a number to update.

## [blocker] Missing "How to deploy" section

PR body has no deploy section per the template. This changes a live-imported
pricing function with multiple callers in \`src/execution/\` — state explicitly
whether anything needs restarting/migrating, or "No deploy needed" with why
that's true given the callers.

## [question] PR opened under account \`P0K5\`, not the operational identity

This is the exact failure mode #998 describes — I can't leave a formal
\`gh pr review --approve\` here (GitHub blocks self-review), so approval on
this PR will need the comment-recorded fallback once the blockers above are
fixed. #998 is being worked in parallel; not this PR's fault, just flagging
why the eventual approval will look different from a normal PR.

Moving #1074 back to **In progress**.

P0K5:
Approved (via comment — GitHub API collision).

Rationale: this PR was opened under P0K5 (the PM's own account) before #1080
(the token-export fix) landed, so \`gh pr review --approve\` is blocked by
GitHub's self-review rule — this is the sanctioned fallback #1080 documents,
not a shortcut.

All blockers from the first pass are resolved: the 7 cascading test failures
are fixed with re-derived arithmetic (verified: fee(60)=1.2¢→breakeven 0.612,
fee(75)=0.9375¢→breakeven 0.759375, fee(65)=1.1375¢→breakeven 0.661375, all
match the CI failure output exactly), the prob_cap_shadow_report no_ask
79→80 shift is a legitimate re-parameterization (the lower fee pushed
no_ask=79 past the admission threshold at the base cap, breaking the test's
"sub-threshold at deployed cap" premise — 80c restores it, no change to the
underlying discovery logic), and the deploy section is filled. CI green
(lint, test, DeepSeek review, gitleaks all pass on the latest SHA).

Merging.

Issue #1077, opened 2026-08-28
Persist real order-book bid/ask and depth into bracket_evals at scan time


P0K5:
## Context

Part of #1073 (**Phase 2 — the only real unlock**). This project has never
recorded an order book.

Across 131,435 in-window rows, `yes_ask + no_ask == 100` on **99.34%** of them. A
real two-sided book always sums to *more* than 100 — you pay a spread on both
sides. What `bracket_evals` stores is a single Gamma `outcomePrices` mid split
into complements, then clamped. **There is no executable price anywhere in the
archive, and no depth at all.**

The consequence for the epic's question: the apparent rail edge is **0.86c —
less than one 1c tick**. A single tick of spread erases it. Whether it survives
is decided entirely by (a) the true sub-penny ask and (b) whether any size rests
there. Neither is knowable without forward book capture.

`scripts/market_arb_scan.py` already states this limitation in its own docstring:
"`yes_ask_size` and `no_ask_size` are hardcoded to 0 unless
`ENABLE_CLOB_ENRICHMENT=true`, and they are not persisted to `bracket_evals` at
all. This script therefore measures HOW OFTEN and HOW WIDE an arb was quoted,
never how much size it would have absorbed."

## Acceptance criteria

- [ ] On every scan poll, top-of-book **bid and ask** (sub-penny floats) and
      **per-level depth** are fetched and persisted to `bracket_evals`
- [ ] At least top 3 levels per side are stored, with size at each
- [ ] Book fetch failures degrade gracefully — the row is still written with book
      fields NULL, never a fabricated value (the #1028 lesson)
- [ ] A NULL book field is distinguishable in the log from a genuine empty book
- [ ] Rate-limit safe: measured request volume documented on the issue before
      enabling continuously
- [ ] `ENABLE_CLOB_ENRICHMENT` (or successor) gates it, default off, with the
      production enable recorded as a dated decision

## Technical notes

- `src/data/polymarket.py:172` `get_orderbook(token_id)` and L186
  `fetch_orderbooks_batch(...)` already exist and already parallelize. They are
  called only at execution time (`order_executor.py:45`, `live_trader.py:118`,
  `order_manager.py:811`) and never at scan time. **Reuse them — do not write a
  third book client.**
- **Rate limits are the real design constraint**: 30 stations x ~11 brackets x 288
  polls/day is ~95k book fetches/day if applied naively. Options to evaluate and
[...]

PR #1081, opened 2026-08-28
fix(#1076): preserve sub-penny prices alongside the clamped cents fields

### LINKED ISSUE #1076: Price clamp destroys sub-penny prices, making the rail question unanswerable
## Context

Part of #1073 (Phase 1). **This is the issue that makes the rest of the epic
possible.**

`src/strategy/scanner.py:235-237` (and again at L355, L369) stores market prices as:

```python
max(1, min(99, round(price * 100)))
```

Any true price below 1.5c is recorded as `1`; anything above 98.5c as `99`. But
Polymarket's tick size **tightens to $0.001 when price > 0.96 or < 0.04** — exactly
the region where 74% of our brackets sit. The venue can legally quote 0.1c and we
record 1c.

The measured consequence (`backtest_results/edge_hunt_2026-08-27.md`): the 1c
longshot leg resolves YES 0.094% against a 1.0% breakeven, apparently
+0.86c/contract on 5,318 rows. Seen as a ladder it is a dutch book of
`Sum(yes_ask) - 100`, median **+9c** — and that 9c is *exactly* 10 legs times the
1c minimum we impose ourselves:

| true longshot leg | Sum(yes_ask) | dutch book | per-contract edge |
|---|---|---|---|
| 1.0c | 109c | +9.0c | +0.818c (matches measured) |
| 0.1c | 100c | 0c | +0.000c |

So the entire apparent edge is either real or a reflection of our own clamp, and
**no archive can distinguish the two** — `bracket_evals` has stored the clamped
integer since day one.

## Acceptance criteria

- [ ] The unclamped price is preserved as a float (e.g. `yes_price_raw`,
      `no_price_raw`) alongside the existing clamped integer fields
- [ ] Raw floats are persisted to `bracket_evals` and to `scan_decisions`
- [ ] Existing `yes_ask_cents` / `no_ask_cents` integer fields keep their current
      values and semantics — no consumer changes behaviour in this issue
- [ ] Rounding is documented at the definition: what is clamped, why the integer
      field is retained, and which field callers should prefer
- [ ] Sub-cent prices survive end to end and are visible in a fresh
      `logs/bracket_evals.*.jsonl` row

## Technical notes

- `src/strategy/scanner.py` L220-237 (`_safe_price`, `_parse_bracket`), L355, L369.
- Related and adjacent: **#1029** (`_safe_price` should skip and alarm rather than
  substitute 0.5). Coordinate — both touch the same function. #1029 was held for
  M3; M3 is over, so it is now unblocked.
- This changes what enters `bracket_evals`. Under the M3 regime that would have
  "split the window" — that constraint is retired with the thesis (#1073 Phase 0),
  so it is safe now and must NOT be deferred on those grounds again.
- Additive schema change only; follow the `src/data/db.py:379` migration pattern.

## Test requirements

- Unit: a Gamma payload quoting `0.003` yields `yes_price_raw == 0.003` and
  `yes_ask_cents == 1`
- Unit: `0.997` yields `no_price_raw == 0.997` and the integer field still `99`
- Unit: mid-range prices are unchanged in both fields
- Round-trip: a written `bracket_evals` row parses back with full float precision
- Regression: existing consumers of `yes_ask` / `no_ask` see identical values

## Dependencies

None technically. Sequence with #1029 to avoid conflicting edits to `_safe_price`.

ClaudioSenior:
## Linked issue
Closes #1076

## What changed
`src/strategy/scanner.py`'s `_safe_price`/`parse_bracket_from_market` (and
the CLOB-enrichment path at `_enrich_from_clob`) clamp every ask price to
`max(1, min(99, round(price * 100)))`. Any true price below 0.5c or above
99.5c is recorded as 1c/99c — exactly the region where Polymarket's own tick
size tightens to $0.001 (price > 0.96 or < 0.04), which is where 74% of our
brackets sit (issue #1076's measured dutch-book artifact). Because the DB
has only ever stored the clamped integer, no archive can tell a real edge
apart from our own rounding.

This PR adds the raw float alongside the existing clamped fields — additive
only, no consumer behaviour change:

- `Bracket.yes_price_raw` / `Bracket.no_price_raw` (`src/model/envelope.py`),
  populated in `parse_bracket_from_market` and updated in `_enrich_from_clob`
  alongside the existing clamped `*_ask_cents` fields.
- `scan_markets`'s snapshot dict carries `yes_price_raw`/`no_price_raw`
  through to both downstream sinks:
  - `bracket_evals.jsonl` (`run.py`'s `_write_bracket_evaluations`)
  - `scan_decisions` (`Database.upsert_scan_decision` / `get_scan_decisions`),
    via an additive `ALTER TABLE` migration following the `db.py:379` pattern
    (including the pre-#912 CHECK-constraint rebuild's column list, so an old
    DB doesn't silently drop the new columns on that one-time rebuild).
- `docs/DB_SCHEMA.md` documents both fields at `scan_decisions` and the
  migration note.
- `Bracket`'s docstring in `envelope.py` documents the clamp itself (what,
  why, which field callers should prefer) per the issue's acceptance
  criteria.

Out of scope (explicitly, per the Tech Lead PM): #1029's behaviour change
(`_safe_price` skipping/alarming instead of substituting 0.5 on a missing
price) is not touched here — the 0.5 fallback is preserved verbatim, now
just also visible in the raw field.

## How to test
```
python -m pytest src/tests/test_scanner.py src/tests/test_scan_decisions.py \
  src/tests/test_bracket_evals.py src/tests/test_scan_decisions_gate_verdict_check_migration.py \
  src/tests/test_analysis_endpoint.py -v
```
New/updated tests:
- `test_scanner.py::TestParseBracketFromMarket` — sub-penny YES/NO price
  clamps to 1c/99c in `*_ask_cents` but preserves full float precision in
  `*_price_raw`; mid-range prices match in both fields.
- `test_scan_decisions.py::TestScanDecisionsUpsert` — `yes_price_raw`/
  `no_price_raw` default to `NULL` when omitted (pre-#1076 rows), and
  round-trip through the DB with full float precision.
- `test_bracket_evals.py` — a sub-penny raw price survives a full
  `_write_bracket_evaluations` write + JSONL round-trip.

Full suite (`pytest src/tests -q`) passes identically to master modulo two
pre-existing, unrelated Windows-only flakes (`os.unlink` on an open sqlite
handle in `finally:` blocks, and a `symlink_to` privilege error) — confirmed
by reproducing both on `master` before this change.

## How to deploy
- [ ] **Code only** — `git pull` + restart `meteoedge-run`
- [x] **DB schema change** — migration applies automatically on service
      start (idempotent `ALTER TABLE ... ADD COLUMN`, `Database._migrate`);
      no manual step
- [ ] **Post-deploy validation** — a fresh `logs/bracket_evals.*.jsonl` row
      should carry non-null `yes_price_raw`/`no_price_raw`


P0K5 (approved):
Approve. Reviewed as a manual deep-dive per protocol (DB schema migration). Verified: additive-only — `yes_ask`/`no_ask` and every gate/trading consumer are byte-for-byte unchanged; the new `yes_price_raw`/`no_price_raw` fields correctly thread through `Bracket` → `parse_bracket_from_market`/`_enrich_from_clob` → both sinks (`bracket_evals.jsonl` via run.py, `scan_decisions` via db.py); the migration follows the `db.py:379` additive pattern AND correctly updates the pre-#912 CHECK-constraint rebuild's column list (easy to miss, wasn't missed); sub-penny survival is tested end-to-end (0.003/0.997 round-tripping through both sinks with full float precision), not just at the parse boundary. #1029's scope (the 0.5 fallback) is untouched, exactly as directed. CI green (lint, test, DeepSeek review, gitleaks). Merging — this unlocks #1077.

Issue #1073, opened 2026-08-28
EPIC: Retire the forecast thesis; instrument the order book to settle the rail question


P0K5:
## Context

The forecast thesis is finished. M3 returned **BSS = −0.4123**, M3b **FAIL**, M3c
**NULL**, and the forecast-stack × σ pivot came in at **ΔBSS = −0.0031** against a
+0.15 bar — shutdown by pre-registration. Live trading is halted unconditionally
(#1053/#1054).

`backtest_results/edge_hunt_2026-08-27.md` adds the test that had never been run,
and it closes the last door:

**The market *encompasses* the model.** Every gate so far asked "is the model
better than the price?" — a test of *standalone* skill. It is not the same
question as "does the model carry information the price lacks?", which is what a
combination trade would need (Fair–Shiller / Granger–Ramanathan encompassing).
On the M3 gate's own population, importing its own population code:

```
logit P(YES) = a + b·logit(p_market) + c·logit(p_model)
  b = +1.0373   (market)
  c = +0.0125   (model)   95% station-day block CI [−0.0875, +0.1281], P(c>0)=0.59
```

`c` is indistinguishable from zero, and **out of sample adding the model makes
the combined forecast worse** (−0.00024). Note `b ≈ 1.04`, `a ≈ 0.03`: the market
needs no recalibration. No better stack, σ lever, or EMOS promotion can beat an
encompassing coefficient of 0.01.

**This epic does not attempt to rescue that thesis.** It retires it, fixes two
real correctness bugs found along the way, and builds the one instrument that can
settle a *different* and much narrower question.

## The different question

Ignoring the model entirely, the 1¢ longshot leg resolves YES **0.094%** of the
time (n=5,318, 638 station-days) against a **1.0% breakeven** — apparently
+0.86¢/contract, robust across every time split. Seen as a ladder it is a dutch
book: `profit = Σyes_ask − 100`, median **+9¢**.

**That 9¢ is exactly 10 legs × the 1¢ minimum we impose ourselves.**
`scanner.py:235-237` stores `max(1, min(99, round(p·100)))`, so any true price
below 1.5¢ is recorded as `1`. Polymarket's tick tightens to **$0.001** above
0.96 / below 0.04 — precisely this region — so the venue may legally be quoting
0.1¢ and we would record 1¢.

| true longshot leg | `Σyes_ask` | dutch book | per-contract edge |
|---|---|---|---|
[...]

PR #1086, opened 2026-08-28
feat(#1077): persist real order-book bid/ask + depth into bracket_evals

### LINKED ISSUE #1077: Persist real order-book bid/ask and depth into bracket_evals at scan time

ClaudioSenior:
Closes #1077

## What changed

`bracket_evals` has never recorded a real order book -- `yes_ask`/`no_ask` are
a single Gamma mid-price split into complements (99.34% of in-window rows sum
to exactly 100), never a two-sided quote, and there was no depth at all.

- Extended `Bracket` (`src/model/envelope.py`) with `{yes,no}_bid_raw`,
  `{yes,no}_bid_levels` / `{yes,no}_ask_levels` (top-3, best-first,
  `[{"price","size"}, ...]`), and `{yes,no}_book_status`.
- Rewrote `_enrich_from_clob` (`src/strategy/scanner.py`) to populate these
  from the order book it was already fetching for the ask price alone --
  **no new HTTP requests**. Failures degrade gracefully: every book field
  stays `None` on failure, never a fabricated price. `yes_book_status`/
  `no_book_status` distinguish `"fetch_failed"` from a genuine `"empty_book"`
  (a real CLOB `/book` response always carries `bids`/`asks` keys, even
  empty; only a failed fetch collapses to a bare `{}`) -- done without
  changing `fetch_orderbooks_batch`'s tested `{}`-on-failure contract, which
  other callers (`_log_open_position_snapshots`) already rely on.
- Wired the new fields through `scan_markets`' snapshot dict and
  `_write_bracket_evaluations` (`src/scripts/run.py`) so they land in
  `bracket_evals.jsonl`.
- **Rate-limit fix + measurement** (found while sizing this change):
  `run.py`'s scan-time token collection iterated *every* market on the Gamma
  weather tag globally (~2,100 as of 2026-08-28, live-queried), not just our
  30 stations -- an order of magnitude past the issue's ~95k/day budget.
  Filtered to `is_highest_temp_market`/`is_lowest_temp_market` matches only,
  the same filter `scan_markets()` itself applies. Full numbers and
  recommendation recorded in `docs/OPERATIONS.md`.
- `ENABLE_CLOB_ENRICHMENT` (existing flag, already default off) gates all of
  this, same as before. **No production-enable decision has been made.**

## Why

Part of #1073 Phase 2. The epic's apparent rail edge (0.86c) is smaller than
one tick -- whether it survives depends entirely on the true sub-penny ask and
bid/depth, neither previously recorded anywhere.

## How to test

```
python -m pytest src/tests/test_scanner.py src/tests/test_batch_clob.py src/tests/test_bracket_evals.py -q
```

New coverage:
- `TestEnrichFromClobBookDepth` (`test_batch_clob.py`) -- mocked book persists
  bid/ask/depth at full float precision; a failed fetch never substitutes a
  price; `fetch_failed` vs `empty_book` are distinguishable; no-token-id
  leaves status `None` (not attempted); flag-off leaves everything untouched.
- `TestScanMarketsOrderBookIntegration` (`test_scanner.py`) -- one poll cycle
  against a recorded book fixture through `scan_markets()` produces snapshot
  rows with book fields populated (and explicit `None`s when disabled).
- `TestBracketEvalsWrite`/new tests (`test_bracket_evals.py`) -- book fields
  round-trip through the JSONL write at full precision; default to explicit
  `null` (not missing keys) when absent; `fetch_failed`/`empty_book` remain
  distinguishable end-to-end.

## Trade-offs / uncertainties

- Kept the rate-limit fix (token-collection filter) in this PR rather than a
  separate issue since it directly serves this issue's "rate-limit safe" AC
  and is a small, low-risk, single-loop change reusing existing filters --
  flag if you'd rather it be split out.
- `ENABLE_CLOB_ENRICHMENT` stays off; a follow-up decision (with a dated
  comment on this issue or a new one) is still needed before any continuous
  production enable, per the issue's own AC.

## How to deploy

- [x] Code-only change -- restart the poll loop (`run.py`) process to pick it
      up. No DB migration (bracket_evals is a JSONL log, not a SQL table). No
      config/env change required (`ENABLE_CLOB_ENRICHMENT` unchanged, still
      default off).


P0K5 (approved):
Approve. Reviewed the full diff (not just the AI summary) given this is the epic's other DB/schema-adjacent, execution-path change. Verified: fetch_failed vs empty_book is correctly distinguished without breaking fetch_orderbooks_batch's existing {}-on-failure contract that other callers (_log_open_position_snapshots) depend on; every book field stays genuinely None on failure, never fabricated (the #1028 lesson, done right); best-ask/best-bid selection is actually more correct than the code it replaced (properly sorts levels before taking top-3, where the old code just took the API's raw first-3 unsorted); ENABLE_CLOB_ENRICHMENT stays default-off, no production-enable decision made, exactly as scoped. The bundled rate-limit fix (token collection was iterating ~2,100 global markets instead of 30 configured stations, ~1M req/day vs the ~95k/day budget) is a legitimate, low-risk, in-scope catch — good that it was found and documented with real numbers rather than left for a future surprise. Test coverage is unusually complete: fetch-failure, empty-book, no-token-id, disabled-flag, and full end-to-end JSONL round-trip are all separately covered. CI green. This completes epic #1073's Phase 1 and Phase 2 — Phase 3 (pre-registration) is gated on the Tech Lead PM's sign-off, already drafted and pending. Merging.

Issue #1099, opened 2026-09-19
Epic A: Copy-trading wallet screening pipeline (productionize spike)


P0K5:
## Context

Epic A from `docs/design/copy-trading-architecture.md`. Productionizes the
screening logic from the copy-trading spike (#1097, PR #1098) into a
scheduled, persisted pipeline. This is the foundation every other
copy-trading epic depends on — no execution work should start before this
produces a stable, queryable candidate list.

Turns `src/scripts/copy_trade_backtest.py`'s median-ROI + `--min-trades` +
`--flat-stake` logic into a scheduled job instead of an ad hoc CLI run,
and adds the stability check the spike had to do manually: `0xd3b034d7...`
looked like the best candidate in one run and reversed completely
(resolved trades 7,498 → 2,271, median ROI +33.4% → -100%) 15 hours later,
because `get_wallet_trades()` is capped at 20,000 trades and appears to
return a recency-biased window for wallets whose true history exceeds
that. A wallet must screen consistently across ≥2 runs before it's
eligible to be followed (Epic B).

## Acceptance criteria

- [x] A scheduled job (cadence TBD — see open question below) pulls the
      leaderboard (`src/data/polymarket_traders.py::get_leaderboard`,
      `/v1/leaderboard`) and screens candidates using the same median-ROI
      scoring as `copy_trade_backtest.py` (reuse the module, don't
      reimplement the math).
- [x] Every run persists one row per (wallet, run) to a new
      `copy_wallet_candidates` table (`src/data/db.py` conventions:
      `CREATE TABLE IF NOT EXISTS` + versioned `_migrate`) — append-only,
      never overwrite prior runs.
- [x] A wallet is only marked eligible-to-follow if its median ROI and
      resolved-trade count are consistent (defined threshold — e.g. median
      ROI sign unchanged and resolved-trade count within some tolerance)
      across its last 2+ screening runs.
- [x] Rate-limit budget for `data-api.polymarket.com`/`gamma-api.polymarket.com`
      is explicitly decided and documented (see architecture doc
      "Isolation" section) — this pipeline must not starve or be starved
      by the weather strategy's own polling if both run concurrently.
- [x] Unit tests covering: persistence of a screening run, the stability
      check's pass/fail logic (including a fixture reproducing the
      `0xd3b034d7` reversal pattern), and rate-limit budget enforcement.

## Technical notes

- Reuse `backtest_wallet()`, `_stats()`, and the median-ROI sort from
[...]

PR #1111, opened 2026-09-19
Add copy_wallet_candidates table and screening-run persistence

### LINKED ISSUE #1108: Add copy_wallet_candidates table and screening-run persistence to db.py
Part of epic #1099.

## Context

Epic A needs to persist every wallet-screening run so the stability check
(story 2 of this epic) can compare a wallet's consecutive runs and decide
eligibility. This story adds only the schema and persistence methods,
following the existing `scan_decisions` / `Database.upsert_scan_decision`
pattern in `src/data/db.py` — with one deliberate difference: this table is
**append-only** (every run keeps its own row), not an upsert-per-day table,
because the stability check in story 2 needs to diff a wallet's last two
runs, and `docs/design/copy-trading-architecture.md`'s `copy_wallet_candidates`
spec is explicit that this must never overwrite prior runs.

## Acceptance criteria

- [ ] New `copy_wallet_candidates` table added to `_DDL` in `src/data/db.py`
      (append it near the `scan_decisions` block, `src/data/db.py:295-345`):
      `id INTEGER PRIMARY KEY AUTOINCREMENT`, `address TEXT NOT NULL`,
      `window TEXT NOT NULL`, `screened_at TEXT NOT NULL`,
      `n_buy_trades INTEGER NOT NULL`, `n_resolved INTEGER NOT NULL`,
      `win_rate REAL`, `mean_roi REAL`, `median_roi REAL`,
      `mirrored_dollar_pnl REAL`, `flat_dollar_pnl REAL`, `flat_stake REAL`,
      `slippage_bps REAL NOT NULL`,
      `eligible_to_follow INTEGER NOT NULL DEFAULT 0`.
      Add `CREATE INDEX IF NOT EXISTS idx_copy_wallet_candidates_address_screened
      ON copy_wallet_candidates(address, screened_at)` immediately after,
      matching the `scan_decisions` / `idx_scan_decisions_station_date`
      pairing right above it.
- [ ] New `Database.insert_wallet_screening(...)` method: keyword-only args
      mirroring `insert_candidate`'s style (`src/data/db.py:1219-1269`),
      one column per field above (minus `id`), always a plain `INSERT`
      (never `INSERT OR REPLACE` / upsert — this is the key difference from
      `upsert_scan_decision`), returns `cur.lastrowid`, takes `self._lock`
      and commits, same as every other writer in the class.
- [ ] New `Database.get_recent_wallet_screenings(address: str, limit: int = 2)
      -> list[dict]` returns the wallet's most recent `limit` rows ordered
      `ORDER BY id DESC` (not `screened_at DESC` — two rows can share a
      timestamp; `id` is the only guaranteed-monotonic tiebreak), each row
      converted with `dict(row)` (the idiom used throughout `db.py`, e.g.
      `src/data/db.py:1180`, `:1204`, `:1733`). Returns `[]` for an unknown
      address, never raises.
- [ ] No existing table, migration, or caller is touched.

## Technical notes

- DDL block to extend: `src/data/db.py:295-345` (`scan_decisions`) — add the
  new `CREATE TABLE` / `CREATE INDEX` pair right after it, inside the same
  `_DDL` string (statements execute via `.strip().split(";")` at
  `src/data/db.py:372-375`, so multiple statements in one block is the
  existing convention, not a new one).
- Insert-method style to mirror: `Database.insert_candidate`,
  `src/data/db.py:1219-1269`.
- Validation/locking style to mirror (but NOT the upsert-per-key behavior):
  `Database.upsert_scan_decision`, `src/data/db.py:1294+`.
- This table is intentionally separate from `scan_decisions`/`candidates` —
  it is copy-trading-only, per the architecture doc's isolation requirement.

## Test requirements

Add to `src/tests/test_db.py` (existing file):
- `insert_wallet_screening` followed by `get_recent_wallet_screenings`
  returns the inserted row with matching field values.
- `get_recent_wallet_screenings(..., limit=N)` returns at most N rows, most
  recent (`id DESC`) first.
- `get_recent_wallet_screenings` on an address with zero rows returns `[]`.
- Two `insert_wallet_screening` calls for the **same** address both persist
  as separate rows (append-only, not overwritten) — this is the behavior
  that differentiates this table from `scan_decisions` and needs an
  explicit regression test.

## Dependencies

None — first story in epic #1099.

## Labels

backend, Mid

ClaudioSenior:
## Linked issue
Closes #1108

## What changed
Adds the schema and persistence layer for the copy-trading wallet-screening
pipeline (epic #1099, story 1):

- New `copy_wallet_candidates` table in `_DDL` (`src/data/db.py`), appended
  right after the `scan_decisions` block, with columns
  `id, address, window, screened_at, n_buy_trades, n_resolved, win_rate,
  mean_roi, median_roi, mirrored_dollar_pnl, flat_dollar_pnl, flat_stake,
  slippage_bps, eligible_to_follow` plus a matching
  `idx_copy_wallet_candidates_address_screened` index, mirroring the
  `scan_decisions` / `idx_scan_decisions_station_date` pairing.
- `Database.insert_wallet_screening(...)` — keyword-only args mirroring
  `insert_candidate`'s style, always a plain `INSERT` (never
  `INSERT OR REPLACE` / upsert). This is deliberately append-only, unlike
  `upsert_scan_decision`: the stability check in story 2 of epic #1099 needs
  to diff a wallet's last two screening runs (e.g. the `0xd3b034d7` reversal
  — resolved trades 7,498 → 2,271, median ROI +33.4% → -100% across two runs
  15 hours apart), which requires every run to keep its own row.
- `Database.get_recent_wallet_screenings(address, limit=2) -> list[dict]` —
  returns the wallet's most recent `limit` rows ordered `ORDER BY id DESC`
  (not `screened_at DESC`, since two rows can share a timestamp and `id` is
  the only guaranteed-monotonic tiebreak). Returns `[]` for an unknown
  address, never raises.

No existing table, migration, or caller is touched. No `_migrate()` entry
needed — this is a brand-new table covered by the existing
`CREATE TABLE IF NOT EXISTS` in `_DDL`, same as `poll_runs`.

## How to test
```
pytest src/tests/test_db.py -v
```
New `TestCopyWalletCandidates` class covers:
- table existence
- insert + get round trip (all field values match)
- `limit` truncation
- `id DESC` ordering when two rows share a `screened_at` timestamp
  (reproduces the `0xd3b034d7` fixture pattern)
- unknown address / zero rows both return `[]`, never raise
- two inserts for the same address persist as two separate rows
  (append-only regression test — the key behavior that differs from
  `scan_decisions`)

Full suite run: `pytest -q` — 35 pre-existing failures reproduce identically
with this branch's changes stashed out (WSL/bash environment issues in
`test_ensure_worktree.py`, unrelated `test_emos_retrain.py` /
`test_shadow_low_rollout.py` column-migration tests, etc.) — none touch
`db.py`'s new table or are introduced by this PR.

## How to deploy
- [x] **DB schema change** — migration applies automatically on service
  start: the new table + index are created via the existing
  `CREATE TABLE IF NOT EXISTS` / `CREATE INDEX IF NOT EXISTS` block executed
  in `Database.__init__`, same mechanism as every other table (e.g.
  `poll_runs`). No manual migration step, no data backfill (table starts
  empty; nothing writes to it yet — see epic #1099 story 2+).

## Checklist
- [x] Tests pass locally (`pytest src/tests/test_db.py -v`)
- [x] Docs updated if this PR adds/changes env vars, DB schema, API
  endpoints, or run modes — N/A, no doc references this table's schema yet
  beyond `docs/design/copy-trading-architecture.md`, which already specifies
  it


P0K5 (approved):
Approving — matches story #1108's acceptance criteria exactly: schema/index mirror the `scan_decisions` pattern, `insert_wallet_screening` is a plain append-only INSERT (correctly not an upsert), `get_recent_wallet_screenings` uses `id DESC` (correct tiebreak reasoning for shared timestamps). Test coverage is thorough, including the append-only regression test and the `0xd3b034d7` reversal fixture ordering test. AI reviewer PASS, all CI green. How-to-deploy section correctly notes no manual migration needed (CREATE TABLE IF NOT EXISTS).

PR #1125, opened 2026-09-19
Add copy_wallets_followed, copy_signals, copy_positions tables

ClaudioSenior:
## Linked issue
Closes #1121

## What changed
Adds the three copy-trading persistence tables epic B (#1101 — signal detection & flat-stake paper execution) needs, following the exact `copy_wallet_candidates` pattern from #1108/PR #1111 (`CREATE TABLE IF NOT EXISTS` in `Database._DDL`, keyword-only insert/get/update methods, `dict(row)` getters). Schema + persistence only — no polling or execution logic, per the issue's scope.

- **`copy_wallets_followed`** — one row per followed wallet, mutated in place (`status`/`paused_reason`/`last_seen_trade_ts`), unlike `copy_wallet_candidates`' append-only history. A wallet can't be followed twice: `insert_followed_wallet` is a plain `INSERT` and raises `sqlite3.IntegrityError` on a duplicate `address`.
- **`copy_signals`** — one row per detected BUY from a followed wallet, executed or skipped. Per the acceptance criteria, I verified a real raw trade record from `data-api.polymarket.com/trades` live (2026-09-19) and confirmed every record carries a stable `transactionHash`, so I added `source_trade_id TEXT` (nullable, no `UNIQUE` constraint — `normalize_trade()` in `src/data/polymarket_traders.py` doesn't surface this field yet, and a single on-chain transaction can legitimately span multiple maker fills at different price levels, so uniqueness can't be enforced at the DB level). Documented in a DDL comment as a known limitation story B3 inherits: dedupe on `source_trade_id` first, fall back to `(address, market, source_price, detected_at)` on a collision.
- **`copy_positions`** — open, unsettled paper positions. Mirrors `open_positions`' shape conceptually but rows are **retained** (not deleted) on settlement — epic C needs to read settled rows later for P&L history, so `status` flips `'open'` → `'settled'` in place instead.

New `Database` methods (all 7 from the acceptance criteria): `insert_followed_wallet`, `get_followed_wallets`, `update_followed_wallet_status`, `update_followed_wallet_last_seen`, `insert_copy_signal`, `insert_copy_position`, `get_open_copy_positions`.

`docs/DB_SCHEMA.md` gets new sections for all three tables, same format as `open_positions`/`risk_state`.

## How to test
```
python -m pytest src/tests/test_db.py -q
```
[...]

Issue #1122, opened 2026-09-19
Add advisory-only wallet-promotion CLI (follow/pause/resume)


P0K5:
Part of epic #1101. Depends on story #B1 (`copy_wallets_followed` table +
persistence methods).

## Context

Epic A's screening pipeline produces `copy_wallet_candidates` rows with
`eligible_to_follow` computed per run, but nothing turns an eligible
candidate into an actually-followed wallet — that promotion step doesn't
exist yet anywhere in this codebase. This story builds it, deliberately
**advisory-only**, mirroring the exact precedent already established in
this repo for the weather strategy's own shadow→live station promotion
(`src/model/promotion_gate.py`, documented in `docs/OPERATIONS.md`'s
"Station Shadow→Live Promotion" section: "Advisory only — this tool never
promotes anything"). Nothing auto-follows a wallet; a human runs an
explicit command.

This also resolves how Epic B (#1101) gets any followed wallets to poll
at all before Epic F ships a followed-wallet-management dashboard UI
(#1104) — that UI is scheduled well after paper execution starts per the
architecture doc's phasing, so Epic B cannot depend on it existing yet.

## Acceptance criteria

- [ ] New `src/scripts/copy_wallet_promotion.py`, CLI (argparse):
      - **Default (no action flag) — advisory report, changes nothing.**
        Lists wallets from `copy_wallet_candidates` whose *latest*
        screening run has `eligible_to_follow=1`, that are not already in
        `copy_wallets_followed`, sorted by `median_roi` descending, with a
        note of how many follow slots remain
        (`COPY_MAX_WALLETS_FOLLOWED` minus the current count of
        `active` rows in `copy_wallets_followed`).
      - `--follow <address> [--stake <usd>]`: explicit promotion.
        Refuses (clear error, no DB write) unless the address's *latest*
        `copy_wallet_candidates` row has `eligible_to_follow=1` — this is
        the safety rail: nothing gets followed based on stale or
        never-passing data. `--stake` defaults to
        `COPY_DEFAULT_FLAT_STAKE_USD` (Epic E's config, `src/config.py`).
        Also refuses if `COPY_MAX_WALLETS_FOLLOWED` active wallets are
        already followed (report the current count vs. the limit).
        Calls `db.insert_followed_wallet(...)` (#B1).
      - `--pause <address> --reason <text>`: calls
        `db.update_followed_wallet_status(address, "paused", reason)`.
      - `--resume <address>`: calls
[...]

Issue #1121, opened 2026-09-19
Add copy_wallets_followed, copy_signals, copy_positions tables


P0K5:
Part of epic #1101.

## Context

Epic B needs three new tables that don't exist yet: `copy_wallets_followed`
(the subset of screened candidates actually being copied — Epic A only
built the screening/candidate side, `copy_wallet_candidates`), `copy_signals`
(one row per detected BUY from a followed wallet, executed or skipped), and
`copy_positions` (open, unsettled paper positions — settlement/PnL is
Epic C's job, this story only creates the table these rows land in). This
story is schema + persistence only, following the exact `db.py` pattern
used in #1108 (`copy_wallet_candidates`) — no polling, no execution logic.

Per the architecture doc's isolation decision (recorded on issue #1100):
these are fully separate tables, not a `strategy` discriminator column on
the weather strategy's `open_positions`/`trades`. `copy_positions` mirrors
`open_positions`' shape (`docs/DB_SCHEMA.md:177-208`) conceptually but is
NOT deleted on settlement the way `open_positions` is — Epic C needs to
read settled rows later for P&L history, so status flips `open`→`settled`
in place.

## Acceptance criteria

- [ ] `copy_wallets_followed` table (`src/data/db.py` `_DDL`, append near
      `copy_wallet_candidates`): `address TEXT PRIMARY KEY`,
      `stake_per_trade REAL NOT NULL`,
      `status TEXT NOT NULL DEFAULT 'active' CHECK(status IN ('active','paused'))`,
      `paused_reason TEXT`, `added_at TEXT NOT NULL`,
      `last_seen_trade_ts INTEGER` (nullable — the high-water-mark story
      #B3 will use to detect new trades; `NULL` until the first poll).
      One row per wallet (not append-only, unlike `copy_wallet_candidates`)
      — status is mutated in place via `UPDATE`, not a new row per change.
- [ ] `copy_signals` table: `id INTEGER PRIMARY KEY AUTOINCREMENT`,
      `address TEXT NOT NULL`, `market TEXT NOT NULL` (condition ID),
      `outcome_index INTEGER`, `source_price REAL NOT NULL` (the trader's
      own trade price), `detected_at TEXT NOT NULL`,
      `order_placed INTEGER NOT NULL DEFAULT 0`,
      `fill_price REAL` (NULL unless `order_placed=1`),
      `size_usd REAL` (NULL unless `order_placed=1`),
[...]

Issue #1102, opened 2026-09-19
Epic C: Copy-trading settlement and P&L tracking


P0K5:
## Context

Epic C from `docs/design/copy-trading-architecture.md`. Extends the
settlement flow to resolve copy-trading positions and record realized
P&L — the data Epic D's risk monitoring and Epic F's dashboard both read.

## Acceptance criteria

- [ ] Copy-trading positions are settled via the same resolution truth
      source as everything else in this repo
      (`src/data/polymarket.py::fetch_market_resolution`), following the
      `settle.py` pattern — never re-derive resolution truth.
- [ ] Realized P&L is recorded per copy-trading position, attributable to
      the source wallet it was copied from.
- [ ] Per-wallet and aggregate realized P&L are queryable (feeds Epic F's
      Positions & P&L view and its backtest-comparison toggle).
- [ ] A backtest-comparison figure is computable: realized paper P&L for a
      followed wallet vs. what the original screening run projected for
      that wallet's flat-stake scenario — this is what the architecture
      doc's phase-7 go/no-go gate depends on, it needs to exist as a
      queryable number, not be computed ad hoc later.
- [ ] Tests covering: settlement of a copy-trading position, per-wallet
      P&L attribution, the backtest-comparison calculation.

## Technical notes

- If Epic A/E decided to reuse `settlements`/`open_positions` with a
  `strategy` discriminator column, extend `settle.py` directly; if
  separate tables, mirror its logic rather than duplicating it ad hoc.

## Dependencies

Epic B (needs positions to settle).

## Labels

epic, backend, Mid

P0K5:
Epic complete — all three stories merged:

- #1131 — settlement math + `Database.settle_copy_position` + per-wallet/aggregate realized-P&L queries (PR #1134, `292b339`)
- #1132 — `src/scripts/copy_settle.py` one-shot job + `meteoedge-copy-settle.service`/`.timer` (hourly), `docs/OPERATIONS.md` updated (PR #1135, `66dab33`)
- #1133 — `src/data/copy_backtest_comparison.py`, per-wallet and aggregate realized-vs-projected divergence (PR #1136, `841699d`)

All five acceptance criteria from this epic's body are satisfied:
1. Settlement uses `fetch_market_resolution` exclusively, following the `settle.py` pattern — no re-derived truth source.
2. Realized P&L is recorded per position (`copy_positions.settled_pnl_usd`), attributable to its wallet via the `address` column.
[...]

PR #1128, opened 2026-09-20
Add signal-detection and flat-stake paper-execution loop (#1123)

### LINKED ISSUE #1123: Add signal-detection and flat-stake paper-execution loop
Part of epic #1101. Depends on story #B1 (schema) and benefits from story
#B2 (needs at least one followed wallet to do anything meaningful, though
this story's own tests use directly-seeded `copy_wallets_followed` rows
and don't call #B2's CLI).

## Context

This is the epic's core: detect new BUY trades from followed wallets and
paper-execute them at the wallet's configured flat stake. **Paper mode
only — no path to `src/execution/live_trader.py` anywhere in this story.**

**Do not call `src/paper_trader.py`'s `PaperTrader.execute_trade()`.**
It writes to the shared `trades` table (violates the isolation decision
on issue #1100 — copy-trading must never touch weather-strategy tables)
and computes win/loss *synchronously* from an `actual_daily_high`
parameter that's already known at call time. Copy-trading signals don't
have a known outcome at detection time — the market resolves later. This
story only creates an **open, unsettled** `copy_positions` row (#B1);
settling it (computing `settled_pnl_usd`) is Epic C's job (#1102), not
this one.

**Required research before finalizing the polling design:**
`get_wallet_trades()` (`src/data/polymarket_traders.py:85-116`) paginates
**oldest-first** with no date filter (confirmed in its own docstring) — it
fetches from a wallet's very first trade forward. That's fine for Epic
A's daily full-history screening, but wrong for a frequent poll loop that
only wants "what's new since last check": walking the full history every
poll cycle to find the tail is wasteful and slow, especially at scale.
Before implementing the poll loop, check whether the underlying
`data-api.polymarket.com` `/trades` endpoint actually supports a
newest-first order or a min/after timestamp filter that
`get_wallet_trades()` just doesn't expose yet (inspect the live API
response, or its documented query params if discoverable). If it does,
add the parameter to `get_wallet_trades()` (or a new
`get_wallet_trades_since(address, since_ts)` wrapper) and use it here. If
it genuinely doesn't, fall back to the existing oldest-first pagination
and filter client-side to `trade["timestamp"] > last_seen_trade_ts`,
**and document this as a known limitation** (mirroring the tone of the
20,000-trade-cap limitation already documented for Epic A) rather than
silently accepting slow polls for high-volume wallets. `COPY_MAX_WALLETS_FOLLOWED`
defaults to 10 and Epic A's stability check already filters out the
highest-volume/unstable wallets, so the followed set is likely to be
small and moderate-volume in practice — note this mitigating context in
the PR either way.

## Acceptance criteria

- [ ] New `src/scripts/copy_signal_loop.py`: a persistent process (NOT a
      systemd oneshot — see story #B4), structured like `run.py`'s own
      poll loop (`src/scripts/run.py` — check its top-level loop
      structure for the pattern to mirror: sleep-between-cycles, graceful
      shutdown handling, per-cycle exception isolation so one bad cycle
      doesn't kill the process).
- [ ] **Startup sanity check, fail loud:** if
      `COPY_MAX_TOTAL_EXPOSURE_USD > COPY_TRADING_CAPITAL_USD` (both from
      `src/config.py`, Epic E), log an error and refuse to start. These
      two values are currently inconsistent by default (100 vs. 250) —
      harmless today since `COPY_TRADING_ENABLED` defaults `False`, but a
      real footgun once an operator turns it on without noticing. This
      check catches that before it can allow overexposure.
- [ ] **Kill switch, checked every cycle, not just at startup:** if
      `COPY_TRADING_ENABLED` (live-read via `get_live_config(db)`, same as
      every other DB-backed config value) is `False`, the cycle does
      nothing (no polling, no execution) — an operator must be able to
      flip this off mid-run and have it take effect on the next cycle,
      not just at process start.
- [ ] Per cycle, for each `active`-status followed wallet
      (`db.get_followed_wallets(status="active")`, #B1):
      1. Fetch new trades since `last_seen_trade_ts` (per the research
         note above).
      2. For each new **BUY** trade (SELLs are not copied, matching the
         spike's own scope), decide execute-or-skip:
         - Skip (`skip_reason="market_resolved"`) if
           `fetch_market_resolution(market)` (`src/data/polymarket.py`)
           already returns non-`None` — no point paper-buying into an
           already-decided market.
         - Skip (`skip_reason="wallet_exposure_limit"`) if this wallet's
           current open exposure (`sum(stake_usd)` over
           `db.get_open_copy_positions(address)`, #B1) plus the stake
           would exceed `COPY_MAX_EXPOSURE_PER_WALLET_USD`.
         - Skip (`skip_reason="total_exposure_limit"`) if total open
           exposure across all wallets
           (`sum(stake_usd)` over `db.get_open_copy_positions()`) plus the
           stake would exceed `COPY_MAX_TOTAL_EXPOSURE_USD`.
         - Otherwise: compute `fill_price` via `apply_slippage(source_price,
           "BUY", slippage_bps=150)` (reuse verbatim from
           `src/scripts/copy_trade_backtest.py` — same 150bps assumption
           the backtest validated and Epic C's go/no-go comparison
           depends on matching), insert an open `copy_positions` row
           (#B1) sized at the wallet's `stake_per_trade`, and mark the
           signal `order_placed=1`.
      3. Insert a `copy_signals` row (#B1) for every detected BUY
         regardless of outcome — executed or skipped, always with a
         specific `skip_reason` when not executed (never a generic
         "skipped").
      4. Update `last_seen_trade_ts` for the wallet (#B1) after
         processing its batch, so already-processed trades are never
         re-signaled on the next cycle.
- [ ] Per-wallet fetch failures (e.g. `get_wallet_trades` transient
      network error) log a warning and move on to the next wallet —
      one wallet's bad cycle must not block the others, matching this
      codebase's existing fail-soft conventions (e.g.
      `get_wallet_trades`'s own per-page failure handling).

## Technical notes

- `apply_slippage()`, `DEFAULT_SLIPPAGE_BPS` — import from
  `src/scripts/copy_trade_backtest.py`, don't reimplement.
- `fetch_market_resolution()` — `src/data/polymarket.py:148`.
- Config reads: `COPY_TRADING_ENABLED`, `COPY_MAX_EXPOSURE_PER_WALLET_USD`,
  `COPY_MAX_TOTAL_EXPOSURE_USD`, `COPY_TRADING_CAPITAL_USD`,
  `COPY_MAX_WALLETS_FOLLOWED` — all added in #1115/PR #1116.
- Poll cadence: new `COPY_SIGNAL_POLL_INTERVAL_SECONDS` config key
  (add to `CONFIG_DEFAULTS`/`_CONFIG_META` in this PR, following the
  exact pattern of the existing `POLL_INTERVAL_SECONDS` — this is a
  legitimate live-read value here, unlike Epic A's screening cadence,
  because this script is a genuine persistent loop like `run.py`, not a
  systemd oneshot). Suggest a default of 300 (5 minutes), matching
  `POLL_INTERVAL_SECONDS`'s own default.

## Test requirements

New `src/tests/test_copy_signal_loop.py`, fully mocked (DB, network):
- Startup sanity check refuses to start when
  `COPY_MAX_TOTAL_EXPOSURE_USD > COPY_TRADING_CAPITAL_USD`.
- Kill switch: `COPY_TRADING_ENABLED=False` produces no polling and no DB
  writes for a cycle.
- A new BUY trade from an active wallet, market not yet resolved, under
  both exposure limits → executes: `copy_signals` row with
  `order_placed=1`, `copy_positions` row created with `entry_price`
  matching `apply_slippage(source_price, "BUY", 150.0)`.
- Same trade but market already resolved → `skip_reason="market_resolved"`,
  no `copy_positions` row.
- Same trade but wallet already at `COPY_MAX_EXPOSURE_PER_WALLET_USD` →
  `skip_reason="wallet_exposure_limit"`.
- Same trade but total exposure already at `COPY_MAX_TOTAL_EXPOSURE_USD`
  → `skip_reason="total_exposure_limit"`.
- A SELL trade is never turned into a signal at all.
- `last_seen_trade_ts` only advances past trades that were actually
  processed this cycle.
- A `get_wallet_trades` failure for one wallet doesn't prevent other
  wallets in the same cycle from being processed.

## Dependencies

Story #B1 (schema). Benefits from #B2 for realistic manual testing but
has no code dependency on it.

## Labels

backend, Complex

ClaudioSenior:
## Linked issue
Closes #1123

## What changed
Adds `src/scripts/copy_signal_loop.py`: a persistent poll loop, structured
like `src/scripts/run.py`'s own loop (sleep-between-cycles, graceful
Ctrl-C shutdown, per-cycle/per-wallet exception isolation). Every cycle,
for each `active` followed wallet:

1. Fetch trades newer than `last_seen_trade_ts` via the new
   `get_wallet_trades_since()` (see "Research finding" below).
2. For each new **BUY** trade (SELLs are never copied), decide
   execute-or-skip against, in order: already-resolved market
   (`skip_reason="market_resolved"`) then missing `outcome_index`
   (`skip_reason="missing_outcome_index"`, see "Judgment calls" below)
   then per-wallet exposure cap (`skip_reason="wallet_exposure_limit"`)
   then total exposure cap (`skip_reason="total_exposure_limit"`) then
   execute.
3. Execution: `fill_price = apply_slippage(source_price, "BUY", 150.0)`
   (reused verbatim from `copy_trade_backtest.py`), an open `copy_positions`
   row sized at the wallet's `stake_per_trade`, and the `copy_signals` row
   marked `order_placed=1` and linked to the new position.
4. Every detected BUY gets a `copy_signals` row regardless of outcome.
5. `last_seen_trade_ts` advances to the max timestamp actually processed
   this cycle (covers BUY and SELL alike, so watermark tracking stays
   correct even though only BUYs generate signals).

**Startup sanity check** (`startup_sanity_check`): refuses to start if
`COPY_MAX_TOTAL_EXPOSURE_USD > COPY_TRADING_CAPITAL_USD`.
**Kill switch**: `COPY_TRADING_ENABLED` is live-read via `get_live_config(db)`
at the top of every cycle, not just at startup.

Never calls `src/execution/live_trader.py` or
`src.paper_trader.PaperTrader.execute_trade()` -- paper mode only, per the
issue's explicit isolation note (#1100). Only ever creates an open,
unsettled `copy_positions` row; settlement is epic C's (#1102) job.

### Research finding (issue's required pre-implementation check)
`get_wallet_trades()`'s docstring claimed oldest-first pagination via
`data-api.polymarket.com/trades`, but that was never actually verified
against live traffic (only field names were, per its 2026-09-18 note).
I verified directly against the live endpoint today (two independent
wallets, 100+ trades each, checked contiguity across offset pages): the
endpoint actually returns **newest-first** by default, and does **not**
honor an explicit since/after/min-timestamp filter (an unrecognized
`after=<ts>` param is silently ignored -- confirmed byte-identical
results with/without it). Added `get_wallet_trades_since()`
(`src/data/polymarket_traders.py`) to exploit the newest-first default --
it walks pages front-to-back and stops as soon as it sees a trade at or
before `since_ts`, so a normal-cadence poll typically costs one page, not
a full-history walk. Corrected `get_wallet_trades()`'s own docstring to
match verified reality (behavior unchanged -- full-history callers
aggregate regardless of order). `COPY_MAX_WALLETS_FOLLOWED` defaults to
10 and Epic A's stability screen already filters the highest-volume /
unstable wallets, so the followed set is expected to be small and
moderate-volume -- noted as mitigating context, not a guarantee, directly
in the new function's docstring (known-limitation tone matching the
existing 20,000-trade-cap note).

### Other additions beyond the literal acceptance criteria (flagging for review)
- **Crash-recovery dedup** (`Database.copy_signal_exists_for_trade`): the
  `copy_signals` schema comment (from #1121 / PR #1125) explicitly hands
  this off to this story (Story B3 must dedupe on `source_trade_id`
  first, falling back to the full `(address, market, source_price,
  detected_at)` tuple when it collides). I implemented dedup keyed on
  `(address, market, source_price, source_trade_id)` -- I read
  `detected_at` in that schema comment as unreliable to match across a
  genuine crash/retry (it is wall-clock write time, not a stable
  trade-identity field), so I substituted `source_trade_id` itself as
  the stable component instead of literally including `detected_at`.
  Flagging this interpretation explicitly in case the intent was
  different.
- Added `source_trade_id` extraction (`transactionHash`) to
  `normalize_trade()` -- the schema comment says this story must add it;
  required an existing exact-dict-equality test update in
  `test_polymarket_traders.py`.
- `missing_outcome_index` skip reason: not one of the issue's three
  enumerated skip categories, but `copy_positions.outcome_index` is
  `NOT NULL`, so a BUY trade `normalize_trade()` could not confidently
  assign a side to cannot be executed. Signal-logged, never silently
  dropped.

## Judgment calls flagged rather than resolved unilaterally
Per your instructions on risk-limit / market-resolution judgment calls:

**First-ever poll for a newly-followed wallet** (`last_seen_trade_ts`
is `NULL`): I implemented the literal reading -- `since_ts` treated as
`0`, so the entire fetched trade history (bounded by `MAX_TRADE_PAGES`,
same cap as `get_wallet_trades()`) is walked through the normal
market-resolved / exposure-limit decision ladder on the wallet's first
poll. This means a wallet with old BUYs on still-open, long-duration
markets (e.g. an election market) could paper-open a `copy_positions`
row today priced at that historical trade's price -- not a
currently-tradable price. I considered special-casing "skip execution,
watermark-only" on first poll to avoid this, but that is a fourth,
unenumerated skip category and a real behavior deviation from the
acceptance criteria as written, so I left it as the literal
implementation and am flagging it here instead of guessing. Happy to add
the bootstrap-skip behavior as a fast-follow if you would prefer that
default.

Also note: `docs/design/copy-trading-architecture.md`'s Epic B summary
and its "Order execution" table row still say trades go through
`paper_trader.py` -- stale versus the issue's explicit instruction not to
use it. Left the doc untouched (out of this PR's file scope) but
flagging so it does not mislead a future reader.

## How to test
```
python -m pytest src/tests/test_copy_signal_loop.py src/tests/test_polymarket_traders.py src/tests/test_db.py src/tests/test_config_api.py src/tests/test_copy_trade_backtest.py src/tests/test_copy_wallet_screening.py -v
```
All new/updated tests pass (243 tests). Full-suite run: 36 pre-existing
failures reproduce identically on an untouched `master` checkout in this
Windows dev environment (Windows-only path-casing / tmp-file-locking
issues in `test_ensure_worktree.py`, `test_shadow_low_rollout.py`, etc. --
unrelated to this change; CI runs on `ubuntu-22.04`). None of the files
this PR touches appear in that failure list.

Manual smoke test (requires `COPY_TRADING_ENABLED=true` and at least one
`copy_wallets_followed` row, e.g. via #1122's CLI):
```
python -m src.scripts.copy_signal_loop --once
```

## How to deploy
- [x] **Code only** -- `git pull`; this script is not yet wired into a
      systemd unit (that is story #B4) -- an operator runs it manually
      (`python -m src.scripts.copy_signal_loop`) until #B4 lands.
- [x] **Config / env change** -- new `COPY_SIGNAL_POLL_INTERVAL_SECONDS`
      (default 300s), live-editable via the dashboard config tab
      (`copy_trading` group) or the `COPY_SIGNAL_POLL_INTERVAL_SECONDS`
      env var. No action needed to pick up the default.

## Checklist
- [x] Tests pass locally (see "How to test")
- [x] Docs updated if this PR adds/changes env vars, DB schema, API endpoints, or run modes -- N/A, no doc changes (see flagged staleness note above)


P0K5:
Closing and immediately reopening — CI has stopped registering entirely for this PR's last two pushes (confirmed via the Actions API: 6 successful runs exist for the first 2 commits, zero for the current head), despite CI working fine on other branches in the same window. This looks like a stuck webhook/dispatch for this specific PR. Reopening to force a fresh pull_request event.

P0K5 (approved):
Approved after a full manual review (trading-logic PRs get the deep-dive regardless of AI verdict, per the pr-review skill). Read copy_signal_loop.py, get_wallet_trades_since(), and the new db.py methods (link_copy_signal_to_position, copy_signal_exists_for_trade) directly — all correct: skip-reason ladder matches spec, exposure checks are atomic under db._lock with the blocking network call correctly kept outside it, watermark advances incrementally and crash-safely, bootstrap guard correctly prevents stale-price backlog execution on a wallet's first poll, and the newest-first pagination research/fix is sound with a sensible early-stop boundary. All CI green including a clean AI/DeepSeek review after 2 rounds of legitimate findings the dev already fixed independently. Great work on this one — the two flagged judgment calls were exactly the right things to surface rather than guess on.

PR #1154, opened 2026-09-20
ci: add guard against graphify-out/ changes in feature PRs

### LINKED ISSUE #1153: Add CI guard against graphify-out/ changes in non-graphify-update PRs
## Context

Third occurrence this session of the same mistake: a feature PR
accidentally bundles a full regenerated `graphify-out/graph.json` (and
its sibling files) diff — 150k+ lines — instead of leaving that directory
untouched. The established convention (documented in `.claude/instructions/graphify.md`
and followed correctly by most PRs) is that `graphify-out/*` is only ever
updated by `.github/workflows/graphify-update.yml`'s own dedicated
follow-up commits (`chore(graph): incremental graph update [skip ci]`),
never bundled into a feature PR. Relying on every agent remembering this
by convention alone has failed three times
(PRs for issues #1140, and twice more during epic F) — this needs an
automated guard, not another reminder.

## Acceptance criteria

- [ ] CI (`.github/workflows/ci.yml`, or a new lightweight workflow) fails
      a PR whose diff touches any path under `graphify-out/` **unless**
      the PR's own commits are exclusively the automated
      `graphify-update.yml` workflow's commits (or some equivalent,
      reliable signal — e.g. check the PR author/actor, or simply: any
      human/agent-authored PR touching `graphify-out/**` fails with a
      clear error message telling the author to revert those files).
- [ ] The check's failure message explains why (points at this
      convention) so a future contributor isn't left guessing.
- [ ] Verify the new check does NOT fire on the actual scheduled
      `graphify-update.yml` runs themselves (test against that workflow's
      own recent commit pattern) — this guard must not break the
      legitimate automation it's protecting.

## Technical notes

- Simplest implementation: a small job/step in `ci.yml` that runs
  `git diff --name-only origin/master...HEAD -- graphify-out/` (or the
  PR's actual base) and fails if non-empty, gated to skip when running
  as part of the `graphify-update.yml` workflow itself (different
  workflow file, so this is naturally already scoped correctly — a step
  in `ci.yml` never runs as part of `graphify-update.yml`).

## Dependencies

None.

## Labels

Simple, backend

P0K5:
Closes #1153

Adds a CI guard that prevents feature/bugfix PRs from accidentally bundling regenerated graphify-out/ diffs.

## Changes

- scripts/check_graphify_guard.sh: New CI guard script that detects graphify-out/ changes
- scripts/test_graphify_guard.sh: Unit tests for the guard script
- .github/workflows/ci.yml: Added graphify-guard job

## How to test

Guard passes when no graphify-out/ changes, fails when graphify-out/ is modified in non-graphify-update PRs.

## How to deploy

No deploy needed — CI-only change.

---
_Generated by [Claude Code](https://claude.ai/code/session_01FytJcua4suHrq3KBBCSpdg)_


P0K5:
Approved (via comment — GitHub API collision).

Rationale: Matches #1153's intent, revised after review. Original draft had two real problems I fixed directly:
1. The "branch/title contains 'graphify'" escape hatch was backwards — `graphify-update.yml` never opens a PR (verified directly: its `on:` block is `push` to master + `schedule` + `workflow_dispatch` only, every job ends in a direct `git push`), so no legitimate PR needs that exemption, and it just gave every PR a trivial bypass. Removed; the guard now fails unconditionally on any PR touching `graphify-out/`, which is what actually would have caught all three real incidents this session (none had "graphify" in their branch names).
2. The original test suite was fake — an elaborate curl-mocking scaffold was built but never invoked; every actual test just grepped the script's own source text for substrings. Rewrote to extract the decision logic into a pure, sourceable function and test it directly against fixture data.

AI reviewer then correctly BLOCKed two more real issues on the first revision, both fixed: a missing pagination loop (a graphify-out/ file beyond the 100-file first page would have silently passed), and asked for the "never opens a PR" claim to be backed by more than assertion (now quotes the actual trigger block in the script's own comment). Added two genuine integration tests exercising the pagination loop via a stub `curl` — which caught a real bug in the stub itself (naive substring matching on "page=1" spuriously matched every page, since "per_page=100" contains that substring) before it shipped. Added a defensive MAX_PAGES cap and a `timeout` around every test invocation so a future bug fails fast instead of hanging CI.

CI green (Lint, Test, AI/DeepSeek review, gitleaks, and the new guard job itself) on `5c3a72b`. Backend/CI-only, no dashboard files touched — no Designer approval needed.

Merging.

---

